PR #1, opened 2018-10-09
Show status of api on frontend

KPobeeNorris:
Call to index calls the API endpoint and returns the status


jennyd:
It looks like this also includes some cleaning up of things we don't need that were generated at the start - please could you split that out into a separate commit? ✨

PR #5, opened 2018-10-10
Add Circleci config

KPobeeNorris:


.circleci/config.yml
```diff
@@ -0,0 +1,66 @@
+version: 2
+jobs:
+  build:
+    parallelism: 1
+    working_directory: ~/allocation-manager
+    docker:
+      - image: circleci/ruby:2.5.1-node-browsers
```
alan: We'll have to change this soon when we do the deployment config

PR #9, opened 2018-10-11
Create CircleCI work-flows configuration

TobyRet:
Setup CircleCi workflows in order to create split jobs and create deployment pipeline

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -50,26 +50,26 @@ jobs:
           command: |
             curl -L https://codeclimate.com/downloads/test-reporter/test-reporter-latest-linux-amd64 > ./cc-test-reporter
             chmod +x ./cc-test-reporter
-      - run:
-          name: install firefox
-          command: |
-            if [[ $(command -v firefox) ]] && [[ $(firefox --version) == "Mozilla Firefox $FIREFOX_VERSION" ]]; then
-              echo "Skiping, Firefox $FIREFOX_VERSION already install."
-            else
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -1,34 +1,55 @@
...
+    <<: *defaults
+    steps:
+      - checkout
+      - attach_workspace:
+          at: ~/repo
+      - run: bundle --path vendor/bundle
+      - run:
+          name: Download Code Climate
+          command: |
+            curl -L https://codeclimate.com/downloads/test-reporter/test-reporter-latest-linux-amd64 > ./cc-test-reporter
```
alan: Can the code climate tool be persisted to the workspace, or is it already cached?

.circleci/config.yml
```diff
@@ -49,11 +70,6 @@ jobs:
               sudo tar -zxvf geckodriver-v0.19.0-linux64.tar.gz -C /opt
```
alan: This might not be needed, the circle browsers image [installs](https://github.com/circleci/circleci-images/blob/0d8b6758dc1214046a3f5e7f14297b1a8572b479/shared/images/Dockerfile-browsers.template#L32) the `geckodriver`

.circleci/config.yml
```diff
@@ -1,34 +1,55 @@
...
+      - attach_workspace:
+          at: ~/repo
+      - run: bundle --path vendor/bundle
+      - run:
+          name: Download Code Climate
+          command: |
+            curl -L https://codeclimate.com/downloads/test-reporter/test-reporter-latest-linux-amd64 > ./cc-test-reporter
+            chmod +x ./cc-test-reporter
       - run:
           name: install firefox
```
alan: If we need this step we should rename it to install geckodriver

PR #48, opened 2018-12-04
[MOICM-77] - Authentication: Users can sign in

TobyRet:
- Endpoints can now be protected if required. 
- Users trying to visit protected endpoints will need to sign in via the Nomis Oauth2 server


app/assets/stylesheets/hmpps_header.scss
```diff
@@ -12,7 +12,7 @@
   height: 30px;
   margin-bottom: -2px;
   margin-right: 10px;
-}
+}s
```
KPobeeNorris: A spare 's'?

TobyRet:
We believe that there may be a regression in Code Climate which means code wrapped in #nocov is not being ignored. Safe to merge

PR #60, opened 2018-12-13
[MOICM-89] - Create Kubernetes production deploy scripts

TobyRet:
- [DO NOT MERGE] - production secrets still need to be added


jennyd:
We need to change the value of `NOMIS_OAUTH_PUBLIC_KEY` in the production deployment.yaml to the prod oauth server's public key - it's still got the same one as staging/T3 at the moment.

TobyRet:
Will reopen eventually

PR #62, opened 2018-12-17
[MOICM-67] - Display Prisoners Awaiting Tiering

TobyRet:
- Displays list of prisoners in the 'awaiting tiering' tab in '/allocations'
- If a prisoner had no release details then nothing is displayed under 'release date'. 
- The client handles 401 responses from the Custody API and raises an APIError. This then needs to be rescued by the caller accordingly (i.e. by returning a Null object).


config/application.rb
```diff
@@ -20,7 +20,7 @@ class Application < Rails::Application
       'OFFENDER_MANAGEMENT_ALLOCATION_API',
       'http://localhost:8000'
     )
-    config.sentry_dsn = ENV.fetch('SENTRY_DSN', nil)
+    config.sentry_dsn = ENV.fetch('SENTRY_DSN', 'nil')
```
rossjones: I'm a bit unsure about the `'nil'`, is this because something somewhere else requires it to be a string?

TobyRet: mmm ... does not seem right. Will check @rossjones

TobyRet: I've removed the default value

PR #63, opened 2018-12-18
[MOICM_94] - Adds a basic dashboard page

rossjones:
Adds the basic dashboard (with links but no search or prison selection) to allow users to navigate through the system as new pages are added. Currently only the allocation links work, the others will just reload the same page.

As part of this PR, the application.css file was renamed to application.scss to make the CSS more usable and minimal changes added for the formatting/display of the dashboard rows.

Moves the status endpoint to /status in case it is still useful.

It is expected that as more pages are available, this page will also be changed as part of those PRs to ensure the link works.

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -13,3 +13,28 @@
  *= require_tree .
  *= require_self
  */
+
+
+ .no-bottom-margin {
+     margin-bottom: 0;
+ }
+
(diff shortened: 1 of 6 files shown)


TobyRet:
Just need to fix failing test :-)

rossjones:
> Just need to fix failing test :-)

And add a test for the new page ... oops

PR #67, opened 2018-12-19
[MOICM-99] - Session Expiry

TobyRet:
- When authenticating a request check to see if the user's Oauth token
expiry time has been reached. If so, destroy their session and redirect
them to Nomis Oauth2 SSO login.


app/controllers/application_controller.rb
```diff
@@ -4,7 +4,7 @@ class ApplicationController < ActionController::Base
   include SSOIdentity
 
   def authenticate_user
-    unless sso_identity
+    unless sso_identity && !session_expired?
       session[:redirect_path] = request.original_fullpath
```
TobyRet: mmm Something not right here methinks ... should we be logging out the user from SSO first?

rossjones: This works okay for me. It does crash on first run of the branch (if you are currently logged in, because your session info doesn't include the expiry field), but it seems to behave the way we discussed.

app/controllers/application_controller.rb
```diff
@@ -17,4 +17,8 @@ def current_user
   def caseload
     sso_identity['caseload']
   end
+
+  def session_expired?
```
rossjones: What do you think about reversing the logic to make this `session_valid?`.  Just a thought.

TobyRet: Yes good point - will change

PR #82, opened 2019-01-04
Adds the mechanics of paging the Custody API calls

rossjones:
Allows a page number (taken from param called page) to be passed to the
Custody API to change which page worth of data is returned. It's worth noting that
the app works on 1-based page numbers (because it makes more sense) and 
we do the translation to the NOMIS 0-based page indexing.

To make this feasible we've change the API calls to return a ApiObjectResponse
or an ApiListResponse rather than just the raw data.  These two structs
contain either page-meta (serialized to an in-memory model) and data
(for the lists) or just the data field (for objects). There is a chance
to the custody client that allows a block to be passed to the get()
call, so that action can be taken before the parsed json is returned -
in our case, it is to throw an APIError if the key for the data is
missing (likely we paged too far ahead).  We also ensure that any calls
to list generating apis (like get_offenders) always return the page
data.


app/services/nomis/custody/api.rb
```diff
@@ -1,5 +1,8 @@
 module Nomis
   module Custody
+    ApiObjectResponse = Struct.new(:data)
+    ApiListResponse = Struct.new(:meta, :data)
```
TobyRet: I wonder if what are returning here is actually a ApiResponse and an _ApiPaginatedResponse_?  mmm ... I guess it may not always be paginated though ...

app/services/nomis/custody/api.rb
```diff
@@ -17,26 +20,44 @@ def initialize
...
-      def get_offenders(prison)
-        route = "/custodyapi/api/offenders/prison/#{prison}?page=0&size=10"
-        response = @custodyapi_client.get(route)
-        response['_embedded']['offenders'].map do |offender|
+      # rubocop:disable Metrics/MethodLength
+      def get_offenders(prison, page = 0)
+        route = "/custodyapi/api/offenders/prison/#{prison}?page=#{page}&size=10"
+        page_meta = nil
+
+        response = @custodyapi_client.get(route) { |data|
```
TobyRet: Just noticed you are passing a block :-)

rossjones: I switched it around, so rather than telling the get() method when to fail, we instead just inject the block in between when the call has completed and before it has returned so we can abort if necessary.  Less direct impact on the client.

app/services/nomis/custody/api.rb
rossjones: I think the presence of the meta part for list calls means it sort of is paginated, even if there are no pages - will update the names.

app/services/nomis/custody/api.rb
TobyRet: Good idea

PR #88, opened 2019-01-09
Create Prometheus exporter

TobyRet:
Add Prometheus exporter to the application.

This will allow us to create and capture application metrics which can then:

- become searchable via Prometheus Querying.
- generate alerts
- graphed via Granfana


jennyd:
There's more discussion on the equivalent PR on the API - the same points apply to this one too: https://github.com/ministryofjustice/offender-management-allocation-api/pull/53#issuecomment-453055259

PR #91, opened 2019-01-10
Allocate a POM - start of prisoner info

KPobeeNorris:
As part of the "allocate a POM" work flow we need to display information
related to the offender needing to be allocated.  The routes and views
have been updated to display the information returned.

We have created a NullOffenderDetails object to handle the offender not
being within NOMIS and we will need to establish how to handle this on
the frontend in a following iteration.


app/views/prisoners/show.html.erb
```diff
@@ -83,9 +83,9 @@
   <section class="govuk-tabs__panel govuk-tabs__panel--hidden"
            id="prisoner-info">
     <div>
-      <%= render 'shared/offence_info' %>
-      <%= render 'prisoners/case_information' %>
-      <%= render 'prisoners/allocation_info' %>
+      <%# <%= render 'shared/offence_info' %> %>
```
TobyRet: Comments?

config/routes.rb
```diff
@@ -18,5 +18,8 @@
 
   get('/prisoners/:id' => 'prisoners#show', as: 'prisoners_show')
 
-  resource :allocate_prison_offender_managers, only: %i[show edit]
+  get('/allocate_prison_offender_managers/:noms_id' => 'allocate_prison_offender_managers#show',
+      as: 'allocate_prison_offender_managers_show')
+  get('/allocate_prison_offender_managers/:noms_id/edit' => 'allocate_prison_offender_managers#edit',
```
TobyRet: This should be a 'create' action

PR #94, opened 2019-01-11
Ignore auth requests when running tests with VCR

rossjones:
The auth requests for VCR are mocked, and do not need to be recorded,
and so there is no need to record the auth requests that are failing (in
this case requests to /auth/oauth/token.


rossjones:
Suggest closing this PR in favour of https://github.com/ministryofjustice/offender-management-allocation-manager/pull/96 as that PR is based off master, which at least works for some of us and resolves the issue when re-creating the cassettes.

PR #98, opened 2019-01-14
[Do not merge] - Refresh CircleCI GPG Key Pair

TobyRet:
- New gpg key pair generated for CircleCI Allocation API job that meets agreed standard.
- Old gpg public key will be removed once new keypair deployed and working ok.


TobyRet:
Closing PR until we resolve issues around private keyword password and CircleCI

PR #109, opened 2019-01-17
Only use the Promentheus Exporter client when running the app in production mode

TobyRet:


jennyd:
How do we work on metrics in development?

rossjones:
When we test that we can send metrics to the exporter, then we should have the rspec tests start/stop the collector (if necessary).  Once we have an idea of which metrics we want to collect (and where) we can come back and change this so it is on by default and can be changed with an env-var, but at present the only card we have is for setup and we've yet to define which app-metrics we want to collect.

PR #113, opened 2019-01-18
Get offenders list using Elite2

rossjones:
Currently we make the call to the Custody API to get the list of
offenders, and then for each offender we make yet another request to get
the release date.

We have now moved to the Elite2 API, and as a result, we have a new API
to get the offenders at a specific prison, along with a new model
(OffenderShort) intended for showing the offenders in lists as there is
a reduced set of information available.

After getting a page worth of offenders (10) we now send all 10 ids to
the endpoint from where we can get the release date (going from 1+10
requests to 2 requests).

The calls to get the release dates is marred by what appears to be a
faraday bug.  If we send 10 items we only get a result for the last
record, if we send 9 we get 9 records back.  By adding an empty
element to the end of the prisoner id list we get the correct results.
This probably needs further investigation.

Some of the changes in the client may not be obvious, but we have
extended the get() method to allow for the optional passing of extra
headers (to support paging) and query parameters as a hash.  The latter
issue is due to the way that faraday encodes parameters with the same
name.  We want to specify offenderNo multiple times, and rather then get
offenderNo=1&offenderNo=2 by default faraday will return
offenderNo[]=1&offenderNo[]=2.  We have invoked the appropriate config
option to fix this.

As a bonus for reading this far, the allocation controller and
appropriate views have also been switched over to use the new model so
that once the new model is used for the 'allocate a prisoner/pom' page
is done, we can remove most of the custody api.


rossjones:
@tobyret Should all be resolved now.

PR #115, opened 2019-01-21
Fetch multiple NDelius records from the fake server

TobyRet:


app/services/ndelius/api.rb
```diff
@@ -4,10 +4,15 @@ class Api
...
       delegate :get_record, to: :instance
+      delegate :get_records, to: :instance
     end
 
     def get_record(nomis_id)
       FakeRecord.generate(nomis_id)
     end
+
+    def get_records(nomis_ids)
+      nomis_ids.map { |id| FakeRecord.generate(id) }
```
rossjones: How will we map the nomis_ids provided to the FakeRecords returned?  
Might it be better to return 

```
{
    "id1": FakeRecord...,
    "id2": FakeRecord...,
}
```

TobyRet: Cool - have pushed the change @rossjones

PR #117, opened 2019-01-21
Fetch the tier info for the allocation tables

rossjones:
Also contains a fix for getting the release dates, which should at some
point be moved into the service rather than be a part of the elite2 api.


spec/services/ndelius/fake_record_spec.rb
```diff
@@ -55,7 +55,11 @@
     context 'when the Nomis id ends with Q, R, S, T' do
       it 'raises a NoTier exception' do
```
TobyRet: We are not expecting an exception any more

rossjones: fixed,

PR #120, opened 2019-01-22
Handle embedded objects in API response

rossjones:
Most of the values related to the hash are strings or integers, but
occassionally we encounter an embedded object.  This object is currently
stored as a string, and decoded from the JSON on first use.

This PR changes how embedded objects are made available in the memory
model.  If we remove the type from the attribute and add a serialize
field like

```
serialise :field, JSON
```

the `:field` will be available as a hash (or list) based on the JSON
attached to that key.


rossjones:
@TobyRet @KPobeeNorris This PR is only changing offender_detail because it's an example of use it's only the changes to memory_model we actually care about here.

PR #121, opened 2019-01-22
Get single offender elite2

KPobeeNorris:


app/services/nomis/elite2/api.rb
```diff
@@ -40,6 +41,21 @@ def get_offender_list(prison, page = 0)
 
         ApiPaginatedResponse.new(page_meta, offenders)
       end
+
+      def get_offender(offender_no)
+        route = "/elite2api/api/prisoners/#{offender_no}"
+        response = @e2_client.get(route) { |resp|
```
rossjones: The block here receives two args, 'data' and 'response' ... I think this is working because it's seeing the first one (the json blob) but it might be worth renaming resp to data to be clear?

app/services/nomis/elite2/api.rb
```diff
@@ -40,6 +41,21 @@ def get_offender_list(prison, page = 0)
 
         ApiPaginatedResponse.new(page_meta, offenders)
       end
+
+      def get_offender(offender_no)
```
TobyRet: The OffenderService is passing in `noms_id` as the param

spec/features/offender_allocations_spec.rb
```diff
@@ -4,7 +4,7 @@
   it 'shows the allocate a POM page', vcr: { cassette_name: :get_offender_details, match_requests_on: [:query] } do
     signin_user
 
-    visit allocate_prison_offender_managers_new_path('G2911GD')
+    visit allocate_prison_offender_managers_new_path('G4273GI')
```
TobyRet: Should the id be stored in a var to avoid confusion (e.g. `noms_id`)

spec/services/offender_service_spec.rb
```diff
@@ -19,6 +19,6 @@
 
   it "gets a single offender", vcr: { cassette_name: :get_single_offender } do
     offender = OffenderService.new.get_offender('G4273GI')
```
TobyRet: Same here - might be worth being explicit that this is a `noms_id` and store in a var

PR #135, opened 2019-01-23
Display POMs

TobyRet:
This pull request removes hard-coded Prisoner Offender Manager (POM) information and replaces (temporarily) with data fetched from the keyworker end point via the Elite2 API.

(Also temporary) A fake request is made to the Allocation API to obtain allocation records for a POM.


app/views/allocate_prison_offender_managers/_pom_tables.html.erb
```diff
@@ -20,23 +20,25 @@ Recommendation:
...
+    <tr class="govuk-table__row">
+      <td aria-label="POM name" class="govuk-table__cell "><%= pom.full_name %></td>
+      <td aria-label="Previous allocation" class="govuk-table__cell">-</td>
+      <td aria-label="Tier A cases" class="govuk-table__cell "><%= pom.tier_a %></td>
+      <td aria-label="Tier B cases" class="govuk-table__cell "><%= pom.tier_b %></td>
+      <td aria-label="Tier C cases" class="govuk-table__cell "><%= pom.tier_c %></td>
+      <td aria-label="Tier D cases" class="govuk-table__cell "><%= pom.tier_d %></td>
+      <td aria-label="Total" class="govuk-table__cell "><%= pom.total_cases %></td>
+      <td aria-label="Working pattern" class="govuk-table__cell"><%= pom.working_pattern %></td>
+      <% if current_page?(action: 'show') %>
```
rossjones: I think these are the wrong way around, but we can fix this when we do the actual allocation.

app/services/nomis/elite2/prisoner_offender_manager.rb
```diff
@@ -0,0 +1,25 @@
+# Note this is currently deserializing keyworker
+module Nomis
+  module Elite2
+    class PrisonerOffenderManager
```
rossjones: I think this is PrisonOffenderManager rather than PrisonerOffenderManager

app/views/allocate_prison_offender_managers/_pom_tables.html.erb
KPobeeNorris: They are; I missed this when we changed the route/action from edit to show in PR#91

rossjones:
@TobyRet This looks good to merge now

app/services/nomis/elite2/prisoner_offender_manager.rb
TobyRet: Good catch @rossjones - will change

PR #138, opened 2019-01-24
Remove offender short accessors

rossjones:
The OffenderShort model currently has attr_accessors for tier and
release_date but these can be attributes, and so this PR converts them
to attributes for consistency.

The second commit does the same for the PrisonOffenderManager model.


rossjones:
Will leave this unmerged until later in case it helps debug codeclimate issue.

PR #139, opened 2019-01-24
Rename VCR cassettes so they can be clearly matched to the correct test

TobyRet:
- VCR cassettes are renamed to close match the spec that it is associated to
- This is to prevent any confusion arising from using the same cassette across multiple specs

### DIFF
--- a/spec/features/allocations_summary_feature_spec.rb
+++ b/spec/features/allocations_summary_feature_spec.rb
@@ -13,7 +13,7 @@ feature 'allocations summary feature' do
       expect(page).to have_css('.pagination ul.links li', count: 7)
     end
 
-    it 'renders allocation offenders at index', :raven_intercept_exception, vcr: { cassette_name: :allocated_offenders } do
+    it 'renders allocation offenders at index', :raven_intercept_exception, vcr: { cassette_name: :allocated_offenders_feature } do
       signin_user
       visit 'allocations'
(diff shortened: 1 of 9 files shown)


spec/features/allocations_summary_feature_spec.rb
```diff
@@ -13,7 +13,7 @@
       expect(page).to have_css('.pagination ul.links li', count: 7)
     end
 
-    it 'renders allocation offenders at index', :raven_intercept_exception, vcr: { cassette_name: :allocated_offenders } do
+    it 'renders allocation offenders at index', :raven_intercept_exception, vcr: { cassette_name: :allocated_offenders_feature } do
```
KPobeeNorris: All VCR names LGTM. However, just wondering about this file and whether they could be named slightly differently as when we add more feature specs we're could end up needing to refer to allocated/awaiting/missing info tabs and there could be some similarly named cassettes used?

TobyRet: @KPobeeNorris I think rose said that this view will be reduced to one call to the controller again so I imagine we will only need one cassette in the near future. Happy to rename now though it it would make things clearer?

KPobeeNorris: Ah - no worries if it's going to be changed!

PR #148, opened 2019-01-28
Production deploy config

jennyd:
Mostly inspired by @TobyRet's previous PRs on the same subject :)

The production secrets file needs some real secrets adding in once we're confident that the deploy pipeline is working correctly.


TobyRet:
LGTM! 🥇

PR #151, opened 2019-01-29
Set up call to 'allocate' endpoint on Allocation API

TobyRet:
- post request and 500 error handling added to Allocation Client

### DIFF
--- a/app/services/allocation/api.rb
+++ b/app/services/allocation/api.rb
@@ -24,9 +24,9 @@ module Allocation
       end
     end
 
-    def allocate(allocation_params)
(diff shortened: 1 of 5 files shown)


app/services/allocation/api.rb
```diff
@@ -22,5 +23,10 @@ def get_allocation_data(staff_ids)
         hash[id] = FakeAllocationRecord.generate(id)
       end
     end
+
+    def allocate(body)
+      route = '/allocate'
```
rossjones: This should be /allocation

PR #160, opened 2019-02-06
Enable requests with LOTS of IDs for sentence info

rossjones:
Currently when we fetch the offender sentences in bulk (to get the
release date) we are limited by server limits on the length of a GET
request.

This PR switches this call over to use POST so that that limit does not
apply.


app/services/nomis/client.rb
```diff
@@ -31,14 +31,24 @@ def get(route, queryparams: {}, extra_headers: {})
...
+          'Content-Type' => 'application/json'
+        }
+      )
+      JSON.parse(response.body)
+    end
+
   private
 
-    def request(method, route, queryparams: {}, extra_headers: {})
+    def request(method, route, body: nil, queryparams: {}, extra_headers: {})
```
TobyRet: Should the body default be an empty hash? `{}`

rossjones: I'm not sure.  It could just as arguably be an empty list, but as both that and {} are valid json I think nil might be a better choice.

PR #161, opened 2019-02-06
Implement allocation happy path

KPobeeNorris:


app/controllers/allocate_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+    @recommended_poms, @not_recommended_poms = pom_response.data.partition { |pom|
+      pom.position_description.include?(@recommended_pom)
+    }
+  end
+
+  def new
+    @prisoner = prisoner
+
+    pom_response = StaffService.new.get_prisoner_offender_managers(caseload)
+    @pom = pom_response.data.select { |p| p.staff_id == nomis_staff_id }.first
```
rossjones: We might want to create a Trello card to track the need for a 'get_pom(staff_id)' entry in the service.

PR #167, opened 2019-02-07
Rework the staff/pom model

rossjones:
Currently we get the in-memory data from the Elite2 API, and have extra
data in our local database (called PrisonOffenderManager).  This gets
confusing very quickly, and doesn't make it easy to work with the data.

This PR does the following:

* Renames PrisonOffenderManager to POMDetail in the database.
* Uses PrisonOffenderManager for the in-memory model
* Adds the POMDetail to the in-memory model


app/services/nomis/elite2/prison_offender_manager.rb
```diff
@@ -29,6 +29,31 @@ class PrisonOffenderManager
...
+        self.tier_c = allocation_counts['C']
+        self.tier_d = allocation_counts['D']
+        self.total_cases = [tier_a, tier_b, tier_c, tier_d].sum
+        self.status = pom_detail.status
+        self.working_pattern = pom_detail.working_pattern
+      end
+
+    private
+
+      def get_allocation_counts(allocations)
```
TobyRet: https://makandracards.com/makandra/31141-ruby-counting-occurrences-of-an-item-in-an-array-enumerable

PR #175, opened 2019-02-11
Makes the allocation summary use a single controller action

rossjones:
Originally this was a single controller action, and then we changed it
to be 3 controller actions (one per tab) so that we could do paging and
breadcrumbs.

This PR changes back to a single action as we will need to obtain ALL of
the offenders in order to be able to add them to the correct tab.

As a result of this change, although pagination is rendered it won't
work particularly well as we will need to re-calculate this for each
newly calculated group.

Breadcrumbs are also a victim of this PR as they are no longer shown
without the different actions.

The point of this PR is so that when we submit the next (grouped offenders according to allocated, unallocated, missing_info) we can just return these three objects without having to re-do the HTML.


app/controllers/allocations_controller.rb
```diff
@@ -2,40 +2,15 @@ class AllocationsController < ApplicationController
...
       caseload,
       page_number: page_number
     )
-    @prisoners = response.data
-    @page_data = response.meta
-  end
 
-  def awaiting
-    breadcrumb 'Awaiting allocation', :allocations_awaiting_path
+    @offenders_allocated = response.data
```
TobyRet: Perhaps reduce to one variable? `@offenders`

rossjones: The point of this PR is so that when we submit the next (grouped offenders according to allocated, unallocated, missing_info) we can just return these three objects without having to re-do the HTML.

app/views/allocations/_allocated.html.erb
```diff
@@ -20,14 +17,14 @@
     </tr>
     </thead>
     <tbody class="govuk-table__body">
-    <% @prisoners.each_with_index  do |prisoner, i| %>
+    <% @offenders_allocated.each_with_index do |offender, i| %>
       <tr class="govuk-table__row offender_row_<%= i %>">
```
TobyRet: Worth extracting to partial?

app/views/allocations/_awaiting.html.erb
```diff
@@ -16,14 +14,14 @@
     </tr>
     </thead>
     <tbody class="govuk-table__body">
-    <% @prisoners.each_with_index  do |prisoner, i| %>
```
TobyRet: Worth extracting to partial?

app/views/allocations/_missing_information.html.erb
```diff
@@ -16,12 +14,12 @@
     </tr>
     </thead>
     <tbody class="govuk-table__body">
-    <% @prisoners.each_with_index  do |prisoner, i| %>
```
TobyRet: Worth extracting to partial?

app/views/allocations/_allocated.html.erb
rossjones: Think this TR is different in each table isn't it?

app/views/allocations/_missing_information.html.erb
rossjones: This has since changed to @ offenders_awaiting_tiering
Needed merging with master.

app/views/allocations/_awaiting.html.erb
rossjones: This has changed to @ offenders_awaiting_allocation as it needed merging with master

rossjones:
I was having doubts about this PR.

It may be that we want to populate the allocated tab with data from our local database (from where we can get the offender identifiers to fetch from elite2), and treat each action as a special case - this would mean we could minimise the amount of work we do if the user does not intend to visit every tab.

However, we are unable to sort at all using this approach and so we need to have a single action where we can prepare all of the data at once.

PR #177, opened 2019-02-11
Perform allocation summarisation

rossjones:
This PR fetches ALL of the offenders for the current prison and
categorises them according to whether they have a tier, or whether they
have been allocated or not.

It required changes to the NOMIS client in order to send a POST request
when trying to retrieve hundreds of release dates.

The general approach is to create a bucket for each group and iterate
through the offenders (which are fetched in batches), allocating them to
the correct bucket.  Each bucket has a capacity, which is the minimum
number of records to be held so that the correct page can be held.  If
we only want page 1, then we only need to hold 10 records, if we want
page 3, then we need to hold 60. Once the capacity is reached, putting
things in the bucket is a NOP.

We currently need to fetch every single offender so that we can
calculate the totals for the tabs.  Without this we could abort the
process early when all of the buckets are full. This would, however,
only give benefits when there are more than 10 in allocated, unallocated
and untiered.  Initially we would have to go through the entire list
becuase they all need a tier, and once all are tiered and allocated we'd
be in the same position as the unallocated list would be empty.

NB: We currently only get tier, we should also store case_allocation
against the OffenderShort model.

We may want to optimise the buckets to keep the last N and throw away
the previous ones as this might waste memory.


app/services/allocation_summary_service.rb
```diff
@@ -0,0 +1,132 @@
+class AllocationSummaryService
+  PAGE_SIZE = 10 # The number of items to show in the view
+  FETCH_SIZE = 500 # How many records to fetch from nomis at a time
+
+  # rubocop:disable Metrics/MethodLength
+  def summary(allocated_page, unallocated_page, missing_info_page, prison)
+    create_buckets(allocated_page, unallocated_page, missing_info_page)
+
+    # We want to store the total number of each item so we can show totals for
```
TobyRet: The use of comments in the class is really helpful :-)
Only to points on this:
- Would be good to pass the initializer params as a hash in order to prevent any possible bugs from modifying the caller
- The test coverage is still 100% but I worried that if we modify this class further down the line, we do not have many tests to tell us if it is working as it should. I definitely don't want to add tests for the sake of tests but given this class is relatively complicated I wonder if testing unhappy paths could be a good idea

PR #178, opened 2019-02-11
Allocation: Override an Prison Offender Manager recommendation

TobyRet:
- This PR allows user to override the recommend Prisoner Offender Manager when allocating to an Offender
- Note: there is current a known issue to be dealt with in another PR. When selecting 'other' as the override reason, the text area box is not appearing (it is in the DOM but hidden)
screen shot 2019-02-11 at 16 29 41


rossjones:
I'm not a fan of the allocate->allocates rename, but LGTM :D

PR #185, opened 2019-02-13
Change override_reason to be an array

KPobeeNorris:
It is possible for staff to choose more than one reason when overriding
the recommended POM type and so we needed to change the override table
from a string to an array.

It also made sense to change the column names in the override and
allocation table from 'override_reason' to 'override_reasons'.

### DIFF
--- a/db/migrate/20190212160351_update_overrides_column_type.rb
+++ b/db/migrate/20190212160351_update_overrides_column_type.rb
@@ -1,0 +1,13 @@
+class UpdateOverridesColumnType < ActiveRecord::Migration[5.2]
+  def up
+    rename_column :overrides, :override_reason, :override_reasons
+    rename_column :allocations, :override_reason, :override_reasons
+    change_column :overrides, :override_reasons, :string, using: "override_reasons::character varying[]"
+  end
+
+  def down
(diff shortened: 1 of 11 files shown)


db/migrate/20190212160351_update_overrides_column_type.rb
```diff
@@ -0,0 +1,13 @@
+class UpdateOverridesColumnType < ActiveRecord::Migration[5.2]
+  def up
+    rename_column :overrides, :override_reasons, :override_reasons
+    rename_column :allocations, :override_reason, :override_reasons
+    change_column :overrides, :override_reasons, :string, using: "override_reasons::character varying[]"
```
TobyRet: Postgres has an array data type - perhaps more appropriate to use this?
See https://blog.arkency.com/2014/10/how-to-start-using-arrays-in-rails-with-postgresql/

KPobeeNorris: I initially tried this, but as we're changing a column that already exists rather than creating a new one the migration failed as was unable to cast automatically.  Therefore we need to use this syntax; which is very similar to the issue we faced in this migration: `update_columns_in_po_ms.rb` where we were trying to change the data type.

PR #186, opened 2019-02-13
Implements view list of poms and view a single POM.

rossjones:
The List of POMs is shown and grouped according to their status
(active/inactive).

The Single POM is shown w/o gender/email as we don't currently know
where to get these.  Functionality is added to the
PrisonOffenderManagerService which allows for the retrieval of a list of
[allocation,offender] but unfortunately in this case the offender is a
SentenceDetail as we are currently unable to get the Offender model from
/locations/description/LEI/inmates due to it not accepting a list of
IDs.

### DIFF
--- a/app/services/nomis/elite2/api.rb
+++ b/app/services/nomis/elite2/api.rb
@@ -10,6 +10,7 @@ module Nomis
         delegate :get_offender, to: :instance
         delegate :get_offender_list, to: :instance
         delegate :get_bulk_release_dates, to: :instance
+        delegate :get_bulk_sentence_details, to: :instance
         delegate :get_offence, to: :instance
         delegate :get_offender, to: :instance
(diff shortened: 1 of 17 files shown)


app/services/nomis/elite2/api.rb
```diff
@@ -102,6 +103,23 @@ def get_bulk_release_dates(offender_ids)
         ApiResponse.new(results)
       end
 
+      def get_bulk_sentence_details(offender_ids)
+        route = '/elite2api/api/offender-sentences'
+
+        # We have to add the empty item to the list because otherwise we only get the
+        # last item. This appears to be a bug in faraday because the Elite2 API works
+        # fine with the same URL when called via Postman.
+        parameters = { 'offenderNo' => offender_ids + [''] }
```
rossjones: Will switch this over to use POST in another PR

PR #188, opened 2019-02-14
Allocation tab cleanup

rossjones:
This PR shows the name of the allocated POM when the user is looking at
the allocated tab in the summary screen.  The tier column has also been
removed and the release date column renamed.

As it was included in the cleanup ticket, once an allocation is complete the user is redirected to the awaiting allocation tab rather than the allocated tab.

### DIFF
--- a/app/services/nomis/elite2/prison_offender_manager.rb
+++ b/app/services/nomis/elite2/prison_offender_manager.rb
@@ -3,7 +3,7 @@ module Nomis
     class PrisonOffenderManager
       include MemoryModel
 
-      attribute :staff_id, :string
+      attribute :staff_id, :integer
       attribute :first_name, :string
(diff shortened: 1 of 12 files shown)


app/services/nomis/elite2/prison_offender_manager.rb
```diff
@@ -3,7 +3,7 @@ module Elite2
     class PrisonOffenderManager
       include MemoryModel
 
-      attribute :staff_id, :string
+      attribute :staff_id, :integer
```
TobyRet: I did not realise the value could be casted  👍

rossjones: Yeah, it was causing issues elsewhere when we were doing comparisons between integer ids and strings.

PR #189, opened 2019-02-14
Manually add tiering and case allocation information for an offender

TobyRet:
- Users can now add case information (tiering and case allocation)for an offender. This is a temporary replacement until the NDelius API is up and running.
- Offenders that do no have case information will appear in the 'Awaiting information' tab on the Allocations summary dashboard


app/views/allocations/_awaiting_information.html.erb
```diff
@@ -0,0 +1,48 @@
...
+      <th class="govuk-table__header sorter-false" scope="col">Action</th>
+    </tr>
+    </thead>
+    <tbody class="govuk-table__body">
+    <% @summary.missing_info_offenders.each_with_index do |offender, i| %>
+      <tr class="govuk-table__row offender_row_<%= i %>">
+        <td aria-label="Prisoner full name" class="govuk-table__cell "><%= offender.full_name %></td>
+        <td aria-label="Prisoner number" class="govuk-table__cell "><%= offender.offender_no %></td>
+
+        <!-- Todo: REFACTOR THIS SECTION TO USE PRESENTERS -->
```
TobyRet: Refactor Friday?

spec/services/allocation_summary_service_spec.rb
```diff
@@ -1,10 +1,12 @@
 require 'rails_helper'
 
 describe AllocationSummaryService do
+  # TODO: - Get feedback on test case
```
TobyRet: @rossjones - want to get your opinion on this - as we are not creating case information for offenders in advance they will all end up in the 'Awaiting info' bucket

app/controllers/case_information_controller.rb
```diff
@@ -0,0 +1,28 @@
...
+  end
+
+  def create
+    CaseInformation.create!(
+      nomis_offender_id: case_information_params[:nomis_offender_id],
+      tier: case_information_params[:tier],
+      case_allocation: case_information_params[:case_allocation]
+    )
+
+    redirect_to allocations_path
```
rossjones: This probably needs to be `allocations_path(anchor:'awaiting-tiering')`

app/views/dashboard/index.html.erb
```diff
@@ -22,9 +22,9 @@
   <div class="govuk-grid-column-one-third">
     <div class="wrapper">
       <h1 class="govuk-heading-s">
-        <%= link_to "Awaiting tiering", allocations_path(anchor: 'awaiting-tiering'), class:"govuk-link" %>
+        <%= link_to "Awaiting information", allocations_path(anchor: 'awaiting-information'), class:"govuk-link" %>
```
rossjones: There may be a few other places where awaiting-tiering is used as the anchor

spec/services/allocation_summary_service_spec.rb
rossjones: Can we create some seed data so that once deployed to staging we still have _some_ in allocated and awaiting allocation?  This will still probably break some tests but at least we can just tweak them

app/controllers/case_information_controller.rb
TobyRet: 👍

spec/services/allocation_summary_service_spec.rb
TobyRet: Yep can definitely add seeding data to staging - 
We can seed the Test db also

rossjones: 🆒

PR #192, opened 2019-02-15
Remove pagination text when only 1 page

rossjones:
When there is only one page of results in the allocation summary, there
is little point in showing the Next/Previous and page number.  This PR
removes them.


app/views/shared/_pagination.html.erb
```diff
@@ -1,6 +1,7 @@
 <% if page_data %>
 <div class="govuk-grid-row pagination">
   <div class="govuk-grid-column-one-half">
+    <% if page_data.page_count > 1 %>
```
TobyRet: Possibly another candidate for a presenter (refactor Friday)

PR #193, opened 2019-02-15
Fix case information

TobyRet:
- CRC form field now submitting correctly
- Increased test coverage - assert a prisoner with case information appears in the 'awaiting allocation' tab

### DIFF
--- a/app/views/case_information/new.html.erb
+++ b/app/views/case_information/new.html.erb
@@ -19,8 +19,8 @@
           <%= label_tag "case_information[case_allocation]", "National Probation Service (NPS)", class: "govuk-label govuk-radios__label" %>
         </div>
         <div class="govuk-radios__item">
-          <%= radio_button_tag("case_information[allocation]", "CRC", false, class: "govuk-radios__input") %>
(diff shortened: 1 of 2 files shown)


app/views/case_information/new.html.erb
```diff
@@ -19,8 +19,8 @@
           <%= label_tag "case_information[case_allocation]", "National Probation Service (NPS)", class: "govuk-label govuk-radios__label" %>
         </div>
         <div class="govuk-radios__item">
-          <%= radio_button_tag("case_information[allocation]", "CRC", false, class: "govuk-radios__input") %>
-          <%= label_tag "case_information[allocation]", "Community Rehabilitation Company (CRC)", class: "govuk-label govuk-radios__label" %>
+          <%= radio_button_tag("case_information[case_allocation]", "CRC", false, class: "govuk-radios__input") %>
```
rossjones: I'm unsure how this fixes the problem with Tier C & D ...

TobyRet: The controller is expecting a `case_allocation` field not `allocation`

PR #194, opened 2019-02-15
POM view of own caseload

KPobeeNorris:
This feature allows signed in POMs to view all their allocated cases, and specifically those newly allocated cases from the past 7 days.

### DIFF
--- a/app/controllers/dashboard_controller.rb
+++ b/app/controllers/dashboard_controller.rb
@@ -1,5 +1,9 @@
 class DashboardController < ApplicationController
   before_action :authenticate_user
 
-  def index; end
+  def index
+    user = Nomis::Elite2::Api.fetch_nomis_user_details(current_user).data
+    poms_list = PrisonOffenderManagerService.get_poms(caseload)
+    @pom = poms_list.select { |p| p.staff_id.to_i == user.staff_id.to_i }.first
+  end
 end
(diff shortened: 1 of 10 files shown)


app/controllers/dashboard_controller.rb
```diff
@@ -1,5 +1,9 @@
 class DashboardController < ApplicationController
   before_action :authenticate_user
 
-  def index; end
+  def index
+    user = Nomis::Elite2::Api.fetch_nomis_user_details(current_user).data
```
rossjones: Should this be in a Service?

app/controllers/poms_controller.rb
```diff
@@ -19,6 +22,22 @@ def show
 
   def edit; end
 
+  def my_caseload
+    user = Nomis::Elite2::Api.fetch_nomis_user_details(current_user).data
```
rossjones: Probably in a service like the others, and then that call wrapped in a method called user?

app/controllers/dashboard_controller.rb
TobyRet: Yep agree!

app/controllers/poms_controller.rb
```diff
@@ -19,10 +22,27 @@ def show
...
+
 private
 
   def pom
     poms_list = PrisonOffenderManagerService.get_poms(caseload)
     @pom = poms_list.select { |p| p.staff_id.to_i == params['id'].to_i }.first
   end
+
+  def signed_in_pom_details
+    user = Nomis::Elite2::Api.fetch_nomis_user_details(current_user).data
```
TobyRet: Same here

PR #203, opened 2019-02-18
Do not show inactive POMs on allocation page

rossjones:
When allocating a prisoner to a POM, we currently show all POMs
regardless of status. This means it is possible to allocate a prisoner
to a POM who is not available.

This PR changes the allocations_controller to only show active POMs.  To
achieve this the `get_poms` call can now take a filtering block that
will allow for get_poms to return just the POMs that caller is
interested in.  In this case, we filter with

```ruby
get_poms('LEI') { |pom|
   pom.status == 'active'
}
```

Currently when we get_pom_detail, we find or create a PomDetail and set
the status to 'inactive' if we create a POM.  This is a stopgap, for
development purposes and will eventually be removed to not create a
PomDetail when not found.  In the meantime, we've changed the default status
to 'active' rather than 'inactive'

### DIFF
--- a/app/services/prison_offender_manager_service.rb
+++ b/app/services/prison_offender_manager_service.rb
@@ -2,17 +2,24 @@ class PrisonOffenderManagerService
   def self.get_pom_detail(nomis_staff_id)
     PomDetail.find_or_create_by!(nomis_staff_id: nomis_staff_id.to_i) { |s|
       s.working_pattern = s.working_pattern || 0.0
-      s.status = s.status || 'inactive'
+      s.status = s.status || 'active'
     }
   end
 
-  def self.get_poms(prison)
+  def self.get_poms(prison, &filter)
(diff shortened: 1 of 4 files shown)


app/services/prison_offender_manager_service.rb
```diff
@@ -2,17 +2,24 @@ class PrisonOffenderManagerService
...
-    poms.map { |pom|
+
+    poms = poms.map { |pom|
       detail = get_pom_detail(pom.staff_id)
       pom.add_detail(detail)
       pom
-    }
+    }.compact
+
+    if filter
```
TobyRet: could also do `if block_given?` if you want to do away with `&filter` param

rossjones: Wanted to name it to be explicit, so that you've an idea of what it does.  block_given? could be anything and I'm not sure it'd be at all obvious.  

Happy to change it though if you think it's more idiomatic though @TobyRet

PR #207, opened 2019-02-18
Refactor routes

TobyRet:
- follow Rails convention and use `resources` where appropriate
- This is a horrible PR to review sorry (Just trust me, he he) - recommend looking at `config/routes.rb` then the rest should make sense


app/views/allocations/_pom_tables.html.erb
```diff
@@ -30,13 +30,13 @@ Recommendation: <%= @recommended_pom %> POMs
...
       <td aria-label="Working pattern" class="govuk-table__cell"><%= pom.working_pattern %></td>
-      <% if current_page?(action: 'show') %>
-        <td aria-label="Action" class="govuk-table__cell table_cell__left_align"><a
-          <%= link_to "Allocate", new_allocations_path(@prisoner.offender_no, pom.staff_id), class:"govuk-link"%></a></td>
-      <% else %>
-        <td aria-label="Action" class="govuk-table__cell table_cell__right_align"><a
-          href="#">Reallocate</a></td>
-      <% end %>
+      <td aria-label="Action" class="govuk-table__cell table_cell__left_align">
+        <% if current_page?(action: :new) %>
```
TobyRet: Currently allocating and re-allocating use the same controller methods so this if block does not currently work. Will fix in next PR

PR #209, opened 2019-02-19
Edit a POM's working pattern/status

KPobeeNorris:
This feature allows a staff to update a POM's working pattern and status.  This involves an edit form with radio buttons, and these buttons are pre-selected with the POM's current details.

There is an option to make a POM inactive, and if selected this de-allocates all the POM's cases so they can be allocated to another POM.  A warning is displayed to avoid this option being chosen without the person understanding this behaviour.

### DIFF
--- a/app/services/prison_offender_manager_service.rb
+++ b/app/services/prison_offender_manager_service.rb
@@ -61,7 +61,7 @@ class PrisonOffenderManagerService
     @pom = poms_list.select { |p| p.staff_id.to_i == user.staff_id.to_i }.first
   end
 
-  def self.edit_a_pom(params)
+  def self.update_pom(params)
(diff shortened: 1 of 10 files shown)


app/services/prison_offender_manager_service.rb
```diff
@@ -55,4 +55,19 @@ def self.get_signed_in_pom_details(current_user)
     poms_list = PrisonOffenderManagerService.get_poms(user.active_case_load_id)
     @pom = poms_list.select { |p| p.staff_id.to_i == user.staff_id.to_i }.first
   end
+
+  def self.edit_a_pom(params)
```
rossjones: update_pom?

PR #210, opened 2019-02-19
Dynamically change link label depending on whether user is allocating or reallocating

TobyRet:
When a user is allocating we want relevant links to be labelled 'Allocate', when a user is reallocating then we want the label to be 'Reallocate'. The current pom details now also displays on `edit_allocation_path`.

One yucky aspect of this is that the PR introduces a new `edit` action on the controller in order to differentiate the  behaviour. We are not really updating a resource - a new allocation event is created every time.

### DIFF
--- a/app/controllers/allocations_controller.rb
+++ b/app/controllers/allocations_controller.rb
@@ -4,20 +4,20 @@ class AllocationsController < ApplicationController
   alias_action :new, :edit
 
   def new
-    @prisoner = get_prisoner(nomis_offender_id_from_url)
+    @prisoner = prisoner(nomis_offender_id_from_url)
     @recommended_pom = @prisoner.current_responsibility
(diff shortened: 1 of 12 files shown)


app/controllers/allocations_controller.rb
```diff
@@ -1,30 +1,23 @@
 class AllocationsController < ApplicationController
   before_action :authenticate_user
 
+  alias_action :new, :edit
+
   def new
-    @prisoner = OffenderService.new.get_offender(nomis_offender_id_from_url)
+    @prisoner = get_prisoner(nomis_offender_id_from_url)
     @recommended_pom = @prisoner.current_responsibility
```
rossjones: Is this actually recommended_pom_type?

TobyRet: Yeah I think this should be renamed - will change

PR #216, opened 2019-02-20
Adds some error pages for 401, 404, and 500.

rossjones:
Adds new error pages (and a controller) for handling 401, 404 and 500
errors within the app.

Requires content review and changes, but should be good to go. Because
they are dynamic templates, we have to be careful not to raise any
errors during rendering or the computer will blow up.


config/environments/test.rb
```diff
@@ -19,7 +19,7 @@
   }
 
   # Show full error reports and disable caching.
-  config.consider_all_requests_local       = true
+  config.consider_all_requests_local       = false
```
KPobeeNorris: @rossjones - set back to false?

PR #225, opened 2019-02-21
Add optional text field to allocation confirmation

KPobeeNorris:
This is an add-on to the current allocation feature and allows whoever is
doing the allocating to send an optional message to the POM.  At present
this doesn't have a word limit, but this can be added if required.  The
Notify template has been updated to include this message option and it
only displays within the email if a message is present.


app/mailers/pom_mailer.rb
```diff
@@ -1,20 +1,24 @@
...
       email_subject: 'New OMIC allocation',
       pom_name: pom.first_name.capitalize,
       offender_name: offender.full_name,
-      nomis_offender_id: offender.offender_no
+      nomis_offender_id: offender.offender_no,
+      message: message
     )
 
-    mail(to: '')
+    mail(to: 'kath.pobee-norris@digital.justice.gov.uk')
```
rossjones: Did you mean to leave this here?

app/services/email_service.rb
```diff
@@ -3,15 +3,16 @@ class EmailService
   def self.send_allocation_email(params)
     offender = OffenderService.get_offender(params[:nomis_offender_id])
     pom = PrisonOffenderManagerService.get_pom(params[:prison], params[:nomis_staff_id])
-    last_allocation = Allocation.where(nomis_offender_id:  params[:nomis_offender_id]).
-        where(active: false).last
+    # last_allocation = Allocation.where(nomis_offender_id:  params[:nomis_offender_id]).
```
rossjones: Is this no longer necessary?

PR #230, opened 2019-02-22
Refactors the API

rossjones:
In an attempt to be more explicit about how our API client works, this PR attempts to break apart the `Nomis::Elite2::Api` class into separate classes that each handles on type of resource.

For instance, the `Nomis::Elite2::Api. get_offender_list` is now available in `Nomis::Api::OffenderApi.list`. Each of the new *_Api classes extends `Nomis::Api::Elite2Api` so that they each have access to an E2 client (`e2_client()`) and and api deserializer.  By using extend we can keep all of the methods at class level in the Apis, but still create a client (from client.rb) on demand.  This should make it easier in future to check out a client from a pool of clients in future.

Tests are separated out to test each *Api class independently.


rossjones:
@KPobeeNorris @TobyRet If possible could you both review this one as it's a bit of a big change.

app/services/nomis/api/offender_api.rb
```diff
@@ -0,0 +1,81 @@
...
+        }
+
+        offenders = data.map { |offender|
+          api_deserialiser.deserialise(Nomis::Models::OffenderShort, offender)
+        }
+
+        ApiPaginatedResponse.new(page_meta, offenders)
+      end
+
+      def self.get(offender_no)
```
KPobeeNorris: Wondering if this could be a bit more explicit e.g. "get_offender"? Someone new to the code might not find it immediately obvious.  This is only tiny though, so feel free to ignore

rossjones: it was originally `get_offender` so no objections to changing it back :)

app/services/movement_service.rb
```diff
@@ -1,7 +1,7 @@
 class MovementService
   # rubocop:disable Metrics/MethodLength
   def self.movements_on(date, direction_filters: [], type_filters: [])
-    movements = Nomis::Elite2::Api.movements_on_date(date)
+    movements = Nomis::Api::MovementApi.movements_on_date(date)
```
TobyRet: I think the split makes sense, but rather than have `Nomis::Api::MovementApi` I think it's worth being more explicit about the top level API that is being used i.e `Nomis::Elite2::MovementApi` so we know where we are accessing the API from.

app/services/nomis/api/elite2_api.rb
```diff
@@ -0,0 +1,16 @@
+module Nomis
+  module Api
```
TobyRet: `module API` seems redundant

app/services/nomis/api/elite2_api.rb
```diff
@@ -0,0 +1,16 @@
...
+    ApiPaginatedResponse = Struct.new(:meta, :data)
+
+    module Elite2Api
```
TobyRet: rename as 'module Elite2` - suffixing API seems redundant

app/services/nomis/api/elite2_api.rb
```diff
@@ -0,0 +1,16 @@
...
+      def e2_client
+        host = Rails.configuration.nomis_oauth_host
+        Nomis::Client.new(host)
```
TobyRet: `Nomis::Elite2::Client.new(host)` ?

rossjones:
@TobyRet this should be fixed now

PR #234, opened 2019-02-22
Allow staff without POM role to access the service

KPobeeNorris:
Following the implementation of the 'my caseload' and 'new allocations'
pages there was an issue that only staff with the POM role could access
all of the application.  This PR removes the reliance on people needing
to have a POM role and if they click on the aforementioned pages they
will be presented with a message confirming they don't have any
allocated cases.

### DIFF
--- a/app/controllers/poms_controller.rb
+++ b/app/controllers/poms_controller.rb
@@ -38,13 +38,15 @@ class PomsController < ApplicationController
 
   def my_caseload
     @pom = PrisonOffenderManagerService.get_signed_in_pom_details(current_user)
-    @allocations = PrisonOffenderManagerService.get_allocated_offenders(@pom.staff_id)
-    @new_cases = PrisonOffenderManagerService.get_new_cases(@pom.staff_id)
+    if @pom.present?
+      @allocations = PrisonOffenderManagerService.get_allocated_offenders(@pom.staff_id)
+      @new_cases = PrisonOffenderManagerService.get_new_cases(@pom.staff_id)
+    end
   end
 
   def new_cases
     @pom = PrisonOffenderManagerService.get_signed_in_pom_details(current_user)
-    @new_cases = PrisonOffenderManagerService.get_new_cases(@pom.staff_id)
+    @new_cases = PrisonOffenderManagerService.get_new_cases(@pom.staff_id) if @pom.present?
   end
 
 private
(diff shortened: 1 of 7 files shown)


app/views/dashboard/index.html.erb
```diff
@@ -35,15 +35,15 @@
   <div class="govuk-grid-column-one-third">
     <div class="wrapper">
       <h1 class="govuk-heading-s">
-        <%= link_to "My caseload", my_caseload_path(@pom.staff_id), class:"govuk-link" %>
+        <%= link_to "My caseload", my_caseload_path(@user.staff_id), class:"govuk-link" %>
```
rossjones: The path doesn't need to take the staff_id here, it isn't used in the controller.

app/views/dashboard/index.html.erb
```diff
@@ -35,15 +35,15 @@
...
       </h1>
       <p class="govuk-body">All cases allocated to you.</p>
     </div>
   </div>
   <div class="govuk-grid-column-one-third">
     <div class="wrapper">
       <h1 class="govuk-heading-s">
-        <%= link_to "New allocations", new_cases_path(@pom.staff_id), class:"govuk-link" %>
+        <%= link_to "New allocations", new_cases_path(@user.staff_id), class:"govuk-link" %>
```
rossjones: Same here, no staff_id required

spec/features/view_allocations_spec.rb
```diff
@@ -33,4 +33,10 @@
     expect(page).to have_content("New allocations")
     expect(page).to have_content("Abbella, Ozullirn")
   end
+
+  it 'allows staff without the POM role to view the my casload page', vcr: { cassette_name: :non_pom_my_caseload }  do
+    signin_user('NON_POM_GEN')
+    visit "/poms/485767/my_caseload"
```
rossjones: Staff id isn't necessary.  Will need to remove this from the routes in routes.rb for my_caseload and new_cases.

app/controllers/poms_controller.rb
```diff
@@ -38,13 +38,17 @@ def update
 
   def my_caseload
     @pom = PrisonOffenderManagerService.get_signed_in_pom_details(current_user)
-    @allocations = PrisonOffenderManagerService.get_allocated_offenders(@pom.staff_id)
-    @new_cases = PrisonOffenderManagerService.get_new_cases(@pom.staff_id)
+    if @pom.present?
```
rossjones: The breadcrumbs in this file (for my_caseload and new_cases both have a hard-coded id of 1, but this can be removed as the ID isn't needed).

PR #266, opened 2019-03-07
Store a user's caseloads in the session

TobyRet:
- user details are now fetched via the CustodyApi instead of the Elite2API
- a user's caseloads are now stores in the session


app/services/nomis/custody/user_api.rb
```diff
@@ -1,6 +1,6 @@
 module Nomis
   module Custody
-    class StaffApi
+    class UserApi
       extend CustodyApi
 
       def self.list_caseloads(username)
```
rossjones: Now sure we'll need this anymore.

TobyRet: I have removed

PR #268, opened 2019-03-08
Initial search UI

rossjones:
Currently broken because when we add a queryparam to the first tab (when
say pending is selected) clicking on that tab breaks everything.


rossjones:
Closing as this is probably too early, and not the correct approach - searching will be via the dashboard.

PR #271, opened 2019-03-08
Retrieve staff emails

KPobeeNorris:


app/services/nomis/elite2/user_api.rb
```diff
@@ -0,0 +1,13 @@
+module Nomis
+  module Elite2
+    class UserApi
+      extend Elite2Api
+
+      def self.fetch_email_addresses(nomis_staff_id)
+        route = "/elite2api/api/staff/#{nomis_staff_id}/emails"
+        emails = e2_client.get(route)
+        emails
```
rossjones: The assignment is unnecessary, you can remove emails and emails= ... it'll return the last thing in the function so just ```
route = 
e2_client.get...
```
should be enough

File: app/services/nomis/models/user_details.rb [hunk 2]
```diff
@@ -13,6 +13,7 @@ class UserDetails
       attribute :status, :string
       attribute :thumbnail_id, :string
       attribute :username, :string
+      attribute :emails, :string
```
rossjones: If you don't specify a type, I think it may even make it a list for you automatically

PR #273, opened 2019-03-11
[DO NOT MERGE] Processes movement events from the movements API

rossjones:
We expect that we need to handle movements of offenders, both between
prisons and as a result of being released.

This PR is expected to be run nightly and will process the list of
movement events for the previous 24 hours.

Currently there is a hard-coded date in the rake task as we do not have
a way of creating movements on demand for testing/development purposes.
We also need some transfers to be created in legacy nomis for
dev/testing to ensure the code is correct.

The events received are:

Prisoner leaves a prison: TRN-OUT (fromAgency -> toAgency) - a prisoner who is in transit will always show as a  TRN-OUT.

Prisoner arrives: ADM-IN (fromAgency->toAgency) 

What isn't 100% clear, is how we tell an ADM is a new offender, or the result of a TRANSFER.  For now, we can only assume that if fromAgency is set that it is a transfer.


rossjones:
Not 100% sure what happened here, going to try again with a new branch on top of master

PR #279, opened 2019-03-11
Content 11 03

rossjones:
All of the content changes for w/c 11/03/19.


app/controllers/overrides_controller.rb
```diff
@@ -21,13 +23,22 @@ def create
...
       active_caseload, override_params[:nomis_staff_id])
+    @complex_label = complex_reason_label
 
     render :new
   end
 # rubocop:enable Metrics/MethodLength
 
 private
 
+  def complex_reason_label
```
TobyRet: Probably a Fixit Friday thing but possibly a bit strange that the controller has view logic mixed in?

app/views/dashboard/index.html.erb
```diff
@@ -1,64 +1,64 @@
...
 
-<h2 class="govuk-heading-m no-bottom-margin">Staff</h2>
+<h2 class="govuk-heading-m no-bottom-margin">Manage staff</h2>
 <hr class="govuk-section-break govuk-section-break--visible">
 <div class="govuk-grid-row dashboard-row">
   <div class="govuk-grid-column-one-third">
     <div class="wrapper">
       <h1 class="govuk-heading-s">
-        <%= link_to "Active POMs", poms_path(), class:"govuk-link" %>
+        <%= link_to "View all offender managers", poms_path(), class:"govuk-link" %>
```
TobyRet: Personally I prefer using 'offender managers' instead of POMs as it is more explicit - but I notice we are still referring to 'POM' in many places still. Is the intention to be consistent does it depend on the context / page?

app/controllers/overrides_controller.rb
rossjones: Yeah, I guess this might be a presenter thing?  It's because this value is used in an `input_tag` so it seemed as messy to try and have a giant if there.  I'll add a fixit friday ticket...

app/views/dashboard/index.html.erb
rossjones: Not sure, but we've had content updates using POM.

PR #281, opened 2019-03-11
Handle 204 responses from API calls

rossjones:
Elite2, in places, uses 204 (no content) to mean that there was
nothing to return, but that the call succeeded. This means however we do
not know whether the intended type of the response was [] or {}.  As it
would be dangerous for us to guess, instead we will return nil from
Client#get and leave it up to the caller to determine the correct type.

### DIFF
--- a/app/services/nomis/client.rb
+++ b/app/services/nomis/client.rb
@@ -18,11 +18,20 @@ module Nomis
       end
     end
 
+    # rubocop:disable Metrics/MethodLength
     def get(route, queryparams: {}, extra_headers: {})
       response = request(
         :get, route, queryparams: queryparams, extra_headers: extra_headers
       )
-      data = JSON.parse(response.body)
+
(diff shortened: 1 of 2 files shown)


app/services/nomis/client.rb
```diff
@@ -18,11 +18,20 @@ def initialize(root)
...
       response = request(
         :get, route, queryparams: queryparams, extra_headers: extra_headers
       )
-      data = JSON.parse(response.body)
+
+      # Elite2 can return a 204 to mean empty results, and we don't know if
+      # it is meant to be a {} or a []. For now, we are going to use nil and
+      # let the caller handle it.
+      if response.status == 204
+        data = nil
```
TobyRet: Wonder if it makes life easier just returning [] so you don't need to modify callers

rossjones: That's what I did originally, but we don't know if it's a [] or {} :(

PR #282, opened 2019-03-12
Add Faraday Instrumentation middleware to log API calls

TobyRet:
Temporary measure until we have time to use (learn) Prometheus queries

### DIFF
--- a/config/initializers/notifications.rb
+++ b/config/initializers/notifications.rb
@@ -1,0 +1,10 @@
+# rubocop:disable Metrics/LineLength
+# rubocop:disable Style/FormatStringToken
+ActiveSupport::Notifications.subscribe('request.faraday') do |_name, start_time, end_time, _id, env|
+  url = env[:url]
+  http_method = env[:method].to_s.upcase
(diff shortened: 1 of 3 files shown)


config/initializers/notifications.rb
```diff
@@ -0,0 +1,10 @@
+# rubocop:disable Metrics/LineLength
+# rubocop:disable Style/FormatStringToken
+ActiveSupport::Notifications.subscribe('request.faraday') do |_name, start_time, end_time, _id, env|
+  url = env[:url]
+  http_method = env[:method].to_s.upcase
+  duration = end_time - start_time
+  STDOUT.puts '[%s] %s %s (%.3f s)' % [url.host, http_method, url.request_uri, duration]
```
rossjones: Did rubocop not complain about not using ""?  Also might it best to use the rails logger if possible?

TobyRet: Updated to use Rails logger - no complaints from Rubocop about single quotes - maybe it's ok for this type of string formatting?

PR #284, opened 2019-03-12
Add Staging integration tests to CircleCI workflow

TobyRet:
- Staging integration tests are run once staging is deployed
- CircleCi does this by cloning the integration tests repo, installing dependencies and running the tests
- Appreciate any feedback on the config - I feel it's getting a little unwieldy and there are perhaps better ways to speed up the build


.circleci/config.yml
```diff
@@ -216,6 +216,51 @@ jobs:
           environment:
             <<: *github_team_name_slug
 
+  test_staging:
+    <<: *test_container_config
+    working_directory: ~/staging
+    steps:
+      - checkout
```
TobyRet: It's possibly a little strange to be checking out at this point (this command does a number of things including cloning the repo that the job is configured to build i.e offender-allocation-manager). You will see that I clone the integration tests repo into this folder. The reason for using `checkout` is that you get other functionality thrown in including adding Github to known-hosts. If we did not do this we would have to add Github to known hosts manually. Happy to change if this feels to messy

rossjones: Perfectly fine to be checking out other repos though.

PR #285, opened 2019-03-13
Enable Notify by fetching staff email 

KPobeeNorris:
We previously added functionality to send email notifications to POMs using Notify, however, there was a delay to our ability to access these email addresses and so we temporarily switched this off.  We are now able to get POM email addresses and so this PR enables the PomMailer to be able to send emails.  It also refactors the code related to fetching POM email addresses as we don't need to get them until further downstream.


app/mailers/pom_mailer.rb
```diff
@@ -1,45 +1,38 @@
-# :nocov:
 class PomMailer < GovukNotifyRails::Mailer
-  # TODO: Add POM email addresses
   # rubocop:disable Metrics/MethodLength
-  def new_allocation_email(pom, offender, message)
-    return unless active?
-
+  #  rubocop:disable Metrics/ParameterLists
+  def new_allocation_email(pom_name, pom_email, offender_name, offender_no, message, url)
```
TobyRet: Convention is to pass params as a hash if you have more than 2-3 params

app/mailers/pom_mailer.rb
```diff
@@ -1,45 +1,38 @@
...
     message = "Additional information: #{message}" if message.present?
-    user = ''
     set_template('9679ea4c-1495-4fa6-a00b-630de715e315')
     set_personalisation(
       email_subject: 'New OMIC allocation',
```
TobyRet: ... you could then merge `email_subject` with your params hash

app/services/email_service.rb
```diff
@@ -1,20 +1,38 @@
...
+        pom.full_name,
+        offender.full_name,
+        offender.offender_no,
+        PrisonService.name_for(pom.agency_id),
+        url
+      ).deliver_later
     end
-    PomMailer.new_allocation_email(pom, offender, message).deliver_now
+
+    PomMailer.new_allocation_email(
```
TobyRet: Worth refactoring duplication of params?

app/services/nomis/models/prison_offender_manager.rb
```diff
@@ -17,6 +17,7 @@ class PrisonOffenderManager
       attribute :schedule_type_description, :string
       attribute :hours_per_week, :integer
       attribute :thumbnail_id, :string
+      attribute :emails
```
TobyRet: Is a type required?

config/environments/production.rb
```diff
@@ -12,6 +12,9 @@
   config.i18n.fallbacks = true
   config.active_support.deprecation = :notify
 
+  # TODO: Add live notify API key once our account goes live
+  config.notify_api_key = ENV['DEV_NOTIFY_API_KEY']
```
TobyRet: Common env vars could go in `application.rb` (DEV_NOTIFY_API_KEY)

config/environments/development.rb
```diff
@@ -28,10 +28,9 @@
   end
 
   # Don't care if the mailer can't send.
-  config.action_mailer.raise_delivery_errors = false
-
+  config.notify_api_key = ENV['DEV_NOTIFY_API_KEY']
```
TobyRet: Common env vars could go in application.rb (DEV_NOTIFY_API_KEY)

app/services/nomis/models/prison_offender_manager.rb
rossjones: I think it behaves better without as it determines the type by what you provide it, rather than trying to force it into a string/integer etc.

PR #286, opened 2019-03-13
Implements search

rossjones:
Adds a search controller and a search service that allows the user to
search for offenders by providing a string that is matched with their
first name, last name or offender number.

Currently we do this in memory because of missing functionality in
Elite2 but will often be working with offenders from the cache.  There
is no relevance as we are doing partial matches and not searching in a
corpus of documents so we can't use td/idf to determine relevance at
all.

### DIFF
--- a/app/controllers/search_controller.rb
+++ b/app/controllers/search_controller.rb
@@ -15,14 +15,17 @@ class SearchController < ApplicationController
 
 private
 
-  # At this point @offenders contains ALL of the offenders at the prison that
-  # match the search term. We will only show 10 at a time, so there is no need
-  # to get the allcocated POM name for all of them.
   def get_slice_for_page(offenders, page_number)
     start = [(page_number - 1) * 10, 0].max
     slice = offenders.slice(start, 10)
 
+    # At this point offenders contains ALL of the offenders at the prison that
+    # match the search term, slice is the current page worth of offenders.
+    # We will only show 10 at a time, so there is no need
+    # to get the allocated POM name for offenders, we will just get them
+    # for the much smaller slice.
     OffenderService.set_allocated_pom_name(slice, active_caseload)
+
     slice
   end
 
--- a/app/controllers/search_controller.rb
+++ b/app/controllers/search_controller.rb
@@ -27,8 +27,6 @@ private
     # to get the allocated POM name for offenders, we will just get them
     # for the much smaller slice.
     OffenderService.set_allocated_pom_name(slice, active_caseload)
-
-    slice
   end
 
   def create_page_meta(total_records, current_view)
--- a/app/controllers/search_controller.rb
+++ b/app/controllers/search_controller.rb
@@ -3,6 +3,8 @@ class SearchController < ApplicationController
 
   breadcrumb 'Search', :search_path, only: [:search]
 
+  PAGE_SIZE = 10
+
   def search
     @q = search_term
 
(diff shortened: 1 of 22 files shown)


app/controllers/search_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+private
+
+  # At this point @offenders contains ALL of the offenders at the prison that
+  # match the search term. We will only show 10 at a time, so there is no need
+  # to get the allcocated POM name for all of them.
+  def get_slice_for_page(offenders, page_number)
+    start = [(page_number - 1) * 10, 0].max
+    slice = offenders.slice(start, 10)
+
+    OffenderService.set_allocated_pom_name(slice, active_caseload)
```
TobyRet: Sorry I don't get why this line here - can you explain?

app/helpers/application_helper.rb
```diff
@@ -15,6 +15,12 @@ def format_date(date_obj)
     date_obj.strftime('%d/%m/%Y')
   end
 
+  def format_date_string(date_string)
+    return '' if date_string.nil?
```
TobyRet: What type is `date_string`?

rossjones: A string in a datelike-format

app/controllers/search_controller.rb
rossjones: We need to set the POM name for the offenders, but rather than do it for all 600, we just do it for the 10 we are about to show on the current page.

rossjones: Will add a comment

app/services/search_service.rb
```diff
@@ -0,0 +1,46 @@
...
+    tier_map = CaseInformationService.get_case_information(prison)
+
+    search_term = text.upcase
+    search_results = []
+
+    (0..number_of_requests).each do |request_no|
+      offenders = OffenderService.get_offenders_for_prison(
+        prison,
+        page_number: request_no,
+        page_size: 250,
```
TobyRet: Could be worth assigning page_size to a constant as it is used further down

app/services/search_service.rb
```diff
@@ -0,0 +1,46 @@
...
+        tier_map: tier_map
+      )
+      break if offenders.blank?
+
+      new_offenders = offenders.select { |offender|
+        offender.last_name.start_with?(search_term) ||
```
TobyRet: Does the search term need to be an exact match for the name?

rossjones: Yeah, I think they need to know the start of either the first/last name to be able to find them.

app/helpers/application_helper.rb
TobyRet: Could `date_string` ever be empty? e.g `''` If so, this method will blow up

app/services/search_service.rb
TobyRet: Does Jay and the design crew know that this is not a fuzzy search?

app/helpers/application_helper.rb
rossjones: Doh. Switched to .empty?

app/services/search_service.rb
rossjones: Yes, we also had a conversation about relevance, "disappointed but resigned" is the best description of the response.

app/helpers/application_helper.rb
TobyRet: I guess you may need to check for nil and empty?

app/controllers/search_controller.rb
rossjones: Yeah, nastily set_allocated_pom_name will edit in place, and then return the things it's just modified.  I think it's working because the list (ref) that is passed is having it's contents mutated.  Fixed in latest commit.

PR #290, opened 2019-03-14
Display POM responsibility

TobyRet:
- Display a POM's responsibility in relation to an allocated offender. 
- Uses rules to determine whether the responsibility is 'Responsible' or 'Supporting' ...
POM is 'Responsible' if offender is 'Welsh' and release date is greater than 10 months, else 'Supporting'

### DIFF
--- a/app/controllers/allocations_controller.rb
+++ b/app/controllers/allocations_controller.rb
@@ -37,6 +37,7 @@ class AllocationsController < ApplicationController
       prison: active_caseload,
       override_reasons: override_reasons,
       override_detail: override_detail,
+      responsibility: ResponsibilityService.calculate_pom_responsibility(prisoner),
       message: allocation_params[:message]
     )
 
@@ -69,7 +70,7 @@ private
     }
 
     pom_response.partition { |pom|
-      pom.position_description.include?(prisoner.current_responsibility)
+      pom.position_description.include?(prisoner.case_owner)
     }
   end
 
--- a/app/models/allocation.rb
+++ b/app/models/allocation.rb
@@ -6,5 +6,6 @@ class Allocation < ApplicationRecord
     :nomis_booking_id,
     :prison,
     :allocated_at_tier,
+    :responsibility,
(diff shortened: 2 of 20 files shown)


app/controllers/allocations_controller.rb
```diff
@@ -69,7 +70,7 @@ def recommended_and_nonrecommended_poms_for(prisoner)
     }
 
     pom_response.partition { |pom|
-      pom.position_description.include?(prisoner.current_responsibility)
+      pom.position_description.include?(prisoner.case_owner)
```
rossjones: Is case_owner in this context Custody/Community?

app/models/allocation.rb
```diff
@@ -6,5 +6,6 @@ class Allocation < ApplicationRecord
     :nomis_booking_id,
     :prison,
     :allocated_at_tier,
+    :responsibility,
```
rossjones: I think the problem with storing it in the allocation, is that it was the responsibility at the time of the allocation, so the 10.months problem won't be reflected automatically (because we're always recalculating based on the current date).  This might be okay, but it depends on whether the expectation is that the responsibility will automatically change on a specific day.

Maybe it's okay for now, but we should raise a ticket to clarify as part of the primary/secondary allocation work we will need to do.

app/controllers/allocations_controller.rb
TobyRet: Yes that's right - it mirrors the language used here - https://allocation-manager-staging.apps.cloud-platform-live-0.k8s.integration.dsd.io/allocations/edit/G4273GI

I thought by doing this it would be be consistent with NOMIS terminology and make it easier to differentiate against POM responsibility. Just a proposal but happy to reconsider

app/models/allocation.rb
TobyRet: Yes good point - I guess it would be good to know whether a new allocation is required if an Offenders sentence hits the < 10 month mark?

PR #291, opened 2019-03-14
Replaces APICache with Rails.cache

rossjones:
Switches to using Rails.cache instead of APICache which exhibited weird
behaviour.

You can turn on caching in dev with:

```
touch tmp/caching-dev.txt  # from project root
```

and disable with

```
rm tmp/caching-dev.txt
```


Gemfile.lock
```diff
@@ -346,4 +342,4 @@ RUBY VERSION
    ruby 2.5.3p105
 
 BUNDLED WITH
-   1.17.2
```
TobyRet: Bundler version has been degraded

PR #293, opened 2019-03-14
Revert "Revert "Enable Notify by fetching staff email ""

KPobeeNorris:
This reverts commit 05a3e610b881496496ab153f00633fbcebb8e622.


app/services/email_service.rb
```diff
@@ -1,20 +1,38 @@
...
     offender = OffenderService.get_offender(params[:nomis_offender_id])
     pom = PrisonOffenderManagerService.get_pom(params[:prison], params[:nomis_staff_id])
-    last_allocation = Allocation.where(nomis_offender_id:  params[:nomis_offender_id]).
-        where(active: false).last
+    last_allocation = Allocation.where(
+      nomis_offender_id: params[:nomis_offender_id],
+      active: false
+    ).last
     message = params[:message]
+    url = Rails.configuration.allocation_manager_host + '/poms/my_caseload'
```
rossjones: Is there a *_path() we could use instead of hard-coding?

app/services/prison_offender_manager_service.rb
```diff
@@ -23,6 +23,9 @@ def self.get_poms(prison)
   def self.get_pom(caseload, nomis_staff_id)
```
rossjones: Not for now, but we should try and optimise this at some point.

app/views/allocations/confirm.html.erb
```diff
@@ -6,7 +6,7 @@
     <%= form_tag(allocations_path, method: :post, id: "confirm_allocation_form") do %>
       <h1 class="govuk-heading-xl govuk-!-margin-top-4">Confirm allocation</h1>
       <p class="govuk-body">You are allocating <%= "#{@prisoner.first_name} #{@prisoner.last_name}".titleize %> to <%= "#{@pom.first_name} #{@pom.last_name}".titleize %>.</p>
-      <p class="govuk-body">We will send a confirmation email to POM-EMAIL-ADDRESS.</p>
+      <p class="govuk-body">We will send a confirmation email to <%= @pom.emails.first %>.</p>
```
rossjones: What does it do if `@poms.emails.empty?` ?

KPobeeNorris: At the moment it displays "We will send a confirmation email to .", which isn't ideal.  However, I've created a ticket asking Leigh to provide some content, and spoken to Jay about the idea of trying to catch POMs not having email addresses earlier in the process.  I'm happy to put something in temporarily?  I can also add a `return if @poms.emails.empty` line into the PomMailer to stop it trying to send messages?

app/services/prison_offender_manager_service.rb
KPobeeNorris: Agreed!

app/views/allocations/confirm.html.erb
rossjones: Yeah sounds good.  Just add something temporary (or even just remove that line entirely) and add it to the ticket for Leigh?  Having the mailer be a bit more defensive when it won't be able to send mail could be a good thing, but also the error telling us a user doesn't have an email is useful.  I'd say for now, go with being defensive.

app/services/email_service.rb
KPobeeNorris: Tried using `my_caseload_path`, but it's not recognised, think we might have to hardcode it

rossjones: Rails.application.routes.url_helpers.my_caseload_oath

PR #300, opened 2019-03-15
Allow switching prisons

rossjones:
If a POM has acess to more than one prison, then this PR allows them to
switch between them.


rossjones:
Should probably remove the old ‘switcher’ from the header before review.

app/controllers/application_controller.rb
```diff
@@ -20,8 +20,15 @@ def active_caseload
...
+  def update_active_caseload(code)
+    session[:sso_data]['active_caseload'] = code
+  end
+
   def caseloads
-    sso_identity['caseloads'] if sso_identity.present?
+    return nil if sso_identity.blank?
+
+    caseloads = sso_identity['caseloads']
+    caseloads.reject { |c| c == 'NWEB' }
```
TobyRet: How come NWEB get's returned in caseloads?

app/views/search/search.html.erb
```diff
@@ -1,3 +1,6 @@
+<% content_for :switcher do %>
```
TobyRet: I wonder if 
```
<% content_for :switcher do %>
  <%= render '/layouts/prison_switcher' %>
<% end %>
```

can be moved into layouts/application.html.erb  - then it's only defined in one place

app/controllers/application_controller.rb
KPobeeNorris: Because NWEB is assigned under the caseload section in NOMIS, the same screen where prisons are added

PR #304, opened 2019-03-18
Display flash messages

TobyRet:
- When an allocation is saved successfully the user will see a 'Success'
flash message appear once they are re-directed to the 'Make allocations'
tab
- Conversely, if the save is unsuccessful an 'unsuccessful' flash
message will appear

flash

### DIFF
--- a/app/controllers/allocations_controller.rb
+++ b/app/controllers/allocations_controller.rb
@@ -51,7 +51,7 @@ class AllocationsController < ApplicationController
     if AllocationService.create_allocation allocation
       flash[:notice] = "#{offender.full_name_ordered} has been allocated to #{pom.full_name_ordered} (#{pom.grade})"
     else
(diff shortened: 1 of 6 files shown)


app/controllers/allocations_controller.rb
```diff
@@ -19,34 +19,49 @@ def edit
...
       override_reasons: override_reasons,
       override_detail: override_detail,
       message: allocation_params[:message]
-    )
+    }
+
+    if AllocationService.create_allocation allocation
+      flash[:notice] = "#{offender.full_name_ordered} has been allocated to #{pom.full_name_ordered} (#{pom.grade})"
+    else
+      flash[:alert] = "#{offender.full_name_ordered} has note been allocated  - please try again"
```
rossjones: Typo: note->not?

PR #313, opened 2019-03-19
Apply new rules for determining case responsibility

TobyRet:
Rules applied as illustrated.

Case_responsibility_decision_tree

### DIFF
--- a/app/services/responsibility_service.rb
+++ b/app/services/responsibility_service.rb
@@ -25,30 +25,26 @@ class ResponsibilityService
     return UNKNOWN if offender.release_date.nil?
     return SUPPORTING unless welsh?(offender)
 
-    return RESPONSIBLE if welsh?(offender) &&
-      nps_case?(offender) &&
+    return RESPONSIBLE if nps_case?(offender) &&
       new_case?(offender) &&
       release_date_gt_10_mths?(offender)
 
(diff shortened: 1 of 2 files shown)


app/services/responsibility_service.rb
```diff
@@ -1,30 +1,92 @@
...
     else
-      'Unknown'
+      UNKNOWN
     end
   end
 
   def self.calculate_pom_responsibility(offender)
-    return 'Unknown' if offender.release_date.nil?
-    return 'Responsible' if assign_responsible?(offender)
+    return UNKNOWN if offender.release_date.nil?
```
TobyRet: I've introduced guard clauses in order to remove if / else statements.
I'm not sure if iterating over rules will bring about much benefit, if anything might make the card harder to read? Welcome opinions!

TobyRet: Also welcome suggestions regarding code climate issues (too may return methods). The class method could be split but then I'm not sure if this then makes the code more readable?

rossjones: Looks already and is pretty clear.  Might be tempted to pre-calculate the values to save calling functions multiple times, and I don't think welsh? is required after the first guard?

TobyRet: Thanks for the comments - I've pushed the changes :-)

PR #314, opened 2019-03-19
Show POM responsibility during allocation

rossjones:
When we allocate an offender we need to show the POM responsibility for
that allocations.  This is to enable future changes where we might store
this responsibility with the allocation.

Also contains a fly-by change removing a few pixels from the prison
switcher

### DIFF
--- a/app/services/responsibility_service.rb
+++ b/app/services/responsibility_service.rb
@@ -24,6 +24,12 @@ class ResponsibilityService
   end
 
   def self.assign_responsible?(offender)
+    # TODO: When we do this check, we should also check what the responsibility
+    # was yesterday, so that we can determine if it has changed.  This will allow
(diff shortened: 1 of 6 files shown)


app/services/responsibility_service.rb
```diff
@@ -24,6 +24,12 @@ def self.nps_calculation(offender)
   end
 
   def self.assign_responsible?(offender)
```
TobyRet: I cannot see this being called anywhere 🤔

rossjones: It's called at https://github.com/ministryofjustice/offender-management-allocation-manager/pull/314/files#diff-0f5c3ebaf1f0ae922a6ea7b08b835923R15 in the fold above

PR #317, opened 2019-03-19
Add more flexible way of getting all offenders

rossjones:
Currently we have more than one way to get lists of offenders (summary
service and search) that are complex and messy and not necessarily doing
identical things - primarily because search doesn't need all of the
information.

This PR is a bit of an experiment on how we might simplify the fetching
of lists of offenders, at the same time that we filter them and add
local data to them.

The intent is that the local API looks as follows where the arguments
passed to add_batch_filter are lambdas that receive a batch of
offenders, optionally adds data to them, and optionally removes some
records.

```ruby
offender_list = OffenderList.new('LEI', batch_size: 200)
offender_list.add_batch_filter(add_release_date)
offender_list.add_batch_filter(remove_no_release_date)
offender_list.add_batch_filter(add_case_information)

offenders = offender_list.fetch
```

Longer term, and as long as the batch filters are re-entrant (
anything captured in the closure is read-only) if should in future be
possible to introduce threads in the fetch method to dramatically reduce
the cold-start time w/o the cache.

### DIFF
--- a/app/services/nomis/elite2/offender_list.rb
+++ b/app/services/nomis/elite2/offender_list.rb
@@ -1,0 +1,69 @@
+# Fetching a list of offenders from the Elite2 API is something that hapens
+# frequently, often then filtered and enriched with local data.  The
+# OffenderList class encapsulates this functionality in one place to provide
+# some consistency with how it is done.
+#
+# Usage:
+#   # Fetch all offenders for a prison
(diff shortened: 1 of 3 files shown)


rossjones:
Going to close this PR, but keep the branch until I have time to work on it.

PR #355, opened 2019-03-27
Only show sentenced offenders

rossjones:
A prisoner will have had a sentence calculation and for our purposes
this means that they will either have a:
1) Release date, or
2) Parole eligibility date, or
3) HDC release date (homeDetentionCurfewEligibilityDate field).

If they do not have either of these we should be checking for a tariff
date.

Once we have all the dates we then need to display whichever is the
earliest one.

If they do not have any of these dates then we can assume they have been
convicted, but are awaiting sentencing and therefore shouldn’t be
showing up in our service.


app/services/nomis/models/offender.rb
```diff
@@ -28,14 +28,36 @@ class Offender
...
+      attribute :release_date, :date
+      attribute :sentence_date, :date
       attribute :parole_eligibility_date, :date
-      attribute :has_indeterminate_release_date
+      attribute :home_detention_curfew_eligibility_date, :date
+      attribute :tariff_date, :date
+
+      def sentence_detail=(sentence_detail)
+        self.release_date = sentence_detail.release_date
+        self.sentence_date = sentence_detail.sentence_date
```
KPobeeNorris: A tiny question: can we rename this to be sentence_start_date?  Sentence date could potentially cause some confusion if the context/datapoint isn't known.

rossjones: Good idea.

PR #356, opened 2019-03-27
Create and use a non-root user in the docker image

digitalronin:
The pod security policy in the live-1 cluster blocks containers which run processes as root.

This change adds a non-root user after all the software installation steps have finished, and then switches to it before running Rails.

Please could you try this and see if it works for your service?


TobyRet:
Closing and raising again as new PR - Circle build not firing

PR #381, opened 2019-03-29
Sort prison switcher

KPobeeNorris:
The prison switcher page currently displays an unsorted list, which makes it more difficult to find the prison you want the longer the list gets.  This will now be sorted alphabetically.

This PR also removes an unused method and corrects an attribute name in the OffenderShort model.


KPobeeNorris:
Superseded by PR#382

PR #382, opened 2019-03-29
Update prison switcher

KPobeeNorris:


app/views/prisons/index.html.erb
```diff
@@ -4,9 +4,9 @@
 <h1 class="govuk-heading-xl govuk-!-margin-top-0 govuk-!-margin-bottom-4">Choose prison</h1>
 
 <ul class="govuk-list">
-  <% caseloads.each do |prison| %>
+  <% PrisonService.prisons_from_list(caseloads).each do | prison| %>
```
rossjones: It should be possible to unpack the array response in the block I think rather than [0] and [1] which aren't very clear.

```
PrisonService.prisons_from_list(['LEI']).each { |code, title| 

 }
```

PR #385, opened 2019-04-01
Whitelist pingdom EU IPs

rossjones:
Only allow pingdom to connect from their EU ips.


amitmuthu1:
All looks good. Couple of non-blocking things:
1. Would be good to have something in the code (comment?) to point maintainers at where the list came from
2. Any reason the first IP has an explicit netmask and none of the others do? Keeping the netmask kept the diff clean but if it's not needed would be good to tidy

rossjones:
The IP addresses are from the EU section of https://help.pingdom.com/hc/en-us/articles/203682601-Pingdom-probe-servers-IP-addresses 

The list is supposed to be a comma-separated list of CIDRs, so the hope was I suspect that some might be non /32. We can definitely clean it up. I've added a ticket to the backlog at https://trello.com/c/HsXnMwGd/644-cleanup-the-production-whitelist-addresses

PR #389, opened 2019-04-03
 Update POM caseload page table header

KPobeeNorris:
The table header in the POMs caseload pages still references 'release/
parole eligibility date', but this has been updated elsewhere to display
'earliest release date', this PR just updates the pages to be in line
with the rest of the site.

### DIFF
--- a/app/services/nomis/models/sentence_detail.rb
+++ b/app/services/nomis/models/sentence_detail.rb
@@ -14,6 +14,18 @@ module Nomis
       attribute :facial_image_id
       attribute :internal_location_desc
 
+      def earliest_release_date
+        dates = [
+            release_date,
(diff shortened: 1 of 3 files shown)


app/services/nomis/models/sentence_detail.rb
```diff
@@ -14,6 +14,18 @@ class SentenceDetail
       attribute :facial_image_id
       attribute :internal_location_desc
 
+      def earliest_release_date
```
rossjones: This may be a better place for this than Offender or OffenderShort, we should remember to see if we can either move it out of there, or store it somehow.

KPobeeNorris: I'll add a ticket so we can come back to this

PR #395, opened 2019-04-05
Merge the two offender models

KPobeeNorris:
We currently have two offender memory models, offender and
offenderShort.  This is due to the fact that there isn't one endpoint
within the elite2 api that we can gather all the relevant information
for an offender from.  However, it's not ideal having two models related
to offenders so this PR merges them into a single model.


KPobeeNorris:
Whilst reviewing this PR a discussion was held within the team around the pros and cons of taking the approach to merge the two offender models; and resulted in other suggestions being made.  

One suggestion was to pass a transform_map to the deserialiser, so if present it would rename fields to the correct name.  So say we only use Offender, we would translate booking_id in the hash to latest_booking_id.

Another suggestion was to use another API that we have access to, but unfortunately there wasn't a single endpoint that we could get all the data from that we need.

In the end an agreement was reached that we would keep the models separate (to  closely mirror the API calls), rename the models, and remove any attributes we don't use.

PR #400, opened 2019-04-05
Refactor offender models

KPobeeNorris:
We currently have two offender memory models as we have to call two
endpoints from the API to fetch all the data we need for an offender.
There are also a number of attributes that we don't use and are storing
needlessly.  

This PR renames the offender model to offenderSummary and tidies up the offender models by removing attributes we don't use, as well as sorting them so it's easier to see/find attributes.


app/services/nomis/models/offender_short.rb
```diff
@@ -5,38 +5,28 @@ module Models
     class OffenderShort
```
rossjones: Should this be Offender?

app/services/api_deserialiser.rb
```diff
@@ -21,15 +21,9 @@ def deserialise_hash(memory_model_class, payload)
       setter = "#{key}="
       if memory_model.respond_to?(setter)
         memory_model.public_send(setter, value)
-      else
-        raise_warning_for(memory_model_class, key)
```
TobyRet: Is this supposed to be deleted @KPobeeNorris ?

PR #412, opened 2019-04-09
Implements the front end for sorting

rossjones:
Makes the appropriate headings (in summary only for now) clickable so
they can sort the page of results.

### DIFF
--- a/app/helpers/sort_helper.rb
+++ b/app/helpers/sort_helper.rb
@@ -29,9 +29,9 @@ module SortHelper
     return '' unless current_sort.present? && current_sort.start_with?(field_name)
 
     if current_sort.end_with?('asc')
-      content_tag(:span, '&#9650'.html_safe, class: "sort-arrow" )
+      content_tag(:span, '&#9650'.html_safe, class: 'sort-arrow')
     else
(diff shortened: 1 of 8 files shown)


app/helpers/sort_helper.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  def display_arrow(field_name)
+    uri = URI.parse(request.original_url)
+
+    query = Rack::Utils.parse_query(uri.query)
+    current_sort = query['sort']
+    return '' unless current_sort.present? && current_sort.start_with?(field_name)
+
+    if current_sort.end_with?('asc')
+      '<span class="sort-arrow up">&#9650;</span>'.html_safe
```
TobyRet: `content_tag(:span, class: "sort-arrow up", "&#9650")`

TobyRet: Using this the `content_tag` helper takes care of html_safe for you

TobyRet: `content_tag(:span, '&#9650'.html_safe, class: "sort-arrow up" )`

rossjones: Thanks

PR #431, opened 2019-04-18
Adds a new allocation detail page

rossjones:
This page is shown when the user clicks on the View link from the
allocations tab of the summary (currently commented out).  It shows
details of the allocation, the prisoner info, and the allocation
history.  It allows for reallocation (goes to current allocation page).

Some of the data in the designs is not yet available, and so the HTML is
commented out so we don't display the columns.


rossjones:
This should be ready now @TobyRet

PR #442, opened 2019-04-24
Display Keyworker on the Allocation Info page for an Offender

TobyRet:
- If an offender does not have a Keyworker assigned then 'Not assigned' will be displayed


app/controllers/allocations_controller.rb
```diff
@@ -20,10 +20,9 @@ def new
   # rubocop:disable Metrics/LineLength
   def show
     @prisoner = offender(nomis_offender_id_from_url)
-
     allocation = AllocationService.active_allocations_with_pom_detail(@prisoner.offender_no).first
     @pom = PrisonOffenderManagerService.get_pom(active_caseload, allocation.nomis_staff_id)
-
+    @keyworker = Nomis::Keyworker::KeyworkerApi.get_keyworker(active_caseload, @prisoner.offender_no)
```
TobyRet: I did not think there was any point putting in a Keyworker service at this stage, but happy to add if preferred

spec/features/allocation_information_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+    end
+
+    it "displays the Key Worker's details" do
+      signin_user
+
+      visit show_allocations_path(nomis_offender_id_with_keyworker)
+
+      expect(page).to have_css('h1', text: 'Allocation information')
+
+      # Prisoner
```
TobyRet: I'm not sure whether I should be adding more assertions here (there are potentially many things to assert on this page)

app/controllers/allocations_controller.rb
rossjones: Can't imagine we'll need anything else from it at this point. If we do we can consider it then.

app/services/nomis/models/keyworker_details.rb
```diff
@@ -8,6 +8,12 @@ class KeyworkerDetails
       attribute :staff_id, :integer
       attribute :first_name, :string
       attribute :last_name, :string
+
+      def full_name
+        return "#{last_name}, #{first_name}".titleize if first_name || last_name
```
rossjones: Should this be `&&` instead of `||`?

PR #443, opened 2019-04-24
Fix the timeline component

rossjones:
Ensures the O is roughly in the middle of the adjacent LI, and that the
vertical border does not go past the last O next to the last LI.

Was:
```
   O
   |
   O
   |
   O
   |
```

Now
```
   O
   |
   O
   |
   O
```

We do this by ensuring the top starts at half of the LI height from the
top (about 94px currently) and the bottom margin is 94px too so that we
stop at the last O.  Ideally if we could fix the height of the LIs then
we can remove the magic number here by calculating those values in sass.


TobyRet:
If this stops working (div sizes change) then maybe we can use the CSS `calc()` function?

PR #451, opened 2019-04-29
Prepare for Co-working POMS

TobyRet:
This is the first of a series of PR's to introduce the Co-working POM feature.

In this PR, preparations are made to handle the Co-working POM in our database. The appropriate schema changes have been made and all tests / references have been updated.

Rather than store a `nomis_staff_id` against each allocation, we will now store the `primary_pom_nomis_id` and the `secondary_pom_nomis_id` (optional)

The association between an Allocation and PomDetail is now no longer makes sense has been removed

### DIFF
--- a/app/models/allocation.rb
+++ b/app/models/allocation.rb
@@ -1,12 +1,26 @@
 # frozen_string_literal: true
 
 class Allocation < ApplicationRecord
-  belongs_to :pom_detail
-
   attr_accessor :responsibility
 
+  scope :active_allocations, lambda { |nomis_offender_ids|
+    where(nomis_offender_id: nomis_offender_ids)
+  }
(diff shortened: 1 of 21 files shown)


app/models/allocation.rb
```diff
@@ -1,12 +1,26 @@
 # frozen_string_literal: true
 
 class Allocation < ApplicationRecord
-  belongs_to :pom_detail
-
   attr_accessor :responsibility
 
+  scope :active_allocations, lambda { |nomis_offender_ids|
```
rossjones: Should the active_allocations scope check active field? Are this and the following scope back to front?

TobyRet: Yep you are right - fixed

PR #453, opened 2019-04-29
Refactoring - move database calls to Rails scopes

TobyRet:
Moving database calls to scopes will dry up the code and decrease the size of services. :-)

### DIFF
--- a/app/controllers/allocations_controller.rb
+++ b/app/controllers/allocations_controller.rb
@@ -20,7 +20,7 @@ class AllocationsController < ApplicationController
   # rubocop:disable Metrics/LineLength
   def show
     @prisoner = offender(nomis_offender_id_from_url)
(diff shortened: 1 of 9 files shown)


app/controllers/allocations_controller.rb
```diff
@@ -20,7 +20,7 @@ def new
   # rubocop:disable Metrics/LineLength
   def show
     @prisoner = offender(nomis_offender_id_from_url)
-    primary_pom_nomis_id = AllocationService.primary_pom_nomis_id(@prisoner.offender_no)
+    primary_pom_nomis_id = Allocation.primary_pom_nomis_id(@prisoner.offender_no)
```
rossjones: I wonder if this might be better as 
```
primary, secondary = AllocationService.get_poms(allocation_id) 
```

TobyRet: Yep that makes sense :-)  - perhaps hold fire on the change until Secondary POMs are introduced (next PR)?

PR #454, opened 2019-04-29
Display link to prisoner's New NOMIS profile

KPobeeNorris:
The new designs for post-MVP include a link on the new allocation &
reallocation pages which allows the user to view a prisoner's profile on
New Nomis.  This is a more detailed profile and makes sense that we
direct the user there, rather than duplicating the information.  Any
user that can sign into our MOIC service is able to access New Nomis so
there are no auth related issues.

The link will open a new browser tab so the user can easily switch back
to the MOIC tab and continue from where the left off.


app/views/shared/_basic_prisoner_info.html.erb
```diff
@@ -14,8 +14,11 @@
       <td class="govuk-table__cell table_cell__left_align govuk-!-width-one-half"><%= format_date(@prisoner.date_of_birth) %></td>
     </tr>
     <tr class="govuk-table__row">
-      <td class="govuk-table__cell hmpps-table-cell__key">Prisoner number</td>
-      <td class="govuk-table__cell table_cell__left_align hmpps-table-cell__key"><%= @prisoner.offender_no %></td>
+      <td class="govuk-table__cell">Prisoner number</td>
+      <td class="govuk-table__cell table_cell__left_align">
+        <%= @prisoner.offender_no %>
+        <a class="govuk-link pull-right" href="<%= Rails.configuration.new_nomis_host %>/offenders/<%= @prisoner.offender_no%>/quick-look" target="_blank">View NOMIS profile</a>
```
rossjones: Not sure if it is possible, but it would be pretty awesome if we could have a named route to an external resource... so for instance `new_nomis_profile_url(offender_no)` but I suspect it might mean a controller in an action somewhere.  @TobyRet do you know if this is possible?

spec/features/allocation_information_spec.rb
```diff
@@ -61,6 +61,23 @@
     end
   end
 
+  describe 'Prisoner profile links', vcr: { cassette_name: :show_allocation_information_new_nomis_profile } do
+    before do
+      create_case_information_for(nomis_offender_id_with_keyworker)
+      create_allocation(nomis_offender_id_with_keyworker)
+    end
+
+    it "displays the Key Worker's details" do
```
rossjones: Is this label right?

README.md
```diff
@@ -88,6 +88,7 @@ Several environment variables are required for the operation of this service.
 | NOMIS_OAUTH_CLIENT_ID | The client ID of the application in OAUTH |
 | NOMIS_OAUTH_AUTHORISATION | Oauth authorisation string (base64 encoded) |
 | NOMIS_OAUTH_PUBLIC_KEY  | This is the base64 encoded public key for decoding Tokens provided by the OAUTH server |
+| NEW_NOMIS_HOST | The host where New NOMIS is hosted |
```
rossjones: Now called 'Digital Prison Services', rather than New Nomis :(

PR #470, opened 2019-05-07
Set up Paper Trail and record changes to Allocations

TobyRet:
Apologies this is a _big_ PR. A necessary evil?

Rather than treat allocations and reallocations as events, we are now updating a single 'Allocation record' (in this PR it's actually a AllocationVersion record - see further for detail) and using Paper Trail to store changes as versions. 

Work included:

1. Setting up and configure Paper trail gem for tracking changes to Allocations

2. Copying over Allocations to a new table - :allocation_version. Note that the schema adds columns for `event` and `event_trigger` where events to be stored include:
- `:allocate_primary_pom`
- `:reallocate_primary_pom`
- `:allocate_secondary_pom`
- `:reallocate_secondary_pom`
- `:deallocate_primary_pom`
- `:deallocate_secondary_pom` 

... and event_triggers include:
- `:user`
- `:offender_movement` 

3. To play around with the versions, reset your db and run the rake task `bundle exec rake allocations:copy`. Get back a version by opening up the rails console and running `AllocationVersion.version.last.reify`

4. Application is rewired to use AllocationVersion

5. A new route / view is now included for 'confirming reallocation' - this is so that we can send through the right events to the model. Happy to discuss this further as there may be a better solution


app/models/allocation_version.rb
```diff
@@ -0,0 +1,49 @@
+# frozen_string_literal: true
+
+class AllocationVersion < ApplicationRecord
+  has_paper_trail
+
+  attr_accessor :responsibility
+
+  enum event: [
+    :allocate_primary_pom,
+    :reallocate_primary_pom
```
TobyRet: Other events need top be added here

app/controllers/allocations_controller.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'byebug'
```
rossjones: Bzzt.

app/services/allocation_service.rb
```diff
@@ -46,16 +52,30 @@ def self.active_allocations(nomis_offender_ids)
   end
 
   def self.previously_allocated_poms(nomis_offender_id)
-    Allocation.inactive_allocations(nomis_offender_id).
-      map(&:primary_pom_nomis_id)
+    allocation = AllocationVersion.find_by(nomis_offender_id: nomis_offender_id)
+
+    if allocation.nil?
```
rossjones: If this simpler as 

```
return [] if allocation.nil?
```

as it'll remove the need for the if/else

app/services/prison_offender_manager_service.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'byebug'
```
rossjones: probably not necessary

rossjones:
A couple of comments added inline.  It'd be cool to find out what `versioning: true` does in the test signatures.

app/services/allocation_service.rb
TobyRet: 👍

app/controllers/allocations_controller.rb
```diff
@@ -46,49 +47,44 @@ def edit
...
-    offender = offender(allocation_params[:nomis_offender_id])
-    pom = PrisonOffenderManagerService.get_pom(
+  def confirm_reallocation
+    @prisoner = offender(nomis_offender_id_from_url)
+    @pom = PrisonOffenderManagerService.get_pom(
       active_caseload,
-      allocation_params[:nomis_staff_id]
+      nomis_staff_id_from_url
     )
+    @event = :reallocate_primary_pom
```
TobyRet: There may be a better way to do this ...
We need to make sure that we know when we are dealing with an allocation or reallocation. Both require a 'confirm' step so now this has been split into confirm(allocation) and confirm_reallocation. The naming could be changed ...

app/models/allocation_version.rb
```diff
@@ -0,0 +1,75 @@
...
+  ALLOCATE_SECONDARY_POM = 2
+  REALLOCATE_SECONDARY_POM = 3
+  DEALLOCATE_PRIMARY_POM = 4
+  DEALLOCATE_SECONDARY_POM = 5
+
+  USER = 0
+  OFFENDER_MOVEMENT = 1
+
+  enum event: {
+    allocate_primary_pom: ALLOCATE_PRIMARY_POM,
```
TobyRet: Enum values are manually assigned in order to prevent problems that arise from using defaults and accidentally changing the order

app/views/allocations/edit.html.erb
```diff
@@ -12,4 +12,4 @@
 
 <hr class="govuk-section-break govuk-section-break--l govuk-section-break--invisible"/>
 
-<%= render 'allocations/pom_tables' %>
+<%= render 'allocations/pom_tables_reallocate' %>
```
TobyRet: When we are 'editing an allocation' we know that we are actually dealing with a reallocation

db/schema.rb
```diff
@@ -11,10 +11,33 @@
 # It's strongly recommended that you check this file into your version control system.
 
 ActiveRecord::Schema.define(version: 2019_05_07_144554) do
```
TobyRet: mmmm ... should the timestamp be different here?

app/controllers/allocations_controller.rb
```diff
@@ -18,19 +22,16 @@ def new
...
     @pom = PrisonOffenderManagerService.get_pom(active_caseload, primary_pom_nomis_id)
     @keyworker = Nomis::Keyworker::KeyworkerApi.get_keyworker(active_caseload, @prisoner.offender_no)
     @history = AllocationService.offender_allocation_history(@prisoner.offender_no)
   end
-  # rubocop:enable Metrics/LineLength
 
-  # rubocop:disable Metrics/MethodLength
   def edit
-    unless AllocationService.active_allocation?(nomis_offender_id_from_url)
+    unless AllocationVersion.find_by(nomis_offender_id: nomis_offender_id_from_url)
```
rossjones: Does this need constraining to active allocations?

app/models/allocation_version.rb
rossjones: What do you think about adding a comment explaining how to add a new event and why it is done that way - to help others add new ones.

app/models/allocation_version.rb
```diff
@@ -0,0 +1,75 @@
...
+
+  enum event_trigger: {
+    user: USER,
+    offender_movement: OFFENDER_MOVEMENT
+  }
+
+  scope :allocations, lambda { |nomis_offender_ids|
+    where(nomis_offender_id: nomis_offender_ids)
+  }
+  scope :all_primary_pom_allocations, lambda { |nomis_staff_id|
```
rossjones: This seems to be all allocations for a primary POM, where the event was an allocate or reallocate.
If the last event on an allocation was neither of these (for instance assigning a coworker) does that mean it won't show up as an allocation for the POM?

app/models/allocation_version.rb
```diff
@@ -0,0 +1,75 @@
...
+  }
+  scope :primary_pom_nomis_id, lambda { |nomis_offender_id|
+    allocations(nomis_offender_id).primary_pom_nomis_id
+  }
+
+  def self.deallocate_offender(nomis_offender_id)
+    allocations(nomis_offender_id).
+      update_all(
+        event: DEALLOCATE_PRIMARY_POM,
+        event_trigger: USER
```
rossjones: Is this always a USER trigger? Seems like the type of thing the model might not know about and have to be told.

db/schema.rb
rossjones: Is papertrail sneakily writing to the file instead of re-generating the schema?  Might be worth a manual regenerate if so to make sure it's different

app/models/allocation_version.rb
TobyRet: 👍

TobyRet: Done

db/schema.rb
TobyRet: Schema regenerated

app/controllers/allocations_controller.rb
TobyRet: It's possible that an Allocation may have no primary pom assigned so I assume we just need to check for the existence of a record else they need to be redirected to create on

app/models/allocation_version.rb
TobyRet: Yes currently that is true and would need to be modified for co-worker once this ticket is worked on. I'll add a todo

app/controllers/allocations_controller.rb
rossjones: Yeah, I think no allocated_poms (primary or secondary) is probably a fair proxy for inactive-allocation

app/models/allocation_version.rb
TobyRet: Good spot - Movements can also deallocate an offender ...
I'll fix

app/controllers/allocations_controller.rb
TobyRet: Done! We now check whether an allocation is active

PR #479, opened 2019-05-10
[POM-69] Prisoner profile

rossjones:
This PR implements the prisoner profile that is linked to from the caseload page.  Currently this is not linked as the prisoner_profile feature is disabled pending other work.


app/services/prison_offender_manager_service.rb
```diff
@@ -84,6 +84,12 @@ def self.get_allocated_offenders(nomis_staff_id, prison, offset: nil, limit: nil
         offender_stub.omicable = record.omicable
       end
 
+      # TODO: Handle this properly, what happens when the sentence_start_date disappears
```
rossjones: Just removing this....

PR #493, opened 2019-05-14
Copies allocations from old table to new table

rossjones:
This PR will need reversing once it has successfully run on prod.


rossjones:
We can't do this, we'll end up with 3 versions of each allocation as each of the 3 pods try and run the task.

PR #513, opened 2019-05-20
 Log directly to elastic_search

rossjones:
Using the semantic logging gem, this PR will start logging directly to
the elastic search whose URL is in the ELASTIC_URL environment variable.
The index created will be offender-management-allocation-ENV where ENV
is either staging or production.  This will allow us to create a mapping
in Kibana for the index.


rossjones:
Closing this (but leaving branch) until cloud platform has support for per-app indices for ES.

PR #514, opened 2019-05-20
POM-89: Migrate 'created_at' dates

TobyRet:
- Created Rake task that migrates Allocation 'created_at' dates to the 'primary_pom_allocated_at' date in the corresponding AllocationVersion record


spec/factories/allocation.rb
```diff
@@ -0,0 +1,37 @@
+require 'faker'
```
rossjones: Is this meant to be in this PR?

spec/services/date_migrater_spec.rb
```diff
@@ -0,0 +1,54 @@
+require 'rails_helper'
+require 'byebug'
```
rossjones: byebug?

spec/factories/allocation.rb
TobyRet: The Faker gem is for creating dummy data. Once the migration is run this factory will be deleted

TobyRet: Yep (it's for creating dummy data, however this factory will be removed once the migration is run)

PR #523, opened 2019-05-22
[POM-103] Change how we send any reallocation emails

rossjones:
Currently the way we determine whether to send a reallocation email,
along with the allocation email, results in us always sending a
reallocation email regardless of that state.  This is because it
just pulls the last allocation and assume it is the previous one which 
worked before we changed to using papertrail.

This PR changes how it is sent, and contains slight changes to the
interface.  It is expected that only a message and the recently created
allocation object will be passed to the EmailService.instance() class
method, the other data is derived from this.

To determine if the service should send a reallocated email, we now
check that the newly created allocation is a `:reallocated_primary_pom`
event, and if so try and determine the previous pom's details.  We do
this by iterating through the previous versions to find the
first primary_pom that is not the pom in the current allocation. 

We _should_ check that there is no a 'version' without a POM in the 
history but this currently relies on the `:reallocated_primary_pom` event being
accurate (as it is over-writing a previously allocated POM) when
triggered by the user interface.  This _should_ be true. We should
ensure that there are no further edge cases. Oh to have a ruby that supports 
'design by contract' :( 

Further refactoing of the EmailService, when we look again at all of the
notifications we might send, may want to break this service up to handle
specific events explicitly rather than handling logic within the service
itself. cf modularised services.

### DIFF
--- a/app/services/email_service.rb
+++ b/app/services/email_service.rb
@@ -19,7 +19,7 @@ class EmailService
     )
   end
 
-  def send_allocation_email
+  def send_email
     return if @pom.emails.empty?
 
     if @allocation.event == 'reallocate_primary_pom' && previous_pom.present?
--- a/app/services/email_service.rb
+++ b/app/services/email_service.rb
@@ -2,29 +2,30 @@
 
 class EmailService
   def self.instance(params)
-    offender = OffenderService.get_offender(params[:nomis_offender_id])
-    pom = PrisonOffenderManagerService.get_pom(
-      params[:prison],
-      params[:primary_pom_nomis_id]
-    )
-    last_allocation = AllocationService.last_allocation(params[:nomis_offender_id])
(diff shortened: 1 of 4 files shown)


app/services/email_service.rb
```diff
@@ -2,29 +2,30 @@
...
+    @allocation = allocation
+
+    @offender = OffenderService.get_offender(@allocation[:nomis_offender_id])
+    @pom = PrisonOffenderManagerService.get_pom(
+      @allocation.prison,
+      @allocation.primary_pom_nomis_id
+    )
   end
 
   def send_allocation_email
```
TobyRet: Maybe `deliver_email` might be more generic than `send_allocation_email`?

rossjones: Fixed now @TobyRet

PR #548, opened 2019-06-10
[POM-63] Move allocation history

theKHutDeveloper:
This PR creates a new route `/allocations/:nomis_offender_id/history` that will display the allocation history for an offender. The allocation history will show the prisoner allocations events grouped by prison.

### DIFF
--- a/app/controllers/allocations_controller.rb
+++ b/app/controllers/allocations_controller.rb
@@ -86,8 +86,7 @@ class AllocationsController < ApplicationController
   end
 
   def history
-    offender = offender(nomis_offender_id_from_url)
-    @prisoner_name = offender_name_by_last_name(offender)
+    @prisoner = offender(nomis_offender_id_from_url)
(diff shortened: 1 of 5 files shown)


app/controllers/allocations_controller.rb
```diff
@@ -183,6 +189,10 @@ def override_detail
   def suitability_detail
     @override[:suitability_detail] if @override.present?
   end
+
+  def offender_name_by_last_name(offender)
```
rossjones: The offender model has a method to do this for you

app/controllers/allocations_controller.rb
```diff
@@ -85,6 +85,12 @@ def create
     redirect_to summary_unallocated_path
   end
 
+  def history
+    offender = offender(nomis_offender_id_from_url)
```
rossjones: You might find this easier to pass @offender to the template, as this will make it easier to call `@offender.full_name` in the title and `@offender.nomis_no` in the backlink.

PR #553, opened 2019-06-11
Change allocation history date

theKHutDeveloper:
This PR will use the field updated date instead of primary_pom_allocated_at in the allocations history page as the primary_pom_allocated_at value can be nil, which would cause failures.

In addition we will not titlecase the Faker::Name generated in tests as it causes test failures.

### DIFF
--- a/spec/features/allocate_feature_spec.rb
+++ b/spec/features/allocate_feature_spec.rb
@@ -188,7 +188,7 @@ feature 'Allocation' do
     formatted_date = allocation.primary_pom_allocated_at.strftime("#{allocation.updated_at.day.ordinalize} %B %Y")
 
     expect(page).to have_css('h1', text: "Abbella, Ozullirn")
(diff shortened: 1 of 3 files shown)


spec/features/allocate_feature_spec.rb
```diff
@@ -185,10 +185,10 @@
     visit allocation_history_path(nomis_offender_id)
 
     allocation = AllocationVersion.find_by(nomis_offender_id: nomis_offender_id)
-    formatted_date = allocation.primary_pom_allocated_at.strftime("#{allocation.primary_pom_allocated_at.day.ordinalize} %B %Y")
+    formatted_date = allocation.primary_pom_allocated_at.strftime("#{allocation.updated_at.day.ordinalize} %B %Y")
```
rossjones: This looks like it still uses primary_pom_allocated_at - is that intentional?

PR #568, opened 2019-06-20
Do not reset prison when deallocating offenders

rossjones:
When we deallocate an offender, we should not set the prison to nil.  We
need to know which prison they were at when they were deallocated,
otherwise the history does not work.


rossjones:
This is being done as part of the refactor or deallocating offenders.

PR #570, opened 2019-06-20
Fix offender deallocation

rossjones:
When an offender moves to a new prison, we want to deallocate them from
their current POM.  This was using update_all to update some fields of
the current allocation for that offender, unfortunately this does not
work with papertrail.  The end result of this not working was that the
current allocation was overwritten, but no new version record appeared.
This breaks the history.

Rather than use update_all, this PR switches to setting the specific
fields in the allocation_version and then re-saving it.


app/models/allocation_version.rb
```diff
@@ -64,17 +64,21 @@ def self.active?(nomis_offender_id)
...
+
+    return if alloc.nil?
+
+    alloc.primary_pom_nomis_id = nil
+    alloc.primary_pom_name = nil
+    alloc.primary_pom_allocated_at = nil
+    alloc.secondary_pom_nomis_id = nil
+    alloc.secondary_pom_name = nil
+    alloc.event = DEALLOCATE_PRIMARY_POM
+    alloc.event_trigger = OFFENDER_MOVEMENT
```
rossjones: We will need to extend this because I think we will also want OFFENDER_RELEASE.

PR #575, opened 2019-06-24
Refine offender movement type

KPobeeNorris:
At present when we run the 'deallocate_offender' method within
AllocationVersion we only state that an offender has moved.  However,
this could mean that they were either transferred, or released.  By
tweaking the event trigger to be more specific this will assist us when
displaying the allocation history.

### DIFF
--- a/app/models/allocation_version.rb
+++ b/app/models/allocation_version.rb
@@ -13,7 +13,8 @@ class AllocationVersion < ApplicationRecord
   DEALLOCATE_SECONDARY_POM = 5
 
   USER = 0
-  OFFENDER_MOVEMENT = 1
+  OFFENDER_TRANSFERRED = 1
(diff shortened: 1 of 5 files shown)


app/models/allocation_version.rb
```diff
@@ -93,6 +95,10 @@ def self.deallocate_primary_pom(nomis_staff_id)
       )
   end
 
+  def self.offender_movement_type(movement_type)
```
rossjones: I don't think we'll need this at all if instead of passing in ADM or REL (which is info this class shouldn't need to know) the movement_service instead passed in AllocationVersion::OFFENDER_TRANSFERRED or AllocationVersion.OFFENDER_RELEASED.

KPobeeNorris: Of course! Refactor on it's way.

PR #577, opened 2019-06-24
Fix names with apostrophes cause test failures

theKHutDeveloper:
Sometimes the tests fail due to the expectation of a name, where
that name has an apostrophe, for example "Bob O'Shea" will
appear as "Bob O'shea". This is not correct the name should be
"Bob O'Shea".

Using the methods titleize or titlecase causes the letter after
the apostrophe to be lowercase, which will cause the expectation
failure.

For consistency we no longer use the titlecase method as it
pretty much does the same thing as titleize.

To solve this issue the titleize method has been overwritten so
that it does not lower case the letter displayed after the
apostrophe.

The bonus of using this overridden method is that previously a
double barrelled name would remove the hyphen which is not ideal.
For example "Bob Cullen-Smith" would have been converted to
"Bob Cullen Smith". Now the hyphen remains in place which is
more accurate.

### DIFF
--- a/config/initializers/inflections.rb
+++ b/config/initializers/inflections.rb
@@ -14,3 +14,13 @@
 # ActiveSupport::Inflector.inflections(:en) do |inflect|
 #   inflect.acronym 'RESTful'
 # end
+
+# encoding: utf-8
+
+module ActiveSupport
+  module Inflector
+    def titleize(word, _keep_id_suffix = false)
+      word.split(/\b/).map(&:capitalize).join
+    end
+  end
+end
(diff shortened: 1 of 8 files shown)


config/initializers/inflections.rb
```diff
@@ -14,3 +14,13 @@
 # ActiveSupport::Inflector.inflections(:en) do |inflect|
```
rossjones: Could we remove this bit, even if it was already there ...

rossjones:
Just the one comment on the code.  It might be kinda nice to have a small test somewhere to just test `titleize` in isolation with `'` and `-`. That way if someone removes it, the tests will break in a more obvious way.

PR #593, opened 2019-06-27
Create contact form

theKHutDeveloper:
This PR creates a simple contact form which consists of a text-area field and a submit button. There is validation on the form to make sure that an empty form can not be submitted.

Screen Shot 2019-06-27 at 15 33 37


app/views/pages/contact.html.erb
```diff
@@ -1,6 +1,28 @@
...
 <div class="govuk-grid-row dashboard-row">
+
   <div class="govuk-grid-column-two-thirds">
+    <%= form_tag(contact_path, method: :post, id: "contact_form") do %>
+      <div class="govuk-form-group">
+        <label class="govuk-label" for="more-detail">
+          You can use this form to ask a question, report a problem or suggest an improvement.
+        </label>
+        <span id="more-detail-hint" class="govuk-hint">
+          Do not include personal or financial information, like your National Insurance number or credit card details.
```
KPobeeNorris: @theKHutDeveloper - would be worth asking Leigh to provide some content for this as this hint is pretty generic/not relevant to our users.

app/views/pages/contact.html.erb
```diff
@@ -1,6 +1,28 @@
...
+            <span class="govuk-error-message">
+              The text box
+              <%= @contact.errors[:more_detail].first %>
+            </span>
+          <% end %>
+        <% end %>
+        <textarea class="govuk-textarea" id="more-detail" name="more_detail" rows="10" aria-describedby="more-detail-hint"></textarea>
+      </div>
+
+      <button class="gem-c-button govuk-button" type="submit">Submit</button>
```
KPobeeNorris: @theKHutDeveloper - haven't come across this class for govuk-button before, can I ask what it does?

app/views/pages/contact.html.erb
theKHutDeveloper: Changed hint text following your advice

Screen Shot 2019-06-28 at 10 37 17

app/views/pages/contact.html.erb
theKHutDeveloper: Changed to the standard govuk button class

app/controllers/pages_controller.rb
```diff
@@ -1,7 +1,30 @@
 # frozen_string_literal: true
 
 class PagesController < ApplicationController
+  include ActiveModel::Validations
```
rossjones: Don’t think we need to do the validation in the controller so we? We can rely on the model to do the validation and generate the errors.

app/controllers/pages_controller.rb
```diff
@@ -1,7 +1,30 @@
...
   def help; end
 
-  def contact; end
+  def contact
+    @contact = nil
+  end
+
+  def create_contact
+    @contact = Contact.new(contact_params[:more_detail])
+    @contact.valid?
```
rossjones: Is this leftover? Is also used below

app/controllers/pages_controller.rb
theKHutDeveloper: Removed from controller

app/controllers/pages_controller.rb
theKHutDeveloper: Removed

PR #596, opened 2019-07-01
Add Stephen as a POM

KPobeeNorris:
We have a new developer joining the team, Stephen, and we need to add
him as a POM on T3.  This changes the numbers of POMs we have and
therefore this PR updates the specs to include the new POM.


amitmuthu1:
LGTM

PR #599, opened 2019-07-02
Add an email field to the POMDetail model

rossjones:
Now that we need to start storing the POM's email, we need to add a
field to the pom_details table as somewhere to store it.


rossjones:
No longer necessary

PR #605, opened 2019-07-03
Remove prison from the case information code

rossjones:
Currently we use the prison code only to try and optimise the grouping
of offenders in the summary service, relying on it to fetch them all
once qnd then match them (or not) to the incoming offender list.

This is problematic as we are not the canonical source of an offender's
prison, and the optimisation was not necessarily faster than checking
the table for each batch of offenders being processed.

After this PR, the prison field still exists in the database (for now)
but is not used anywhere in the code.


app/services/summary_service.rb
```diff
@@ -25,11 +25,11 @@ def summary(summary_type, prison, page, params)
     # each type of record.
     @counts = { allocated: 0, unallocated: 0, pending: 0, all: 0 }
 
-    tier_map = CaseInformationService.get_case_information(prison)
+    # tier_map = CaseInformationService.get_case_information(prison)
```
KPobeeNorris: @rossjones - missed one :)

rossjones: @KPobeeNorris should be fixed now.

starswan:
Is there a reason why we haven't dropped the column in this PR? I have another PR which might 'accidentally' re-populate it which would/will be caught by a test if we had dropped the column...

rossjones:
@starswan This should be up to date now, and contains a migration to remove the field.

PR #607, opened 2019-07-04
Fix VCR cassette generation

rossjones:
This PR attempts to fix the VCR cassettes so that they are more consistent, however it does not wholly achieve that.

The PR takes steps to ensure that:

* Use of cassettes is consistent and never at the feature level
* Auth data is not held in cassettes
* They can be replayed consistently
* There are no duplicate uses of cassettes

Now, the intended use is that the first time you use the tests, and any time that you make changes to  tests that make new network calls, you should use

```
VCR=1 rspec
```

on following runs when you are just changing code, you should use

```
rspec
```

However, if you follow this approach, once you have run with `VCR=1` there is an issue with timeouts, after a period of time tests will start failing as a result of attempting to start the oauth flow, and you will be required to run the tests with `VCR=1` again.  

There are also bearer tokens stored in the SSO tests, it may be that we want to either remove those tests, or mock them (which seems pointless as they are mocked elsewhere for tests). Changing the response from SSO is feasible in a VCR#before_record but then the tests will fail on playback.


spec/spec_helper.rb
```diff
@@ -25,14 +25,29 @@
...
+  config.default_cassette_options = { record: vcr_record_mode }
+  config.default_cassette_options = { match_requests_on: [
+    :method,
+    :query,
+    :path,
+    :body,
+    :paging_headers
+  ] }
+
+  config.register_request_matcher :paging_headers do |r1, r2|
```
starswan: Not sure about this style - shouldn't the paging_headers constant be PAGING_HEADERS? 
and unsure whether deleting items from the headers hash with keep_if is a good idea - isn't this just the same as r1.headers.select { |k, _| paging_headers.include?(k) } == r2.headers.select { |k, _| paging_headers.include?(k) } ?

starswan:
I think I've worked out the confusion. We're trying to mock the SSO connection in our tests, but the desire to go out to real services (when VCR=1 on this branch) contradicts this. I guess the 'real' SSO must win mostly to make the tests pass when run against T3 NOMIS

PR #608, opened 2019-07-04
refactor URLs to include selected prison

starswan:


rossjones:
LGTM. The only issue I found was that when I click the title in the header "Allocate to a prison offender manager" the browser redirects to /.  If I've switched from say LEI to WVI then when I go to / I'm defaulted to my main prison (as we'd sort of expect).

Might be worth in `_header.html.erb:10` to include the link to `/prisons/CURRENT/dashboard`.

PR #615, opened 2019-07-09
Search on caseload page

starswan:
Added searching and sorting to caseload page.

### DIFF
--- a/app/controllers/caseload_controller.rb
+++ b/app/controllers/caseload_controller.rb
@@ -9,34 +9,75 @@ class CaseloadController < PrisonsApplicationController
 
   PAGE_SIZE = 10
 
+  # rubocop:disable Metrics/MethodLength
   def index
-    return if pom.blank?
-
-    @allocations = PrisonOffenderManagerService.get_allocated_offenders(
-      pom.staff_id, active_prison,
-      offset: offset, limit: PAGE_SIZE
-    )
-    @new_cases_count = PrisonOffenderManagerService.get_new_cases_count(
-      pom.staff_id, active_prison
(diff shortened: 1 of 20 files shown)


app/controllers/caseload_controller.rb
```diff
@@ -9,34 +9,74 @@ class CaseloadController < PrisonsApplicationController
...
+  def sort_allocations(allocations)
+    if params['sort'].present?
+      sort_field, sort_direction = params['sort'].split.map(&:to_sym)
+    else
+      sort_field = :last_name
+      sort_direction = :asc
+    end
+
+    # cope with nil values by sorting using to_s - only dates and strings in these fields
+    allocations = allocations.sort_by { |sentence| sentence.public_send(sort_field).to_s }
```
rossjones: Am I correct in thinking this might fail if the sentence object (an offender stub really) doesn't respond to `sort_field` ?

rossjones:
I've tried this out, and a couple of things I noticed running it that I didn't notice in review.  I think the intended (but not documented) behaviour is that the search term should appear in the search box when showing results in the same way that the main search (`/prisons/LEI/search?utf8=✓&q=g34`) does.

Also, it's missing styling on the search results. It should be easy to get that with the highlight helper which takes some text and the search term and adds `<mark>` around the search term. e.g. ```highlight(offender.full_name, @q)```

app/controllers/caseload_controller.rb
starswan: yes it will. Not sure how to deal with that though..? The sort_field parameter is always hard-coded in the view, so shouldn't ever fail. There was more of an issue around nil values - nil dates don't compare with each other.

PR #622, opened 2019-07-12
Include allocation override detail in allocation history

KPobeeNorris:
When an SPO is allocating a POM they have the option to override the
recommended POM type, and are then asked to select a reason/reasons
for this decision.  This needs to be displayed on the allocation history
page, specifically when showing information related to allocating a
primary POM.

### DIFF
--- a/app/helpers/override_helper.rb
+++ b/app/helpers/override_helper.rb
@@ -10,47 +10,53 @@ module OverrideHelper
   end
 
   def display_override_pom(allocation)
-    if allocation.recommended_pom_type.present? && allocation.recommended_pom_type == 'prison'
-      "Probation POM allocated instead of recommended Prison POM"
-    elsif allocation.recommended_pom_type.present? && allocation.recommended_pom_type == 'probation'
-      "Prison POM allocated instead of recommended Probation POM"
+    if allocation.recommended_pom_type.present? &&
+      allocation.recommended_pom_type == 'prison'
+      'Probation POM allocated instead of recommended Prison POM'
+    elsif allocation.recommended_pom_type.present? &&
+      allocation.recommended_pom_type == 'probation'
+      'Prison POM allocated instead of recommended Probation POM'
     else
-      "Prisoner not allocated to recommended POM"
+      'Prisoner not allocated to recommended POM'
     end
   end
 
   def display_override_details(reason, allocation)
     case reason
-    when "no_staff"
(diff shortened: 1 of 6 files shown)


app/helpers/override_helper.rb
```diff
@@ -8,4 +8,55 @@ def complex_reason_label(recommended_type)
 
     'Prisoner assessed as suitable for a prison officer POM despite tiering calculation'
   end
+
+  def display_override_pom(allocation)
+    if allocation.recommended_pom_type.present? &&
```
starswan: You don't need the present? test when checking for equality...

spec/helpers/override_helper_spec.rb
```diff
@@ -1,6 +1,32 @@
 require 'rails_helper'
 
 RSpec.describe OverrideHelper do
+  let(:allocation_one) {
```
starswan: more idiomatic to call 'build' on a factory so that the test only sets the fields it cares about?

app/views/allocations/_allocate_primary_pom.html.erb
```diff
@@ -5,7 +5,15 @@
     - <%= mail_to(@pom_emails[allocation.primary_pom_nomis_id]) %>
   <% end %>
   <br/>
-  Tier: <%= allocation.allocated_at_tier %>
+  Tier: <%= allocation.allocated_at_tier %></p>
+  <% if allocation.override_reasons.present? %>
+    <p class="govuk-body govuk-!-margin-bottom-1">
+      <%= display_override_pom(allocation) %></p>
+    <p class="govuk-body govuk-!-margin-bottom-1">Reason(s):</p>
+    <% JSON.parse(allocation.override_reasons).each do | reason| %>
```
starswan: This doesn't look right - none of the other code needs to JSON.parse override_reasons. I think it might be YAML serialised...?

spec/helpers/override_helper_spec.rb
KPobeeNorris: RSpec complains with: 
`build` is not available on an example group (e.g. a `describe` or `context` block). It is only available from within individual examples (e.g. `it` blocks). 
 As I use them across two different describe blocks I pulled them out to avoid code duplication

KPobeeNorris: Scrap that - done!

PR #633, opened 2019-07-16
Add a rake task for downloading the latest delius attachment

rossjones:
This PR adds a new rake task `delius_email:fetch_latest` that will
connect to the appropriate IMAP server, navigate to a specific folder
and then download the latest `application/zip` attachment it can find.

The single argument to the rake task, is the name to give to the
downloaded attachment.  The username, password and folder to check are
specified in environment variables.  These are `DELIUS_EMAIL_USERNAME`,
`DELIUS_EMAIL_PASSWORD`, and `DELIUS_EMAIL_FOLDER`.

To test that this works correctly the `Net::IMAP` client is mocked and
the key methods implemented, working with json fixtures on disk so no
access to a working IMAP server is required.  The folders in the 'IMAP
server' are folders in the `fixtures/imap/` directory, each JSON file
represents a message where the first part of the filename represents the
IMAP message id.

## TODO

- [x] Better documentation
~~[ ] Handle cleanup of old messages~~


spec/lib/delius/emails_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+    }
+  end
+
+  it 'can sort messages in a folder' do
+    described_class.connect(username, password) { |emails|
+      emails.folder = 'sorted_small'
+      messages = emails.sorted_mail_messages
+
+      expect(messages.count).to eq(3)
+      expect(messages[0].date).to eq(Date.parse('21/06/2019'))
```
starswan: I think this could be: expect(messages.map(&:date)).to eq([Date.parse('21/06/2019'), Date.parse('21/1/2019'), Date.parse('10/1/2019')])
so that we can see that they are in date-descending order

lib/delius/emails.rb
```diff
@@ -0,0 +1,95 @@
...
+    end
+
+    def folder=(folder)
+      @imap.examine(folder)
+    end
+
+    def latest_attachment
+      messages = sorted_mail_messages
+      messages.find do |msg|
+        attachment = zip_attachment(msg)
```
starswan: do we use detect (used in zip_attachment) rather than find for arrays?

PR #636, opened 2019-07-17
Create coworking pom

theKHutDeveloper:
### Add pages 'Allocate a co-working Prison Offender Manager' and 'Confirm co-working allocation'

**_- Link to creating a co-working POM is added to Allocation Information page_**

Screen Shot 2019-07-17 at 12 26 15

 **_- Allocate a co-working POM page_**

Screen Shot 2019-07-17 at 10 12 04

**_- Confirm co-working POM page_** 

Screen Shot 2019-07-17 at 10 15 58


app/controllers/coworking_controller.rb
```diff
@@ -0,0 +1,62 @@
...
+  def confirm
+    @prisoner = offender(nomis_offender_id_from_url)
+    @primary_pom = PrisonOffenderManagerService.get_pom(
+      prison_id_from_url, primary_pom_id_from_url
+    )
+    @secondary_pom = PrisonOffenderManagerService.get_pom(
+      prison_id_from_url, nomis_staff_id_from_url
+    )
+
+    @event = :allocate_secondary_pom
```
rossjones: Might it be better to use the constants in AllocationVersion for these?

app/controllers/coworking_controller.rb
```diff
@@ -0,0 +1,62 @@
...
+
+  def nomis_staff_id_from_url
+    params.require(:secondary_pom_id)
+  end
+
+  def primary_pom_id_from_url
+    params.require(:primary_pom_id)
+  end
+
+  def current_pom_for(nomis_offender_id)
```
rossjones: I think we have/use this functionality in another place as well, is it worth adding to AllocationService if it isn't there already?

app/views/coworking/new.html.erb
```diff
@@ -0,0 +1,37 @@
...
+<div class="govuk-grid-row"><div class="govuk-grid-column-full govuk-!-margin-top-4"></div></div>
+
+<h2 class="govuk-heading-s">Prison Officer POMs</h2>
+
+<%= render :partial => 'poms_table', :locals => { poms: @prison_poms } %>
+
+<!-- Spacer -->
+<div class="govuk-grid-row"><div class="govuk-grid-column-full govuk-!-margin-top-4"></div></div>
+
+<% if @unavailable_poms.count > 0 %>
```
starswan: There's an idiom for this: https://apidock.com/rails/ActionView/Helpers/TextHelper/pluralize

spec/features/coworking_feature_spec.rb
```diff
@@ -0,0 +1,69 @@
+require 'rails_helper'
+
+feature 'Co-working' do
```
starswan: I would have expected the 'feature' test to be more 'feature-y' e.g. load the page, allocate a co-worker, look at the allocations page to see that it shows up.

theKHutDeveloper: @starswan I don't think that I can allocate a co-worker at the moment as their needs to be changes to the create_or_update method in AllocationService, currently it only adds/updates on primary_pom

starswan: Ok understand

rossjones:
+1 just hope it doesn't turn into a merge nightmare if we leave it too long.

app/controllers/coworking_controller.rb
theKHutDeveloper: @rossjones We can't actually use the constants in AllocationVersion as we need to pass the event hash e.g event[:allocate_primary_pom]. Seems like AllocationVersion then converts that hash into an integer value. The reason the tests currently pass is because this PR does not include the creation of the co-worker so we are not saving an AllocationVersion. I will had another commit to revert the changes made.

PR #645, opened 2019-07-22
first cut of mapping MAPPA level in from nDelius

starswan:
MAPPA levels 1,2 and 3 mapped only

### DIFF
--- a/app/helpers/offender_helper.rb
+++ b/app/helpers/offender_helper.rb
@@ -54,7 +54,7 @@ private
     if SentenceTypeService.indeterminate_sentence?(offender.imprisonment_status)
       [indeterminate_handover_date(offender), 'NPS Inderminate']
     elsif offender.parole_eligibility_date.present?
-      [parole_handover_date(offender), 'NPS Determinate Parole Case']
+      [offender.parole_eligibility_date - 8.months, 'NPS Determinate Parole Case']
     elsif offender.mappa_level == 1 || offender.mappa_level.blank?
       [mappa1_handover_date(offender), 'NPS Determinate Mappa 1/0']
     else
@@ -86,10 +86,6 @@ private
       crd_ard,
       offender.home_detention_curfew_eligibility_date
     ].compact.min
-  end
-
-  def parole_handover_date(offender)
-    offender.parole_eligibility_date - 8.months
   end
 
   def early_allocation_handover_date(offender)
--- a/app/jobs/process_delius_data_job.rb
+++ b/app/jobs/process_delius_data_job.rb
@@ -1,0 +1,101 @@
+class ProcessDeliusDataJob < ApplicationJob
+  queue_as :default
+
+  include ApplicationHelper
+
+  def perform(nomis_offender_id)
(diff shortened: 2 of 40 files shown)


starswan:
not quite in the right place

app/helpers/offender_helper.rb
```diff
@@ -88,7 +92,7 @@ def mappa1_handover_date(offender)
     ].compact.min
   end
 
-  def parole_handover_date(offender)
+  def parole_responsibility_date(offender)
     offender.parole_eligibility_date - 8.months
```
rossjones: Do we care that this may be nil? Unfortunately we can't make any safe assumptions about anything that came from sentence details :(

app/jobs/process_delius_data_job.rb
```diff
@@ -0,0 +1,67 @@
...
+      else
+        process_record(delius_data.first)
+      end
+    end
+  end
+
+private
+
+  def process_record(delius_record)
+    case_information = CaseInformation.find_or_initialize_by(
```
rossjones: I think, unless we match on noms_no _and_ CRN we might not catch duplicates, that is more than one offender being given the noms_no in Delius.  This would just overwrite possibly the wrong offender I think.

db/migrate/20190725082404_add_ldu_and_team_to_case_information.rb
```diff
@@ -0,0 +1,8 @@
+class AddLduAndTeamToCaseInformation < ActiveRecord::Migration[5.2]
+  def change
+    change_table :case_information do |t|
+      t.string :ldu
```
rossjones: Are these (ldu and team) just the IDs?

starswan: These are the names - not sure that the IDs are useful..? Actually I don't know which ones I am meant to be bringing in.

app/jobs/process_delius_data_job.rb
starswan: Yes that thought crossed my mind in the last couple of days too.

db/migrate/20190725082404_add_ldu_and_team_to_case_information.rb
rossjones: The problem with the names is that they're inconsistent in the data file, for instance Gwent appears as "Gwent" and "Gwent LDU".  Once we get to https://dsdmoj.atlassian.net/browse/POM-187 we'll need to be able to get the email address of the LDU.  Maybe we should worry about that once we know what the email addresses are keyed against?

starswan: It's not clear what we want to do here. We're not doing anything with the data once we bring it in - we're just preventing allocations unless it has something?

rossjones: I think one of the pages might have it used in display (prisoner profile?) but that's probably a later thing.

PR #646, opened 2019-07-23
Adds a rake task to run the delius ETL

rossjones:
The included rake task `rake delius_etl:process` will run all of the
steps necessary to update the case information for offenders, from a
daily dump of delius data.

These steps mostly exist as existing rake tasks, and this task
integrates them so that they are all executed and any downloaded data is
cleaned up afterwards.

The steps it performs are:

* Download the latest attachment
* Unzip the attachment
* Decrypt the unzipped file
* Load the decrypted file into the database
* Create CaseInformation from the newly loaded data (pending).

As soon as the process has completed (or failed) the working folder
where files were stored is deleted.

The README.md file has been updated with required environment variables
for the ETL process.


starswan:
It would be nice IMHO to put all the tasks under a common namespace.

lib/delius/etl.rb
```diff
@@ -0,0 +1,95 @@
...
+    #
+    #   mkdir ./build
+    #   cd ./build
+    #   git clone https://github.com/herumi/cybozulib
+    #   git clone https://github.com/herumi/msoffice
+    #   cd msoffice
+    #   make -j RELEASE=1
+    #   mv ./bin/msoffice-crypt.exe <SOMEWHERE_IN_PATH>/mssoffice-crypt
+    #
+    def decrypt_xlsx(encrypted_filename, decrypted_filename)
```
starswan: There is already a delius_decrypt:decrypt rake task...

starswan:
Superceded by other branch with all tasks as an active job

PR #653, opened 2019-07-25
create case information from nDelius data

starswan:


app/models/delius_data.rb
```diff
@@ -10,19 +10,16 @@ class DeliusData < ApplicationRecord
     end
   end
 
-  # TODO: this will need changing in Rails 6 as it actually has an upsert method.
-  def self.upsert(record)
-    DeliusData.find_or_initialize_by(crn: record[:crn]).update(record.without(:crn))
-  end
-
   def omicable?
+    # WPT is the ldu_code for all Welsh prisons
```
rossjones: Sorry missed this.

```suggestion
    # WPT is the first 3 chars of the ldu_code for all Welsh LDU (Local Divisional Unit).
```

app/helpers/application_helper.rb
```diff
@@ -71,6 +71,11 @@ def humanized_bool(bool_value)
     bool_value ? 'Yes' : 'No'
   end
 
+  def auto_delius_import_enabled?(prison)
+    Flipflop.auto_delius_import? ||
+      (ENV['AUTO_DELIUS_IMPORT'] || '').split(',').include?(prison)
```
rossjones: If we want to set this env var, it'll need defining in https://github.com/ministryofjustice/offender-management-allocation-manager/blob/master/deploy/staging/shared-environment.yaml (or the prod/preprod equivalent).

starswan: Yes - this is designed as a slow-rollout to a few prisons to see how they cope

app/views/summary/pending.html.erb
```diff
@@ -50,7 +50,13 @@
           <% end %>
         </td>
         <td aria-label="Waiting" class="govuk-table__cell"><%= offender.awaiting_allocation_for %> days</td>
-        <td aria-label="Action" class="govuk-table__cell "><%= link_to "Edit", new_prison_case_information_path(@prison, offender.offender_no) %></td>
+        <td aria-label="Action" class="govuk-table__cell ">
+          <% if auto_delius_import_enabled?(@prison) %>
+            <%= link_to "Update", prison_case_information_path(@prison, offender.offender_no) %>
```
rossjones: Is this View rather than Update?

starswan: That's what was in the design...

PR #658, opened 2019-07-26
Add content to Allocation History deallocation events

theKHutDeveloper:
Screen Shot 2019-07-26 at 15 33 22

When a POM or a co-worker has been deallocated (this does
not include when a prisoner has been transferred), add
content detailing the name and email address (if present)
of the POM or co-worker.

Move the functionality of determining whether the POMs
email address is present or not from the views into a
helper method.

### DIFF
--- a/app/views/allocations/_deallocate_pom.html.erb
+++ b/app/views/allocations/_deallocate_pom.html.erb
@@ -5,7 +5,11 @@
     <% end %>
   </h2>
   <p>
-  &nbsp;
+    <% unless allocation.event_trigger == 'offender_transferred' %>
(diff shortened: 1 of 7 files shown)


app/views/allocations/_deallocate_pom.html.erb
```diff
@@ -5,7 +5,11 @@
     <% end %>
   </h2>
   <p>
-  &nbsp;
+    <% unless allocation.event_trigger == 'offender_transferred' %>
```
starswan: This raises the interesting question of what we should display when someone is transferred - the history will look strange maybe?

PR #666, opened 2019-07-30
spike for webmock usage

starswan:
One test reworked from VCR to webmock. Much more visibility and control over what
happens depending on what data is returned


rossjones:
This looks like it could work ;) How do we envision this working when we have a lot of data to return from the API call?  Could we load some data from fixtures for mocks?

PR #680, opened 2019-08-01
Do not process movement with no target agency

rossjones:
When a movement event is found, but there is no target agency, we will not set the prison for that offender.


app/models/allocation_version.rb
```diff
@@ -117,7 +117,6 @@ def self.deallocate_offender(nomis_offender_id, movement_type)
     alloc.recommended_pom_type = nil
     alloc.event = DEALLOCATE_PRIMARY_POM
     alloc.event_trigger = movement_type
```
KPobeeNorris: @rossjones - now you're removing this line you can also reset the validation on line 148.  The validation was amended when this line was originally added so can also be updated?

PR #687, opened 2019-08-05
Streamline CI so we don't have to wait so long

rossjones:
Rather than wait for three tasks to complete for rubocop,
security-analysis and tests, put them all in one job. We order the jobs
to fail fast.

We will need to change the github branch settings so that it doesn't wait for ever for the now non-existent jobs to complete.


starswan:
I'm not sure that this actually improves anything speed-wise...

rossjones:
It reduces the number of containers we use, which is limited in out MoJ account (and shared with all the other projects).  So it's not so much about speed, but rather queues :D 

And the ordering is so you don't want 5 minutes to find out rubocop failed.

PR #697, opened 2019-08-08
convert Movement rake task to 2 jobs 

starswan:
-one to kick-off the day, and one to do each movement so that 1 failed movement record doesn't prevent the rest of the movements from being processed. 
The desired behaviour is that a failed movement still creates a Sentry alert.

### DIFF
--- a/app/jobs/movement_job.rb
+++ b/app/jobs/movement_job.rb
@@ -1,0 +1,7 @@
+class MovementJob < ApplicationJob
+  queue_as :default
+
+  def perform(movement)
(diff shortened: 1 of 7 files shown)


lib/tasks/delius_trigger.rake
```diff
@@ -0,0 +1,14 @@
+namespace :delius do
```
rossjones: Is this on the correct branch?

starswan: No you're right it belongs on the nDelius branch...

PR #732, opened 2019-08-16
Increase number of offenders displayed per page

KPobeeNorris:
We have received feedback that users are finding our current pagination
system frustrating as you can only go through one page at a time, or
jump to the very beginning, or very end.  This means that if they are
trying to get to surnames beginning with "M" it can take a lot of time.
As a temporary solution we are going to start displaying 20 offenders
per page, rather than the 10 we do at present.  This will reduce the
number of times they will need to click "next" or "previous" to get to a
group of particular offenders.

This will then be raised with our designer so we can find a better
solution to make it better for our users.


KPobeeNorris:
Issues with specs taking a long time to run, and a couple of them are flaky so want to look into it a bit more

PR #735, opened 2019-08-20
[POM-298] Add debugging page

KPobeeNorris:
During the onboarding process we have received a number of queries from
users in regards to missing offenders, offenders with incorrect
information and so forth.  For example: for an offender to show up in
our service they need to be over 18, convicted, sentenced, etc.  At the
moment a dev needs to go and run a number of commands to try and find
out the various pieces of information and pass it along to whoever is on
duty.  Therefore, it was agreed that we should add a debugging page,
that anyone in the team can access, that provides a breakdown of the
individual offender to help us answer these types of queries.

### DIFF
--- a/app/controllers/debugging_controller.rb
+++ b/app/controllers/debugging_controller.rb
@@ -5,7 +5,7 @@ class DebuggingController < PrisonsApplicationController
     nomis_offender_id = id
 
     @offender = offender(nomis_offender_id)
-    unless @offender.blank?
+    if @offender.present?
       @allocation = AllocationVersion.where(nomis_offender_id: @offender.offender_no)
       @movements = Nomis::Elite2::MovementApi.movements_for(@offender.offender_no).first
     end
--- a/app/controllers/debugging_controller.rb
+++ b/app/controllers/debugging_controller.rb
@@ -1,18 +1,17 @@
 # frozen_string_literal: true
 
 class DebuggingController < PrisonsApplicationController
-
   def debugging
     nomis_offender_id = id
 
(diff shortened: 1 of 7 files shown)


KPobeeNorris:
This PR relies on changes implemented in https://github.com/ministryofjustice/offender-management-allocation-manager/pull/736 and will need to be rebased after that PR has been approved/merged.

app/services/nomis/models/offender_base.rb
```diff
@@ -44,6 +44,10 @@ def sentenced?
         SentenceTypeService.indeterminate_sentence?(imprisonment_status)
       end
 
+      def over_18?
+        age > 18
```
rossjones:
```suggestion
        age >= 18
```

app/controllers/debugging_controller.rb
```diff
@@ -0,0 +1,25 @@
...
+  end
+
+private
+
+  def id
+    params[:offender_no]
+  end
+
+  def offender(offender_no)
+    return [] if offender_no.blank?
```
rossjones:
```suggestion
    return nil if offender_no.blank?
```

app/controllers/debugging_controller.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+class DebuggingController < PrisonsApplicationController
+  def debugging
+    nomis_offender_id = id
+
+    @offender = offender(nomis_offender_id)
+    if @offender.present?
```
rossjones:
```suggestion
    unless @offender.blank?
```

PR #736, opened 2019-08-20
[POM-303] Update OffenderApi

KPobeeNorris:
Whilst working on ticket POM-298 we found out that the endpoint to fetch
a single offender on Elite2 ("/elite2api/api/prisoners/OFFENDER_NO") now
returns 200 with an empty array/list, whereas it previously returned a
500 if the offender number was malformed, or the offender didn't exist.
This means that we can remove the NullOffender object and handle not
finding offenders differently.

### DIFF
--- a/app/services/nomis/elite2/offender_api.rb
+++ b/app/services/nomis/elite2/offender_api.rb
@@ -42,14 +42,10 @@ module Nomis
 
       def self.get_offender(offender_no)
         route = "/elite2api/api/prisoners/#{URI.encode_www_form_component(offender_no)}"
-        response = e2_client.get(route) { |data|
-          raise Nomis::Client::APIError, 'No data was returned' if data.empty?
-        }
+        response = e2_client.get(route)
+        return '' if response.empty?
 
         api_deserialiser.deserialise(Nomis::Models::Offender, response.first)
-      rescue Nomis::Client::APIError => e
-        AllocationManager::ExceptionHandler.capture_exception(e)
-        Nomis::Models::NullOffender.new
       end
 
       def self.get_offence(booking_id)
--- a/app/services/nomis/elite2/offender_api.rb
+++ b/app/services/nomis/elite2/offender_api.rb
@@ -43,7 +43,7 @@ module Nomis
       def self.get_offender(offender_no)
         route = "/elite2api/api/prisoners/#{URI.encode_www_form_component(offender_no)}"
         response = e2_client.get(route)
-        return '' if response.empty?
+        return nil if response.empty?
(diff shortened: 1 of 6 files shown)


spec/services/nomis/elite2/offender_api_spec.rb
```diff
@@ -59,13 +59,13 @@
       expect(response).to eq('C')
     end
 
-    it 'returns null if unable to find prisoner', :raven_intercept_exception,
+    it 'returns if unable to find prisoner',
```
rossjones: Although this tests the OffenderApi, is it worth having a test for OffenderService.get_offender() as well?

app/services/offender_service.rb
```diff
@@ -3,6 +3,8 @@
 class OffenderService
   def self.get_offender(offender_no)
     Nomis::Elite2::OffenderApi.get_offender(offender_no).tap { |o|
+      next false if o.blank?
```
rossjones: Does this need to be `next false` instead of just `next`?

app/services/nomis/elite2/offender_api.rb
```diff
@@ -42,14 +42,10 @@ def self.list(prison, page = 0, page_size: 10)
 
       def self.get_offender(offender_no)
         route = "/elite2api/api/prisoners/#{URI.encode_www_form_component(offender_no)}"
-        response = e2_client.get(route) { |data|
-          raise Nomis::Client::APIError, 'No data was returned' if data.empty?
-        }
+        response = e2_client.get(route)
+        return '' if response.empty?
```
rossjones:
```suggestion
        return nil if response.empty?
```

spec/services/nomis/elite2/offender_api_spec.rb
```diff
@@ -59,13 +59,13 @@
...
        vcr: { cassette_name: :offender_api_null_offender_spec  } do
       noms_id = 'AAA22D'
 
       response = described_class.get_offender(noms_id)
 
-      expect(response).to be_instance_of(Nomis::Models::NullOffender)
+      expect(response).to eq ""
```
rossjones:
```suggestion
      expect(response).to be_nil
```

app/services/offender_service.rb
```diff
@@ -3,6 +3,8 @@
 class OffenderService
   def self.get_offender(offender_no)
     Nomis::Elite2::OffenderApi.get_offender(offender_no).tap { |o|
+      next false if o.nil?
```
rossjones: Missed this one?

PR #737, opened 2019-08-20
Change the way get the allocated POM's name for an offender in SummaryService

theKHutDeveloper:
We were using the method `set_allocated_pom_name` in OffenderService but this is a painfully slow function call as it re-fetches the allocation for each offender being shown in the allocated tab to work out the name of the POM.

This commit removes the call to `set_allocated_pom_name` and grabs the primary_pom_name and primary_pom_allocated_at date from the allocations we have already obtained.

`OffenderService.set_allocated_pom_name` is used in SearchController so it has been left untouched for now.

### DIFF
--- a/app/services/summary_service.rb
+++ b/app/services/summary_service.rb
@@ -64,13 +64,14 @@ class SummaryService
 
     # For the allocated offenders, we need to provide the allocated POM's
     # name
+    offender_items = bucket.take(wanted_items, from) || []
+
     if summary_type == :allocated
-      offender_items = OffenderService.set_allocated_pom_name(
-        bucket.take(wanted_items, from) || [],
-        prison
-      )
-    else
-      offender_items = bucket.take(wanted_items, from) || []
+      offender_items.each { |offender|
+        alloc = active_allocations_hash[offender.offender_no]
+        offender.allocated_pom_name = alloc.primary_pom_name.titleize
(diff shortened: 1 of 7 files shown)


app/services/summary_service.rb
```diff
@@ -64,13 +64,14 @@ def self.summary(summary_type, prison, page, params)
...
-      offender_items = OffenderService.set_allocated_pom_name(
-        bucket.take(wanted_items, from) || [],
-        prison
-      )
-    else
-      offender_items = bucket.take(wanted_items, from) || []
+      offender_items.each { |offender|
+        alloc = active_allocations_hash[offender.offender_no]
+        offender.allocated_pom_name = alloc.primary_pom_name.titleize
+        offender.allocation_date = alloc.primary_pom_allocated_at
```
rossjones: Do we know that alloc.primary_pom_allocated_at is set for all allocations?  What should we use otherwise?

rossjones: Looks like this is already using this in the current code and so it's fine. 

For the record, this drops the number of SQL queries in the allocated summary from 37 to 11 which is a useful and significant drop.

rossjones:
It's likely we'll need to change the order of the POM's name when saving, and then add a workaround for existing names to swap the name around - so please don't merge yet.

PR #740, opened 2019-08-21
[POM-287] Increase offenders displayed per page

KPobeeNorris:
We discussed the option of increasing the number of offenders we display on a page, upping it from 10 to 20.  This is a temporary solution whilst further research is done into how we can improve the pagination situation.


rossjones:
Does this have any effect on performance?

rossjones:
@KPobeeNorris looks like this missed `p.size = 10` in summary.rb as it doesn't work if there are less than 20 items

PR #741, opened 2019-08-23
Waiting to allocate date

theKHutDeveloper:
Currently the "Waiting to Allocate" date shows the date the offender was sentenced, this PR is to provide a more accurate reflection of the actual time the offender entered the current prison as offenders are often transferred to other prisons during their stay.

This PR amends the MovementAPI movement_for method to accept an array of offenders to obtain the prison the offender is currently located at. This information is used to determine when the waiting to allocate date occurs at either from the omic start date or the date the offender entered the prison.

### DIFF
--- a/app/services/summary_service.rb
+++ b/app/services/summary_service.rb
@@ -85,7 +85,7 @@ class SummaryService
 
     offenders = offender_list.map(&:offender_no)
 
-    return if offenders.nil?
+    return if offenders.empty?
 
     omic_start_date = Date.new(2019, 2, 4)
     movements = Nomis::Elite2::MovementApi.movements_for(offenders)
--- a/app/services/summary_service.rb
+++ b/app/services/summary_service.rb
@@ -84,24 +84,16 @@ class SummaryService
     return if offender_list.empty?
 
     offenders = offender_list.map(&:offender_no)
-    prison_dates = []
 
-    unless offenders.nil?
-      omic_start_date = Date.new(2019, 2, 4)
-      movements = Nomis::Elite2::MovementApi.movements_for(offenders)
+    return if offenders.nil?
 
-      prison_dates = movements.map { |movement|
-        if movement.create_date_time.nil? ||
-            movement.create_date_time < omic_start_date
-          start_date = (Time.zone.today - omic_start_date).to_i
-        else
-          start_date = (Time.zone.today - movement.create_date_time).to_i
-        end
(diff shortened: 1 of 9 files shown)


spec/helpers/summary_helper_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+          { offender_no: "G9765JZ", days_count: 50 }
+      ]
+    end
+
+    it "formats an offender's prison start date if offender exists" do
+      expect(start_date(prison_dates, "A1120GH")).to eq(199)
+    end
+
+    it "formats an offender's prison start date if offender does not exist" do
+      expect(start_date(prison_dates, "Z1234WQ")).to eq('-')
```
rossjones: Presumably we wouldn't ever actually see this offender in the list, so this is the only way to ever see '-'?  

I think returning a string or an integer might be confusing, so perhaps we can just choose a sentinel value (like 0) as it'll never actually be shown.

app/services/summary_service.rb
```diff
@@ -75,10 +75,26 @@ def self.summary(summary_type, prison, page, params)
...
+  # rubocop:enable Metrics/LineLength
+  # rubocop:enable Metrics/MethodLength
+  # rubocop:enable Metrics/PerceivedComplexity
+
+  def self.entered_prison_dates(offender_list)
+    return if offender_list.empty?
+
+    offenders = offender_list.map(&:offender_no)
+
+    return if offenders.nil?
```
starswan: This can't actually happen - map cannot return nil

app/services/summary_service.rb
```diff
@@ -75,10 +75,26 @@ def self.summary(summary_type, prison, page, params)
...
+
+    omic_start_date = Date.new(2019, 2, 4)
```
starswan: I think we have a constant for this date somewhere complete with a comment?

rossjones: That would certainly help with POM-290. The start dates look like they may end up being different for England and Wales, so it will end up being prison-dependent - so might want to put the constant somewhere other than summary or handoverdate services.

app/services/summary_service.rb
theKHutDeveloper: Fixed in the latest commit

app/services/summary_service.rb
theKHutDeveloper: Happy to do this in a separate PR

PR #749, opened 2019-08-29
Non-omicable offenders are always SUPPORTING responsibility

rossjones:


rossjones:
It appears the Custody/Community responsibility is determined from the Responsible/Supporting value - which means for omicable offenders this will always be wrong. 

Boo....

rossjones:
We will need to change the Custody/Community responsibility based on the handover date for the offender.

rossjones:
Don't think this is required any longer, it has been rewritten.

PR #760, opened 2019-09-03
Create Maintenance page

theKHutDeveloper:
This PR creates a maintenance page designed to be used if the service is unavailable

### Page view
Screen Shot 2019-09-05 at 11 39 07


rossjones:
I think the content may need to change, it talks about us fixing a problem, but I think the point of the maintenance is that we're doing some planned work. Obvs check with LM, but I think something more along the lines of "The system is currently undergoing planned maintenance and will be available again shortly".

It might also be worth updating the wiki to describe how we turn on maintenance mode (based on the turnout docs) somewhere underneath the `Dev book` folder.

PR #773, opened 2019-09-06
Handover date calculation

theKHutDeveloper:
Currently the calculation of the offender’s handover date is done in a view helper.
This PR moves the handover calculations to a service - HandoverDateService. The tests
associated with the handover calculations have been moved out of the helper_spec and into
the service_spec.

The handover methods have now been added to the OffenderBase class and called directly
in the views.


app/services/handover_date_service.rb
```diff
@@ -0,0 +1,85 @@
+# frozen_string_literal: true
+
+class HandoverDateService
+  # rubocop:disable Metrics/LineLength
+  def self.handover_start_date(offender)
+    if offender.nps_case? && !offender.early_allocation?
+      return [offender.earliest_release_date - 8.months, 'NPS Indeterminate'] if SentenceTypeService.indeterminate_sentence?(offender.imprisonment_status)
```
rossjones: Better to have this as an if block rather than have the linelength exclusion for rubocop I think...

PR #777, opened 2019-09-09
[POM-307] Zendesk implementation

KPobeeNorris:
This commit sets up the codebase for the implementation of Zendesk by:
* Adding the zendesk api gem
* Adding the zendesk shared envs to the deployment envs
* Updating the routes
* Creating a new table to the database via a migration
* Adding the zendesk api and client files
* Removing the old contact model and replacing it with the
contact_submission model
* Updating the 'help' view to use the newly designed form
* Creating the Zendesk ticket job
* Updating the 'user_details' model to fetch the staff's email address


app/controllers/pages_controller.rb
```diff
@@ -3,27 +3,55 @@
...
+                                       prison: PrisonService.name_for(
+                                         @user.active_case_load_id)
+      )
+    else
+      @contact = ContactSubmission.new
+    end
+  end
+
+  def create_help
+    @user = Nomis::Elite2::UserApi.user_details(current_user) if current_user.present?
```
rossjones: What is this used for?

KPobeeNorris: To create the actual help form and trigger the creation of the Zendesk ticket if it's correctly formatted i.e. has all the right information filled in

PR #794, opened 2019-09-16
Fix debugging page

KPobeeNorris:
A recent change to the MovementApi mean that we started getting all
movements for an offender, rather than just the latest.  The debugging
page displays the last movement, but as they're not returned in date
order it meant it could be showing an old movement.  Therefore this PR
adds the api call to get just the latest movement for an offender to
ensure the debugging page shows the right information.

### DIFF
--- a/app/services/nomis/elite2/movement_api.rb
+++ b/app/services/nomis/elite2/movement_api.rb
@@ -26,6 +26,15 @@ module Nomis
           api_deserialiser.deserialise(Nomis::Movement, movement)
         }
       end
+
+      def self.latest_movement_for(offender_no)
+        route = '/elite2api/api/movements/offenders?movementTypes=ADM&movementTypes=TRN&movementTypes=REL'
+
+        data = e2_client.post(route, [offender_no])
+        data.sort_by { |k| k['movementTime'] }.map{ |movement|
+          api_deserialiser.deserialise(Nomis::Movement, movement)
+        }
+      end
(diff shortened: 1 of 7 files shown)


app/services/nomis/elite2/movement_api.rb
```diff
@@ -26,6 +26,15 @@ def self.movements_for(offender_no)
           api_deserialiser.deserialise(Nomis::Movement, movement)
         }
       end
+
+      def self.latest_movement_for(offender_no)
```
rossjones: Seems a shame to have the same code as the other function, might be worth replacing with something like ...

```suggestion
    movements_for(offender_no)&.first
```

KPobeeNorris: The issue is that the API doesn't return the movements in a date ordered list, so you can't guarantee that the first/last thing in the list is going to actually be the latest movement

app/services/nomis/elite2/movement_api.rb
```diff
@@ -26,6 +26,15 @@ def self.movements_for(offender_no)
...
+        route = '/elite2api/api/movements/offenders?movementTypes=ADM&movementTypes=TRN&movementTypes=REL'
```
rossjones: Seems a shame to have the same code as the other function, might be worth replacing with something like ...

```suggestion
    movements_for(offender_no)&.first
```

PR #797, opened 2019-09-16
Pom 309 new pagination

starswan:
New pagination (done via Kaminari gem)


app/views/shared/_kaminari.html.erb
```diff
@@ -0,0 +1,15 @@
+<div class="govuk-grid-row pagination">
+  <div class="govuk-grid-column-one-half">
+    <div class="govuk-grid-row">
```
rossjones: This div seems to push the pages too far to the left and slightly up,  it looks fine after removing it though

PR #799, opened 2019-09-18
Internal prison debugging page.

rossjones:
As we don't show every single offender in prison, we have added a page
showing which offenders have been filtered out and for what reason.
Currently we only exclude offenders < 18yo, unsentenced offenders and
civil cases.  This page (/prisons/CODE/debugging/prison) is currently
only intended for internal use (although it is not restricted).


app/services/offender_service.rb
```diff
@@ -73,13 +73,9 @@ def get_offenders_for_prison(page_number:, page_size:)
       mapped_tiers = CaseInformationService.get_case_information(nomis_ids)
 
       offenders.select { |offender|
```
starswan: This should now be an 'each' rather than a select.

starswan: This can now be an 'each' - a bit naughty of it to be a select that changes the iterable...

PR #805, opened 2019-09-19
ability to override responsibility

starswan:


app/views/responsibilities/new.html.erb
```diff
@@ -0,0 +1,66 @@
...
+          <% if @responsibility.errors[:reason_text].present? %>
+            <span class="govuk-error-message">
+              <%= @responsibility.errors[:reason_text].first %>
+            </span>
+          <% end %>
+          <div class="govuk-form-group">
+            <label class="govuk-label" for="other_reason_text">
+              Enter reason
+            </label>
+            <input class="govuk-input govuk-!-width-one-third govuk-textarea--error" id="other_reason_text" name="responsibility[reason_text]">
```
rossjones: Is this intended to have the `govuk-textarea--error` all the time? Rather than just when there are errors?

app/views/responsibilities/confirm.html.erb
```diff
@@ -0,0 +1,41 @@
...
+        <h1 class="govuk-fieldset__heading">
+          Confirm change of responsibility for this case
+        </h1>
+      </legend>
+    </fieldset>
+    <p class="govuk-body">
+      The community probation team will be responsible for this case
+    </p>
+    <p class="govuk-body">
+      We will email the community probation team at <%= @ldu_email_address %>.
```
rossjones: Was there any discussion on what should happen if no LDU email is available?  Presumably something like giving the LDU name, and then suggesting they contact them?

starswan: no you're right - I'm unsure what is supposed to happen here.

app/views/responsibilities/new.html.erb
starswan: This actually mirrors the other screen done like this - it would certainly be better if it only showed error when there was one I agree. Will look into it

app/views/responsibilities/confirm.html.erb
rossjones: I think @lmoney-moj will be able to help with content for that case.

app/models/nomis/offender_base.rb
```diff
@@ -75,6 +75,14 @@ def earliest_release_date
       sentence.earliest_release_date
     end
 
+    def sentence_start_date
+      sentence.sentence_start_date
+    end
+
+    def pom_responsibility
+      ResponsibilityService.calculate_pom_responsibility(self)
```
rossjones: I know it doesn't make an API call or anything, but is this worth memoizing?

app/services/responsibility_service.rb
```diff
@@ -15,7 +15,7 @@ def to_s
   # rubocop:disable Metrics/PerceivedComplexity
   # rubocop:disable Metrics/CyclomaticComplexity
   def self.calculate_pom_responsibility(offender)
-    return RESPONSIBLE if offender.sentence.earliest_release_date.nil?
+    return RESPONSIBLE if offender.earliest_release_date.nil?
     return SUPPORTING unless welsh_offender?(offender)
```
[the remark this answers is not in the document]
starswan: oh dear what are we supposed to do then? Oct 1st is only a week away...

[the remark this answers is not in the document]
rossjones: I'll submit a separate PR we can deploy next week.

PR #806, opened 2019-09-19
[POM-342] Provides a rake task for privately loaded LDU email addresses

rossjones:
This PR contains a rake task to enable the loading of a CSV containing
LDU email addresses into the 'local_divisional_units' table by matching
the existing entries to the rows in the CSV.  The CSV must have the
following format (where heading text is not important):

```
heading1,heading2
lducode1,emailaddress
lducode2,emailaddress
```

The matching from CSV to DB row is via the ldu code. Valid entries in
the CSV (those containing an email address) will attempt to find the
LocalDivisionalUnit object that contains that code, and will then update
the email address.

This task will never set an email address to nil, and will not create
new database records. It can be executed with

```
rails ldu_email:import[path-to-csv.csv]
```


lib/ldu_email_importer.rb
```diff
@@ -0,0 +1,16 @@
...
+
+class LDUEmailImporter
+  def self.import(filename)
+    CSV.read(filename, headers: true).each do |row|
+      code = row[0]&.strip
+      address = row[1]&.strip
+
+      next if address.blank?
+
+      LocalDivisionalUnit.where(code: code).update_all(email_address: address)
```
starswan: update_all isn't a good method to use except in a migration - it bypasses validations. Just update! is probably fine here.

PR #809, opened 2019-09-20
Pom 290 new responsibility calculations

starswan:
New responsibility calculations


spec/services/responsibility_service_spec.rb
```diff
@@ -1,166 +1,332 @@
...
+            it 'is supporting' do
               resp = described_class.calculate_pom_responsibility(offender_welsh_nps_new_case_lt_10_mths)
 
-              expect(resp).to eq 'Supporting'
+              expect(resp).to eq ResponsibilityService::SUPPORTING
             end
           end
         end
 
         context 'when old case (sentence date before February 4 2019)' do
```
rossjones: Do we want a 'if new case' and if 'old case' test for English offenders as well?

app/services/responsibility_service.rb
```diff
@@ -6,39 +6,94 @@ class ResponsibilityService
   COWORKING = 'Co-Working'
   NPS = 'NPS'
 
-  # rubocop:disable Metrics/PerceivedComplexity
-  # rubocop:disable Metrics/CyclomaticComplexity
+  WELSH_POLICY_START_DATE = DateTime.new(2019, 2, 4).utc.to_date
+  ENGLISH_POLICY_START_DATE = DateTime.new(2019, 10, 1).utc.to_date
+
   def self.calculate_pom_responsibility(offender)
```
rossjones: Is it worth making it clear here that we are using the POM responsibility to elsewhere determine the case responsibility (Custody/Community)?  Just as a means of clarifying that if the Community is responsible, then any pom will be have a SUPPORTING responsibility, if it's Custody then RESPONSIBLE.

starswan: Yes - any suggestions as to how? I thought you were doing something to tie the concepts directly together

rossjones: I think I was pondering whether this class should be returning the :custody/:community responsibility, then working out the pom responsibility from that.  At the moment I think we work out the pom's responsibility and then workout :custody/:community from that.

But maybe just a class-level comment?

starswan: The whole Prison/Probation thing (which is then converted to Custody/Community) is done by OffenderBase#case_owner. Need to have a think about refactoring...

spec/services/responsibility_service_spec.rb
starswan: There was - I've updated the comments to make this more explicit

PR #812, opened 2019-09-23
Allocations back links and redirects

theKHutDeveloper:
Currently when an user is on a paginated list and they click to view or edit an allocation, they always return to page 1 of the paginated list regardless of the page they were on originally.

This PR attempts to fix that issue by using the javascript.history.back method and by storing the page no of the paginated list in the params when redirecting back to the paginated list.

I also had to make changes to the overrides_controller to correctly identifying whether the event is an allocation or reallocation. Previously all overrides were being flagged as `allocate_primary_pom` events and would return the user to `Make allocations` tab even if the event was actually `reallocate_primary_pom` and should have been returned to `See allocations` tab.

Added the 'geckodriver-helper' gem to test the javascript in the tests.

### DIFF
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -12,7 +12,6 @@ require 'capybara/rspec'
 require 'webmock/rspec'
 require 'paper_trail/frameworks/rspec'
 
-Capybara.javascript_driver = :selenium
 Capybara.default_max_wait_time = 4
 Capybara.asset_host = 'http://localhost:3000'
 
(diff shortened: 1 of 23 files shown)


spec/rails_helper.rb
```diff
@@ -12,6 +12,10 @@
 require 'webmock/rspec'
 require 'paper_trail/frameworks/rspec'
 
+Capybara.javascript_driver = :selenium
```
starswan: Think it would be good to only use the new driver where we actually need it.

PR #814, opened 2019-09-23
Store the COM when importing from Delius data.

rossjones:
When we import the delius data and create/update a CaseInformation for
the offender, we should also attach the COM name to their current
allocation (if any).

This PR also changes the prisoner profile to show the community
information (when it is available).


db/schema.rb
```diff
@@ -133,6 +134,15 @@
     t.index ["nomis_staff_id"], name: "index_pom_details_on_nomis_staff_id"
   end
 
+  create_table "responsibilities", force: :cascade do |t|
```
starswan: looks like this got added by accident - no harm done in this case though.

rossjones: Good catch, will take it out pending POM290 :)

PR #816, opened 2019-09-25
Pom 133 manual parole review date

starswan:
This ticket just adds parole_review_date to case_information. No behavioural changes


app/controllers/case_information_controller.rb
```diff
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class CaseInformationController < PrisonsApplicationController
+  before_action :set_case_info, only: [:edit, :edit_prd, :show]
```
rossjones: and :update?

app/models/case_information.rb
```diff
@@ -27,4 +27,8 @@ class CaseInformation < ApplicationRecord
   # nil means MAPPA level is completely unknown.
   # 0 means MAPPA level is known to be not relevant for offender
   validates :mappa_level, inclusion: { in: [0, 1, 2, 3], allow_nil: true }
+
+  acts_as_gov_uk_date :parole_review_date
+
+  validates :parole_review_date, date: { after: proc { Date.yesterday }, allow_nil: true }
```
rossjones: What is Date.yesterday for?

db/schema.rb
```diff
@@ -134,6 +135,15 @@
     t.index ["nomis_staff_id"], name: "index_pom_details_on_nomis_staff_id"
   end
 
+  create_table "responsibilities", force: :cascade do |t|
```
rossjones: Accidentally included?

starswan: yes good spot

app/models/case_information.rb
starswan: to make sure that we don't enter a date in the past

rossjones: TIL

PR #818, opened 2019-09-25
Make handover date the responsibility of the sentence_detail

rossjones:
Currently the offender model is responsible for calculating the handover
date using local methods and the contents of the SentenceDetail object
stored in `@sentence`.

In cases where we don't have a whole offender (but do have the case
information for that offender) such as the caseload_controller it would
be better if the handover calculation could be done with just the case
information and sentence details.

This PR moves the responsibility for asking for handover dates to
SentenceDetail where it should be provided an instance of something that
responds to `nps_case?, early_allocation? mappa_level, and
indeterminate_sentence?).  The newly moved methods will use this object
along with the local dates to request the handover dates.

## Problem

To use this in the caseload it would be necessary to replace the indeterminate_sentence? requirement with earliest_release_date.nil? because we can only determine the imprisonment type by fetching the offender.

### DIFF
--- a/app/models/nomis/offender.rb
+++ b/app/models/nomis/offender.rb
@@ -25,18 +25,6 @@ module Nomis
       fields.each { |k, v| instance_variable_set("@#{k}", v) } if fields.present?
     end
 
-    def early_allocation?
-      false
-    end
-
-    def nps_case?
-      case_allocation == 'NPS'
-    end
-
-    def pom_responsibility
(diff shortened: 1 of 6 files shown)


rossjones:
We don't keep a reference to the CaseInformation model, so there won't be one hanging around when we need it.  Test fails fixed, and coverage dropped 0.04% so will see if I can find it.

rossjones:
Closing this as we're not convinced this is any better than the current solution.

PR #827, opened 2019-09-26
POM-336 Show upcoming handovers in caseload

rossjones:
On the POM caseload page, we now show the number of handovers due in the
next 7 days, clicking the link will take the POM to a new page where
they can view the offenders pending the start of handover.


app/controllers/caseload_controller.rb
```diff
@@ -3,19 +3,21 @@
 class CaseloadController < PrisonsApplicationController
   before_action :ensure_pom
 
-  breadcrumb -> { 'Your caseload' }, -> { prison_caseload_index_path(active_prison) }
+  breadcrumb -> { 'Your caseload' },
+             -> { prison_caseload_index_path(active_prison) }
   breadcrumb -> { 'New cases' },
              -> { new_prison_caseload_path(active_prison) }, only: [:new]
+  breadcrumb -> { 'Cases close to handover' },
```
starswan: These breadcrumbs look a little strange  - we seem to have a link to ourselves (even for new) which looks wrong. I thought the idea was to show links backwards, not including the page that we are actually on.

config/routes.rb
```diff
@@ -29,7 +29,9 @@
   resources :prisons do
     resources :prisons, only: :index
     resources :dashboard, only: :index
-    resources :caseload, only: %i[ index new ]
+    resources :caseload, only: %i[ index new]
+    get('/caseload/handover_start' => 'caseload#handover_start', as: 'caseload_handover_start')
```
starswan: I think you could have nested this route

spec/features/pom_caseload_spec.rb
```diff
@@ -49,6 +55,72 @@
     create(:allocation_version, nomis_offender_id: 'G4706UP', primary_pom_nomis_id: '485637', nomis_booking_id: '1180800')
     create(:allocation_version, nomis_offender_id: 'G9344UG', primary_pom_nomis_id: '485637', nomis_booking_id: '841994')
     create(:allocation_version, nomis_offender_id: 'G4273GI', primary_pom_nomis_id: '485637', nomis_booking_id: '1153753')
+
+    stub_request(:get, elite2listapi).
```
starswan: I think we need to start removing the duplication in these tests real soon now [tm]

app/controllers/caseload_controller.rb
rossjones: Think we need to investigate whether the gem let's us define our own views so that last is inactive.

PR #828, opened 2019-09-27
Pom 288 sort by arrival date

starswan:


app/models/nomis/offender_base.rb
```diff
@@ -66,18 +63,7 @@ def over_18?
...
-                          ResponsibilityService::ENGLISH_POLICY_START_DATE
-                        end
-
-      if sentence.sentence_start_date.nil? ||
-          sentence.sentence_start_date < omic_start_date
-        (Time.zone.today - omic_start_date).to_i
-      else
-        (Time.zone.today - sentence.sentence_start_date).to_i
-      end
+      (Time.zone.today - reception_date).to_i
```
rossjones: Reception date is only available in offender_summary, and this is only called/used for summary pages, so perhaps might be better in the subclass?

app/services/summary_service.rb
```diff
@@ -46,6 +46,7 @@ def self.summary(summary_type, prison, page, params)
     end
 
     if params.sort_field.present?
+      bucket.items.each { |offender| add_reception_date(offender) } unless summary_type == :allocated
```
rossjones: If this is sorting before only taking a page worth of offenders for the view, it might well trigger hundreds of api calls via add_reception_date :(

starswan: agreed 100%

starswan: yes - that's why the call it is making is cached. It takes around 30 seconds uncached, but is almost instantaneous when cached.

rossjones:
Not sure I've had enough coffee to decode the test fail, but this looks good to me otherwise.

PR #849, opened 2019-10-02
Changes how caseload/pom_details gets allocated offenders

rossjones:
Because the way we were previously working out the responsibilities for
allocated offenders was incorrect (we were using fake offender
instances) this PR changes to filter the offenders in a prison to only
show those that are allocated to a POM.

### DIFF
--- a/app/services/prison_offender_manager_service.rb
+++ b/app/services/prison_offender_manager_service.rb
@@ -65,7 +65,7 @@ class PrisonOffenderManagerService
 
     # Get all of the offenders at this prison and select only those that
     # appear in an allocation.  Then for each one apply a map to determine
-    # the responsibility and pair it with the allocation and sentece.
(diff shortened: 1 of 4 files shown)


app/services/prison_offender_manager_service.rb
```diff
@@ -55,45 +55,40 @@ def self.get_user_name(username)
...
-      if record.present?
-        offender_stub.tier = record.tier
-        offender_stub.case_allocation = record.case_allocation
-        offender_stub.welsh_offender = record.welsh_offender
-      end
 
-      if alloc.for_primary_only?
+    # Get all of the offenders at this prison and select only those that
+    # appear in an allocation.  Then for each one apply a map to determine
+    # the responsibility and pair it with the allocation and sentece.
```
KPobeeNorris:
```suggestion
    # the responsibility and pair it with the allocation and sentence.
```

PR #855, opened 2019-10-04
POM-379 - Responsibility override fix

KPobeeNorris:
We recently saw a Sentry alert
(https://sentry.service.dsd.io/mojds/live1-allocation-manager-produ/issues/37899/?referrer=slack)
which had the error 'undefined method `emails' for nil:NilClass'.
Following an investigation into the error it was found that when
creating a responsibility override we're calling a method to get the signed in pom details,
which causes a couple of issues - firstly the way that method works it
always returns nil for the email address as we only get the email address
when we fetch one POM rather than all POMs, but more importantly we’re
not looking for a POM's email address, we’re looking for the SPO’s email
address.  So, unless they happen to have a POM role too then it won’t
find them using this method.  Therefore, we need to call
'Nomis::Elite2::UserApi.user_details(current_user).email_address.first'
as it will definitely get the current user’s details, which includes their email address


app/controllers/responsibilities_controller.rb
```diff
@@ -21,7 +21,7 @@ def confirm
   def create
     @responsibility = Responsibility.create! responsibility_params
 
-    me = PrisonOffenderManagerService.get_signed_in_pom_details(current_user, @prison).emails.try(:first)
+    me = Nomis::Elite2::UserApi.user_details(current_user).email_address.first
```
rossjones: Still worth wrapping the email call with try() in case it’s nil

KPobeeNorris: Good idea! Done and done

PR #858, opened 2019-10-07
Pom 45 Early Allocation Part I

starswan:


rossjones:
Little bit of weird behaviour...

* If I make a mistake in the date (by adding letters) I don't seem to be able to select any radios on the newly rendered error page. Submitting the page with more errors does seem to have picked up the values though as they are rendered on the second view of the error page.
* After completing the process and viewing the PDF, it opens in the same window, going back throws an error about unknown format.

Might be best to deliver the PDF as a content-disposition: attachment? Or in a new tab?

app/controllers/early_allocations_controller.rb
```diff
@@ -0,0 +1,69 @@
...
+    end
+  end
+
+  def show
+    @offender = EarlyAllocation.find_by! offender_id_from_url
+    @prisoner = OffenderService.get_offender(params[:offender_id])
+    @allocation = AllocationVersion.find_by!(offender_id_from_url)
+    @pom = PrisonOffenderManagerService.get_pom(@prison, @allocation.primary_pom_nomis_id)
+
+    respond_to do |format|
```
rossjones: If this is still pending (I think there's another ticket to be able to mark it as complete), should we show stage2 for the html view?

config/routes.rb
```diff
@@ -69,6 +69,11 @@
     get('/summary/allocated' => 'summary#allocated')
     get('/summary/unallocated' => 'summary#unallocated')
     get('/summary/pending' => 'summary#pending')
+
+    resources :offenders, only: [] do
```
rossjones: Do we definitely need another version of /prisoners?

starswan: no - we didn't have a resource for it. We do now :-)

rossjones: Probably, yes.

app/views/early_allocations/new.html.erb
```diff
@@ -0,0 +1,59 @@
+<% content_for :switcher do %>
+  <%= render '/layouts/prison_switcher' %>
+<% end %>
+
+<% if @offender.errors.count > 0 %>
+  <%= render :partial => "/shared/validation_errors", :locals => { :errors => @offender.errors } %>
+<% end %>
+
+<h1>Assessment for early allocation to the community probation team</h1>
```
rossjones: h1's need a special class (otherwise it renders with Times as a font). There are a list of options at https://design-system.service.gov.uk/styles/typography/ ..

```
* class="govuk-heading-xl"
* class="govuk-heading-l"
* class="govuk-heading-m"
* class="govuk-heading-s">govuk-heading-s
```

rossjones:
When I tried this I got ```
undefined method `reason' for #<EarlyAllocation:0x00007fbf9ebe16e8>
<%= form.text_area(:reason, rows: 3, class: 'govuk-textarea') %>
```

I think it's now called :other_reason?

File: app/models/early_allocation.rb [hunk 4]
```diff
@@ -0,0 +1,83 @@
...
+  end
+
+  # This field is only prompted for if extremism_separation is true
+  validates(:due_for_release_in_less_than_24months, inclusion: {
+              in: [true, false],
+              allow_nil: false }, if: -> { extremism_separation })
+
+  attribute :stage2_complete, :boolean
+
+  validates :reason, presence: true, if: -> { stage2_complete }
```
rossjones: :other_reason?

app/controllers/early_allocations_controller.rb
```diff
@@ -0,0 +1,79 @@
...
+    params.require(:early_allocation).
+      permit(EarlyAllocation::STAGE1_BOOLEAN_FIELDS +
+                EarlyAllocation::ALL_STAGE2_FIELDS +
+                [:oasys_risk_assessment_date_dd,
+                 :oasys_risk_assessment_date_mm,
+                 :oasys_risk_assessment_date_yyyy,
+                 :oasys_risk_assessment_date,
+                 :stage1_complete,
+                 :stage2_complete,
+                 :reason,
```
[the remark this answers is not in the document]
starswan: no - other_reason is the boolean, reason is the actual text. Not sure why your test didn't work - maybe you had run the migration before? It seems that you need to reset your test db when you change a migration.

config/application.rb
```diff
@@ -51,5 +51,12 @@ class Application < Rails::Application
       [config.zendesk_username, config.zendesk_url, config.zendesk_password].all?
 
     config.cache_expiry = 60.minutes
+
+    # config.action_view.field_error_proc = Proc.new { |html_tag, instance|
```
KPobeeNorris: @starswan - are these commented out lines needed, or can they be deleted?

PR #859, opened 2019-10-07
Add LDU information to seeds

KPobeeNorris:
We have implemented the responsibility override feature that allows
SPOs to change the responsibility from custody to community.  However,
so that the team can use this feature we need to add some LDU
information into the seeds file; this PR does that.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -45,6 +45,12 @@ LocalDivisionalUnit.create!(
     email_address: "EnglishNPS@example.com"
 )
 
+LocalDivisionalUnit.create!(
+    code: "OTHERLDU",
+    name: "English LDU 2",
+    email_address: nil
+)
+
 # The offenders below are those with release dates a few years in the future and can therefore use the
 # responsibility override workflow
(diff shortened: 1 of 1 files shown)


rossjones:
Might want to add an LDU with no email address as well?

KPobeeNorris:
New LDU without email address added

db/seeds.rb
```diff
@@ -32,3 +32,64 @@
...
+# The offenders below are those with release dates a few years in the future and can therefore use the
+# responsibility override workflow
+
+CaseInformation.find_or_create_by!(
+    nomis_offender_id: 'G7658UL',
+    tier: 'A',
+    case_allocation: 'NPS',
+    welsh_offender: 'Yes',
+    manual_entry: true,
+    local_divisional_unit_id: 1,
```
rossjones: Rather than assume id of 1, might it be best to use the result of find_or_create_by! above to use that id?

KPobeeNorris: Implemented

PR #866, opened 2019-10-07
Allocation API Authentication

theKHutDeveloper:
Requests to the MOIC API must be authenticated requests to ensure the data behind the API is only available to services with a specific role who have a valid token.

Services accessing the API must supply a bearer token that qualifies the following:

* decrypted with the SSO public key
* not be an expired token
* have a read scope


app/controllers/api_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+  end
+
+  def valid_token?(access_token)
+    return false if Nomis::Oauth::Token.new(access_token: access_token).expired?
+
+    if Nomis::Oauth::Token.new(access_token: access_token).valid_token?
+      true
+    else
+      false
+    end
```
rossjones: So I don't think you need this block.  The if statement returns a boolean, and then you're going to return the same boolean.  You should be able to replace it all with

```suggestion
Nomis::Oauth::Token.new(access_token: access_token).valid_token?
```
Having said that, I'd check the valid_token? first, and just put the expired check inside valid_token?

PR #873, opened 2019-10-09
POM - 380: Fetching images of offenders sometimes fail

theKHutDeveloper:
When we collect the sentence details, we fetch the facial image id for the offender. If there is no image for the offender we get a status 500 error.

This commit uses a default image if no image exists for the offender thus avoiding status 500 errors.


app/services/nomis/elite2/offender_api.rb
```diff
@@ -98,11 +98,15 @@ def self.get_bulk_sentence_details(booking_ids)
       end
 
       def self.get_image(booking_id)
+        # This method returns the raw bytes of an image, the equivalent of loading the
+        # image from file on disk.
         details_route = '/elite2api/api/offender-sentences/bookings'
         details = e2_client.post(details_route, [booking_id])
 
         image_route = "/elite2api/api/images/#{details.first['facialImageId']}/data"
```
rossjones: This might fail as well if 'facialImageId' is not in the `details` hash, so might be worth checking that too

rossjones:
Looking good.  I think we need another test, where after calling `/elite2api/api/offender-sentences/bookings` it returns a hash that is missing a facialImageId - you should be able to stub this like the others.

PR #874, opened 2019-10-09
Pom 30 send emails Early Allocation Part III

starswan:


app/controllers/early_allocations_controller.rb
```diff
@@ -0,0 +1,99 @@
...
+  def load_prisoner
+    @offender = OffenderService.get_offender(params[:prisoner_id])
+    @allocation = AllocationVersion.find_by!(offender_id_from_url)
+    @pom = PrisonOffenderManagerService.get_pom(@prison, @allocation.primary_pom_nomis_id)
+  end
+
+  def save_model_and_email
+    @early_assignment.save!
+
+    PomMailer.send_early_allocation_email(email: @offender.ldu.email_address,
```
rossjones: Do we have a clear steer on what is expected if there is no LDU email address, or do we not even get this far?

app/controllers/early_allocations_controller.rb
```diff
@@ -0,0 +1,99 @@
...
+                                          prisoner_name: @offender.full_name,
+                                          prisoner_number: @offender.offender_no,
+                                          pom_name: @allocation.primary_pom_name,
+                                          pom_email: 'x',
```
rossjones: 'x'

app/controllers/early_allocations_controller.rb
starswan: Yes the story now includes a dead-end page if we don't have an LDU email address

rossjones:
Have told codeclimate to be quiet, just down 0.07% coverage

PR #876, opened 2019-10-10
POM-369 Implement an endpoint for retrieving allocations

rossjones:
This API adds a single endpoint (at `/api/allocation/{offender_no})`
which will retrieve the POM details for any allocation against that
offender.

To make the request a current, valid, authorisation token is required
signed by the HMPPS SSO. Unauthorized access will generate a 401
response, a missing allocation or a missing offender will return a 404.

Swagger documentation (generated with `rake rswag:specs:swaggerize`) is
available to end-users are /api-docs. Future changes to the API (and
therfore the rswag specific tests in spec/api/) will need to manually
update the swagger documentation should more endpoints be added.


spec/api/allocation_api_spec.rb
```diff
@@ -0,0 +1,94 @@
+# frozen_string_literal: true
+
+require 'swagger_helper'
+
+# rubocop:disable RSpec/DescribeClass
```
starswan: would be nice to document why you had to do this

starswan: would be nice to document why you had to disable this in rubocop

PR #883, opened 2019-10-11
Try to introduce VCR=0 as a faster way of running tests

starswan:
This PR tries to introduce VCR=0 rspec as a 2x faster way of running tests - using your existing VCR recordings from an earlier run. 

It seems to be slightly flawed, but possibly due to the way that we authenticate to NOMIS - the tests seem have a habit of randomly authenticating during a different exchange to that which was recorded (due to rspec randomising test execution order). 

Suggestions welcomed...

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -31,17 +31,18 @@ end
 
 require 'vcr'
 
-recording = ENV['VCR']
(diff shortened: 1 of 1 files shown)


starswan:
Think there is a simpler solution

PR #891, opened 2019-10-16
Refactor early allocation

starswan:


rossjones:
When I get the date wrong at stage 1, I get the following screen ...
Manage_Offenders_in_Custody-5
When I correct the errors and progress to stage 2, I see the following before I do anything...
Manage_Offenders_in_Custody-6

starswan:
Thanks @rossjones I've improved the test and fixed a couple of issues - one with the error template itself

rossjones:
Looks like that fixed it.

PR #893, opened 2019-10-16
Change open prison responsibility and notify LDU

rossjones:
NPS Offenders who are in open prisons should be Supporting/Community. As
the offender moves into the open prison, their LDU should be notified
that they have to take responsibility for that offender - should the
offender have an LDU with a valid email address.  This is done during
the processing of transfers to check if the target prison is open or
not. If it is, then an email is sent to the LDU with some offender
details.

As a drive-by, this PR also ups the max CyclomaticComplexity to 10 so we
don't get punished just for using if statements.


spec/jobs/open_prison_transfer_job_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+  end
+
+  it 'does not send an email when no LDU email address' do
+    allow(OffenderService).to receive(:get_offender).and_return(Nomis::Offender.new.tap{ |o|
+      o.prison_id = open_prison_code
+      o.offender_no =  nomis_offender_id
+      o.case_allocation = 'NPS'
+    })
+
+    stub_request(:get, "#{elite2api}/staff/#{nomis_staff_id}/emails").
```
starswan: There's a 'stub_poms' method in ApiHelper - IMHO we should be adding to that rather than doing 'naked' API stubbing

PR #895, opened 2019-10-16
Fix error where it was possible to edit PRD too early

rossjones:
Currently the link to edit the offender's PRD is available on the
case_information#show page - but unfortunately there is no case
information to edit at that point.

This PR adds a local to wherever the partial is rendered specifying
whether it should be possible to edit the PRD from that page (and if so,
showing the link).

Parole review date was also duplicated in case information and so has
been removed and left in offence_info.


app/views/case_information/show.html.erb
```diff
@@ -65,7 +65,7 @@
       </tr>
     </tbody>
   </table>
-  <%= render 'shared/offence_info' %>
+  <%= render partial: 'shared/offence_info', locals: {editable_prd: false} %>
```
starswan: I thought the bug was to do with editing a PRD when there was no case_information? This is in the 'case information show' action, so surely it's ok here?

rossjones: This is used for when we are showing why we failed to import from DeliusData, so no case info available yet.

PR #901, opened 2019-10-17
Allow get_multiple_offenders to return more than 10 offenders

rossjones:
This call previously had page limits with a default of 10, we now set
the page limit to the number of IDs passed.  This means the caller will
have to handle paging if they are requesting an unreasonable number of
offenders...


rossjones:
Just realised this is incomplete so might be worth holding off merging this.

PR #906, opened 2019-10-21
POM - 418: redirection fails when caseload is nil

theKHutDeveloper:
This bug occurs when a POM's Nomis caseload has not been set up properly, resulting in their caseload being nil.

Currently when we check the POM's prison access, the app crashes because we are not expecting the POM's caseload to be nil.

This commit fixes this by redirecting to /401 if the POM's caseload is nil.


rossjones:
Is this something we can test or is that problematic?

rossjones:
@theKHutDeveloper the best way to test this is to add a controller test in spec/controllers/* - if you check out allocations_controller_spec.rb you'll see that it has `stub_sso_data(prison)` in a before block, you should be able to do the same and pass nil so that it stubs a token with no prison.

PR #910, opened 2019-10-22
POM-432 Multiple POM records bug

KPobeeNorris:
We had been contacted by a number of SPOs advising that they were unable
to update their POM working patterns.  An investigation led to finding
17 POMs who have two PomDetail records, which meant that one record was
being updated, but the other being displayed with the incorrect data.
We are adding a uniqueness constraint to the nomis staff ID field in
both the db and model to prevent this happening again.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -153,7 +153,7 @@ ActiveRecord::Schema.define(version: 2019_10_10_090002) do
     t.string "status"
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
-    t.index ["nomis_staff_id"], name: "index_pom_details_on_nomis_staff_id"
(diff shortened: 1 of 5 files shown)


db/schema.rb
```diff
@@ -153,7 +153,7 @@
     t.string "status"
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
-    t.index ["nomis_staff_id"], name: "index_pom_details_on_nomis_staff_id"
+    t.index ["nomis_staff_id"], name: "index_pom_details_on_nomis_staff_id", unique: true
```
rossjones: Does the table already have this unique constraint or do we need to add it in a migration?

KPobeeNorris: Need to add it in a migration!

PR #911, opened 2019-10-22
Pom 430 create prison model

starswan:


app/models/prison.rb
```diff
@@ -0,0 +1,145 @@
...
+
+  def unfiltered_offenders
+    # Returns all offenders at the provided prison, and does not
+    # filter out under 18s or non-sentenced offenders in the same way
+    # that get_offenders_for_prison does.
+    OffenderEnumerator.new(@code)
+  end
+
+  # Note - get_poms and get_pom return different data...
+  def poms
```
rossjones: Should we memoize this to avoid all the redis round trips?

app/services/allocation_service.rb
```diff
@@ -153,7 +153,7 @@ def self.current_pom_for(nomis_offender_id, prison_id)
     current_allocation = active_allocations(nomis_offender_id, prison_id)
     nomis_staff_id = current_allocation[nomis_offender_id]['primary_pom_nomis_id']
 
-    PrisonOffenderManagerService.get_pom(prison_id, nomis_staff_id)
+    Prison.new(prison_id).get_pom(nomis_staff_id)
```
rossjones: Could the whole surrounding method move into Prison so we can avoid constructing another one here? Or pass in the prison instead of the prison_id?

app/services/email_service.rb
```diff
@@ -10,8 +10,7 @@ def initialize(message:, allocation:, pom_nomis_id:)
     @allocation = allocation
 
     @offender = OffenderService.get_offender(@allocation[:nomis_offender_id])
-    @pom = PrisonOffenderManagerService.get_pom(
-      @allocation.prison,
+    @pom = Prison.new(@allocation.prison).get_pom(
```
rossjones: Can we pass in the prison here?

app/services/offender_service.rb
```diff
@@ -93,36 +36,15 @@ def self.get_multiple_offenders(offender_ids)
...
   def self.get_sentence_details(booking_ids)
     Nomis::Elite2::OffenderApi.get_bulk_sentence_details(booking_ids)
   end
 
   # Takes a list of OffenderSummary or Offender objects, and returns them with their
   # allocated POM name set in :allocated_pom_name.
   # This is now only used by the SearchController.
-  def self.set_allocated_pom_name(offenders, caseload)
-    pom_names = PrisonOffenderManagerService.get_pom_names(caseload)
+  def self.set_allocated_pom_name(offenders, prison_id)
```
rossjones: Suspect this should be part of AllocationWithSentence. For later.... But can we pass in the prison rather than the id again?

app/services/allocation_service.rb
starswan: This is a hint that we have the design wrong - yes we shouldn't be arbitrarily created model objects in this layer

app/services/email_service.rb
starswan: again, this is a design issue. It's just more obvious now we have a data layer

app/services/offender_service.rb
rossjones: I've refactored AllocationWithSentence elsewhere, so this has somewhere to live now, so we can ignore this comment.

app/services/offender_service.rb
```diff
@@ -93,36 +36,15 @@ def self.get_multiple_offenders(offender_ids)
     }
   end
 
-  def self.get_multiple_offenders_as_hash(offender_ids)
```
rossjones: Is there a reason this is removed? It's not related to a Prison, assume it's because it doesn't do much?

spec/spec_helper.rb
```diff
@@ -7,9 +7,9 @@
   add_group "Services", "app/services"
 
   # Try to set this to current coverage levels so that it never goes down after a PR
-  # 26 lines uncovered at 99.03% coverage
-  minimum_coverage 99.02
-  maximum_coverage_drop 0
+  # 23 lines uncovered at 99.16% coverage
```
rossjones: 99.16 doesn't match the 99.15 below.

app/services/offender_service.rb
starswan: it was completely unused

PR #915, opened 2019-10-23
Ensure we close the IMAP connection before processing data

rossjones:
We're seeing the IMAP connection fail in production, but this was
because we were running the data processing steps inside the closure
(that handles fetching the attachment).

This PR now only processes the attachment _after_ the IMAP connection is
closed.

Also reduces logging during tests.

### DIFF
--- a/app/jobs/delius_import_job.rb
+++ b/app/jobs/delius_import_job.rb
@@ -19,10 +19,6 @@ class DeliusImportJob < ApplicationJob
   def perform
     ActiveRecord::Base.connection.disable_query_cache!
 
-    unless Rails.env.test?
-      Rails.logger = Logger.new(STDOUT)
-    end
-
     username = ENV['DELIUS_EMAIL_USERNAME']
     password = ENV['DELIUS_EMAIL_PASSWORD']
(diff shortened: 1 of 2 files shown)


rossjones:
Yes, you're mostly correct.   process_attachment() is provided the base64 decoded content from the message (a 9+Mb parameter!) and doesn't actually download it.  

Will decode it and assign  before passing to process_attachment (and add a comment).

rossjones:
I've changed this to work with a byte array now which I think is probably more approprite for the contents of the attachment.

rossjones:
@KPobeeNorris @starswan is it possible to get a re-review, otherwise we might miss tonight's run

app/jobs/delius_import_job.rb
```diff
@@ -15,45 +15,66 @@ class DeliusImportJob < ApplicationJob
       :mappa, :mappa_levels, :date_of_birth
     ].freeze
 
+  # rubocop:disable Metrics/MethodLength
   def perform
     ActiveRecord::Base.connection.disable_query_cache!
-    Rails.logger = Logger.new(STDOUT)
+
+    unless Rails.env.test?
```
starswan: don't think we need to do this...

PR #916, opened 2019-10-23
Pom 438 early allocation rework

starswan:


rossjones:
It would be really useful (to me at least) if there was an easy way to determine if the offender has any (active) early allocation assessments that are pending a community response. Any chance?

app/helpers/early_assignment_pdf_helper.rb
```diff
@@ -68,9 +68,9 @@ def add_assessment_info(pdf, early_assignment)
       end
     end
 
-    assessment_info = info_hash.map { |k, v| [DESCRIPTIONS[k], v] }
+    # assessment_info = info_hash.map { |k, v| [DESCRIPTIONS[k], v.nil? ? 'N/A' : v] }
```
rossjones: Looks unnecessary?

config/routes.rb
```diff
@@ -17,12 +17,14 @@
...
 
-      # show is only used with format PDF
-      resource :early_allocation, only: [:new, :create, :show, :edit, :update] do
+      resource :early_allocations, only: [:new, :create] do
         post('discretionary')
-        get('community_decision')
-        put('community_decision' => 'early_allocations#record_community_decision')
       end
+
+      # show most recent assessment with format PDF
```
rossjones: Comments seem a bit wobbly

PR #939, opened 2019-11-01
[SPIKE + WIP] Provide faceting and filtering of a POM's caseload

rossjones:
An upcoming ticket focused on redesign of various screens has a
component that describes filters for a POM's caseload.  As part of that
component it shows a facet of the total number of allocations (post
application of filters if any).

This PR implements (currently) just the calculation of the facet counts
for each PomCaseload.


rossjones:
Going to close this, a very long way from being needed I think, and will not delete branch so it's still there if necessary.

PR #943, opened 2019-11-04
POM-444 de-allocate both primary and co-working POMs

starswan:
for this prison - when POM marked inactive

### DIFF
--- a/app/models/allocation_version.rb
+++ b/app/models/allocation_version.rb
@@ -39,11 +39,6 @@ class AllocationVersion < ApplicationRecord
   scope :allocations, lambda { |nomis_offender_ids|
     where(nomis_offender_id: nomis_offender_ids)
   }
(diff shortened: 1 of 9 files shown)


app/models/allocation_version.rb
```diff
@@ -112,8 +107,8 @@ def self.deallocate_offender(nomis_offender_id, movement_type)
     alloc.save!
   end
 
-  def self.deallocate_primary_pom(nomis_staff_id)
-    all_primary_pom_allocations(nomis_staff_id).each do |alloc|
+  def self.deallocate_primary_pom(nomis_staff_id, prison)
+    active_pom_allocations(nomis_staff_id, prison).each do |alloc|
```
rossjones: Do we also need to do this for cases where they are the secondary pom?

rossjones: Answering my own question: Keep reading down ...

PR #945, opened 2019-11-05
Show correct responsibility when it has been overridden

rossjones:
When an offender has their responsibility overridden to be PROBATION,
the currently allocated POM should show as SUPPORTING, and not
RESPONSIBLE.  Allocated secondary POM will become CO-WORKING.

Currently this was not handled by the POMCaseload or the previous
PrisonOffenderManagerService function.

This PR now checks for an overriddent responsibility when determining a
POMs caseload.

### DIFF
--- a/app/models/pom_caseload.rb
+++ b/app/models/pom_caseload.rb
@@ -23,7 +23,6 @@ class PomCaseload
 
 private
 
-  # rubocop:disable Metrics/MethodLength
   def load_allocations
     allocation_list = AllocationVersion.active_pom_allocations(
       @staff_id, @prison_id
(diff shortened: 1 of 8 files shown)


app/models/pom_caseload.rb
```diff
@@ -32,25 +32,40 @@ def load_allocations
     offender_ids = allocation_list.map(&:nomis_offender_id)
     offenders = OffenderService.get_multiple_offenders(offender_ids)
 
+    # Lookup responsibility overrides and store them in a hash for this
```
starswan: eek - don't think we should be jumping through this many hoops just to do something really simple. A responsibility should be attached somewhere in the rails model (sorry probably me forgot to do it) and we should just go and get it. Then use the presenter to display it if it's there, rather than writing all this code. We need to push towards a DRY and simple approach.

rossjones: 👍 
Was thinking about moving it to the offender#pom_responsibility method, but I guess the presenter is a good place.

PR #946, opened 2019-11-06
Add extra metrics (sidekiq) and enable custom items

rossjones:
For discussion

Currently we use the prometheus exporter gem which means we have another
process in another container collecting metrics and making them
available at a /metrics endpoint.  This however is not fun to work with.

This PR change how we collect metrics to use the Prometheus provided
Ruby client, which is a lot simpler and exposes a /metrics endpoint
within our application.

The MetricService (I know, I know) shows how we can create counters and
gauges and the SearchController shows how we would record them.

Metrics from sidekiq is a more complicate issue and so would need to
provide its own /metrics endpoint.  Trials with prometheus_sidekiq (from
fastly) were problematic because of a version conflict with the client.


rossjones:
One problem that I've encountered, is metrics that might be triggered by the rails console, and this causes a problem with prometheus-client in that the metrics will already be registered in the singleton instance that allows for them to be written (this isn't the same singleton running in the web app) and it will fail trying to re-register the metrics.

Gemfile.lock
```diff
@@ -98,14 +97,14 @@ GEM
     ethon (0.12.0)
       ffi (>= 1.3.0)
     execjs (2.7.0)
-    factory_bot (5.1.0)
+    factory_bot (5.1.1)
```
starswan: there are some slightly odd changes in here not to do with adding metrics..?

rossjones:
Closed in favour of https://github.com/ministryofjustice/offender-management-allocation-manager/pull/962 which is a smaller, less impactful change.

PR #948, opened 2019-11-06
Access to various controllers was not correctly enforced

rossjones:
POMs should not see SPO task links on the dashboard, and we should
ensure the user has the correct role to be able to access those
controllers.

Also changes search so that unless the user is an SPO, they will be
redirected to caseload search instead of the actionable search (which
allows allocating etc).


spec/controllers/dashboard_controller_spec.rb
```diff
@@ -0,0 +1,117 @@
...
+        expect(response.body).to have_content('Make new allocations')
+        expect(response.body).to have_content('Update case information')
+        expect(response.body).to have_content('View all offender managers')
+      end
+    end
+  end
+end
+# require 'rails_helper'
+
+# feature 'get dashboard' do
```
starswan: commented-out code - remove or revive?

PR #949, opened 2019-11-07
Add useful logging to movements

rossjones:
Currently we're not logging very much (and not very searchably given our
logging woes) and so this PR adds more logging and prefixes it with
[MOVEMENT] so we can search it in Kibana.


rossjones:
It's more removing a mistakenly implemented feature I think.

rossjones:
Have removed the sneaky feature into its own PR.

PR #950, opened 2019-11-07
Do not let invalid allocations appear in a POM's caseload.

rossjones:
Currently the caseload for a POM gets all of their allocations and the
matching offenders. This currently includes invalid allocations, where
the offender is no longer convicted, or no longer at the same prison as
the POM/Allocation.

To avoid this we are now excluding allocated offenders from the caseload
where:

  * The offender is reported at a different prison/out
  * The offender is not convicted

We _do not_ de-allocate the offender (although we _could_ as long it was
constrained to any allocations at active prison), but we do log the
error with a [ALLOC] tag so that it is reasonably easy to locate them in
logs.

Currently this does not solve the problem, as the allocation still
exists (although if they are allocated at another prison then the
incorrect allocation will be corrected), so we still need to decide what
to do about those incorrect allocations as they may result in incorrect
numbers.

DISCUSS:  When we find an invalid allocation, we should do something
like:

```
DeallocateOffenderAtPrisonJob(prison, offender).perform_later
```

Worth noting that doing this based on absence from the current prison MIGHT be that they're on a hospital visit, or short-term-transfer. We could however do this if the offender is explicitly OUT.


rossjones:
The two failing tests on circleci run fine locally for me. Not quite sure what is going on.

app/models/allocated_offender.rb
```diff
@@ -33,6 +33,21 @@ def new_case?
     }.any?
   end
 
+  def valid?
+    unless @offender.convicted?
+      Rails.logger.warn("[ALLOC] #{@offender.offender_no} has an allocation but is not convicted")
```
starswan: There is no point logging warnings in our current infrastructure as they don't show up anywhere useful. We need to work out how we should raise such issues

app/models/pom_caseload.rb
```diff
@@ -37,11 +35,13 @@ def load_allocations
...
-      AllocatedOffender.new(
+      allocated_offender = AllocatedOffender.new(
         @staff_id,
         allocation,
         offender
       )
-    }
+
+      allocated_offender if allocated_offender.valid?
+    }.compact
```
starswan: This could be better implemented as <collection>.select(&:valid?)

spec/models/pom_caseload_spec.rb
```diff
@@ -17,15 +17,28 @@
   before do
     stub_poms(prison, pom)
 
+    # We cannot use stub_sso_data here because we do not have a session for it to fill
+    allow(Nomis::Oauth::TokenService).to receive(:valid_token).and_return(OpenStruct.new(access_token: 'token'))
```
starswan: the existence of this looks like a code smell. This object should be given its data rather than going to get it. If it were moved to being a presenter it would be more obvious that this was poor design decision

app/models/allocated_offender.rb
rossjones: They turn up in Kibana if we log to STDOUT.

spec/models/pom_caseload_spec.rb
rossjones: It looks like `stub_sso_pom_data` et al only expect to be called in feature/controller tests.  If we're stubbing API requests, the API code will want to be able to get a valid token.

I guess it means we should stub `AllocationVersion.active_pom_allocations` and `OffenderService.get_multiple_offenders` directly rather than the API calls in this case.

rossjones: Fixed the tests to not use stub_* and therefore don't need to be able to generate a token (as we're not testing the APIs).

app/models/allocated_offender.rb
starswan: but they will probably just get ignored - unless we have a process to look for warnings and errors in the Kibana logs

app/models/pom_caseload.rb
```diff
@@ -42,6 +40,6 @@ def load_allocations
         allocation,
         offender
       )
-    }
+    }.select(&:valid?)
```
starswan: Not sure I understand how an offender can be valid or not?

app/models/allocated_offender.rb
rossjones: That seems like too big a task for this PR.  If we want to know how many did or did not fail, then that's a job for metrics (PR incoming), but if we want to investigate detail, that's a logging thing.

app/models/pom_caseload.rb
starswan: Really unsure about this filtering. The warnings from the presenter will just end up in a log which we never look at. Also 'presenters' are just supposed to present, sorting and filtering is normally done in the controller.

PR #952, opened 2019-11-08
Pom 135 recall responsibilties

starswan:
Don't display handover dates for recalled offenders


KPobeeNorris:
This PR is no longer relevant as a larger piece of work has been done around calculations - see PR POM-574

PR #958, opened 2019-11-12
Pom 480 pom position AO or PPO

starswan:
The get_poms API is quite exactly what it seems - remove the potential duplicates when calling API

### DIFF
--- a/app/services/prison_offender_manager_service.rb
+++ b/app/services/prison_offender_manager_service.rb
@@ -6,13 +6,11 @@ class PrisonOffenderManagerService
     poms = Nomis::Elite2::PrisonOffenderManagerApi.list(prison)
     pom_details = PomDetail.where(nomis_staff_id: poms.map(&:staff_id).map(&:to_i))
 
-    poms = poms.map { |pom|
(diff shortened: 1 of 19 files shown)


app/services/prison_offender_manager_service.rb
```diff
@@ -3,16 +3,16 @@
...
 
-    poms = poms.map { |pom|
+    poms.map { |pom|
       detail = get_pom_detail(pom_details, pom.staff_id.to_i)
       pom.add_detail(detail, prison)
       pom
-    }.compact
-
-    poms
+    }.select { |pom| pom.prison_officer? || pom.probation_officer? }.uniq(&:staff_id)
```
rossjones: This might end up with people disappearing from the list of POMs where they might mistakenly have turned up.  Might be worth warning those on support of what the likely issue is if this occurs.

starswan: Don't think so. We were crashing if the position came back as anything other that 'PO' or 'PRO' and also filtering later - so a POM was potentially there but not in either list when it came to allocation

PR #961, opened 2019-11-13
POM - 395: Error messages should conform to GDS guidelines

theKHutDeveloper:
Our error messages currently do not conform to GDS guidelines. This PR ensures that we are following the error message standards as set by GDS.

Screenshots:

Overrides
Screen Shot 2019-11-19 at 14 59 30

Contact Us
Screen Shot 2019-11-19 at 14 54 17

Case Information
Screen Shot 2019-11-19 at 14 57 08

Responsibility Override
Screen Shot 2019-11-19 at 14 57 59

Screen Shot 2019-11-19 at 14 58 18


app/controllers/pages_controller.rb
```diff
@@ -52,6 +52,13 @@ def redirect_path
   end
 
   def help_params
-    params.permit(:message, :email_address, :name, :prison, :job_type)
+    # Unpermitted params when running development and test env
```
rossjones: Not sure we should do this, do we know why the extra params make an appearance in test?

app/views/case_information/_errors.html.erb
```diff
@@ -0,0 +1,20 @@
...
+        <li><a href="#case_information_welsh_offender_Yes"><%= errors[:welsh_offender].first %></a></li>
+      <% end %>
+
+      <% if errors[:case_allocation] %>
+        <li><a href="#case_information_case_allocation_NPS"><%= errors[:case_allocation].first %></a></li>
+      <% end %>
+
+      <% if errors[:tier] %>
+        <li><a href="#case_information_tier_A"><%= errors[:tier].first %></a></li>
+      <% end %>
```
rossjones: Rather than manually do this for each form, it might be better to make use of https://github.com/ministryofjustice/offender-management-allocation-manager/blob/master/app/views/shared/_field_errors.html.erb which should handle the errors for a form, and provide links etc

app/controllers/pages_controller.rb
theKHutDeveloper: In config/environments/development.rb we has this line config.action_controller.action_on_unpermitted_parameters = :raise, which is why it is a problem in dev and test. As soon as I click on the submit button I got the unpermitted parameters issue. This was the quickest way I was able to progress. I will try to find another solution.

app/views/case_information/_errors.html.erb
theKHutDeveloper: I never saw this 🤕 I could have saved so much time. I'll make the changes

app/controllers/pages_controller.rb
starswan: Its there deliberately so that we don't accidentally submit invalid parameters (by silently ignoring them, which is the production setting). It's set that way in production so if someone tried to attack your server they don't end filling your error logs with random rubbish, however its useful to have it in dev/test to show you that you're doing something wrong.

app/views/case_information/_errors.html.erb
rossjones: Is it safe to remove this file from the PR now?

app/views/case_information/_form.html.erb
```diff
@@ -21,7 +21,7 @@
 
         <div class="govuk-radios">
           <div class="govuk-radios__item">
-            <%= radio_button_tag("case_information[welsh_offender]", 'Yes', @case_info.welsh_offender == 'Yes', class: "govuk-radios__input") %>
+            <%= radio_button_tag("case_information[welsh_offender]", 'Yes', @case_info.welsh_offender == 'Yes', class: "govuk-radios__input", id: 'welsh_offender') %>
```
rossjones: This is okay, but I'd be tempted to put the id="welsh-offender" on the previous line (in the div). And same for the following ones, but it's up to you.

app/views/pages/contact_us.html.erb
```diff
@@ -32,53 +32,67 @@
...
         <div class="<%= field_error(@contact.errors, :email_address) %>">
-          <%= label_tag "email_address", "Email address", class: 'govuk-label' %>
+          <%= f.label "email_address", "Email address", class: 'govuk-label' %>
           <% if @contact.errors[:email_address].present? %>
                 <span class="govuk-error-message">
                   <%= @contact.errors[:email_address].first %>
                 </span>
+            <%= f.email_field("email_address", id: "email_address", class: "govuk-input govuk-input--error") %>
+          <% else %>
+            <%= f.email_field("email_address", id: "email_address", class: "govuk-input") %>
```
rossjones: Is there a better way of including `govuk-input--error` when there's an error?  I think the duplication is going to susceptible for errors if it changes and someone forgets to do both.

app/views/pages/contact_us.html.erb
```diff
@@ -32,53 +32,67 @@
...
         <div class="<%= field_error(@contact.errors, :job_type) %>">
-          <%= label_tag "Your job role", "Your job role", class: 'govuk-label' %>
+          <%= f.label "Your job role", "Your job role", class: 'govuk-label' %>
           <% if @contact.errors[:job_type].present? %>
                   <span class="govuk-error-message">
                     <%= @contact.errors[:job_type].first %>
                   </span>
+            <%= f.text_field("job_type", id: "job_type", class: "govuk-input govuk-input--error") %>
+          <% else %>
+            <%= f.text_field("job_type", id: "job_type", class: "govuk-input") %>
```
rossjones: Same as https://github.com/ministryofjustice/offender-management-allocation-manager/pull/961/files#r347998830

app/views/shared/_field_errors.html.erb
```diff
@@ -5,9 +5,13 @@
...
-        <li>
-          <%= link_to errors.full_messages_for(field).first, "##{form.object.class.name.underscore}_#{field}_error" %>
-        </li>
+        <% if local_assigns[:disable_refresh] %>
+          <li>
+            <%= link_to errors.full_messages_for(field).first, "##{field}", data: { turbolinks: false } %>
+          </li>
+        <% else %>
+          <li><%= link_to errors.full_messages_for(field).first, "##{field}" %></li>
+        <% end %>
```
rossjones: I'm not sure I understand why turbolinks is causing problems here, but I wonder if it is possible to replace the two mostly duplicated lines with a single line.  This is currently the equivalent of 
```ruby
if true 
   return false 
else 
   return true 
end 
``` 
and it should be possible to reverse the logic so :disable_refresh becomes :enable_refresh and then the single line can be ...  
```
<%= link_to errors.full_messages_for(field).first, "##{field}", data: { turbolinks: local_assigns.fetch(:enable_refresh, false) } %>
``` 
or something similar.

theKHutDeveloper:
POM 391 Investigate gov form-builder / SCSS shortcuts / Form-helpers might be a better solution to making our error messages conform to GDS guidelines. As a result I am closing this PR

PR #970, opened 2019-11-15
POM-409 enable Heroku deployments

starswan:


app.json
```diff
@@ -0,0 +1,23 @@
+{
+  "scripts": {
+    "postdeploy": "bundle exec rails db:seed",
+    "pr-predestroy": "bundle exec rake heroku:pr_destroy"
```
rossjones: Where does this rake task come from?

KPobeeNorris: Thanks for spotting this, it made it's way back in by mistake! New commit to remove this

PR #973, opened 2019-11-18
OffenderPresenter has a missing function that should be delegated

rossjones:
The OffenderPresenter needs to delegate nps_case? for situations where
it is used for determining the POM recommendation.

This was not caught because the tests used a struct where this was
defined.  Adding this function should make the test reflect the object
and should enable open prisons to allocate offenders.


rossjones:
Closed in favour of https://github.com/ministryofjustice/offender-management-allocation-manager/pull/973#pullrequestreview-318400880

PR #976, opened 2019-11-19
Pom 450 new arrivals

rossjones:
Show 'new arrivals' in summary

As per POM-450 this PR adds a new summary type (:new_arrivals) to the
summary, which is "offenders that would be in missing information if
they have been waiting more than 2 days".  To achieve this means we need
to include the arrival date for all offenders in all summary_types so
that we can calculate totals. This is not ideal.


app/services/summary_service.rb
```diff
@@ -39,17 +45,19 @@ def self.summary(summary_type, prison, page, params)
           bucket.items << offender if summary_type == :unallocated
           counts[:unallocated] += 1
         end
-      else
+      elsif offender.awaiting_allocation_for > 2
```
starswan: Would be nice if this were configurable or at least a named constant. Are we sure that awaiting_allocation_for is a number of days. If so it needs a better name

app/views/layouts/application.html.erb
```diff
@@ -11,7 +11,6 @@
 
   <body class="govuk-template__body js-enabled">
     <%= render '/layouts/header' %>
-    <%= render '/layouts/phase_banner' %>
```
kjdchapman: I'm not sure why this has been deleted in this PR?

KPobeeNorris: Good catch @kyle, it shouldn't have been deleted.  Putting it back now

app/views/layouts/errors_and_contact.html.erb
```diff
@@ -11,7 +11,6 @@
 
   <body class="govuk-template__body js-enabled">
     <%= render '/layouts/errors_and_contact_header' %>
-    <%= render '/layouts/phase_banner' %>
```
kjdchapman: I think this removal is accidental as well?

kjdchapman: (I could re-add this myself, but I want to double-check I haven't misunderstood something!)

app/views/summary/new_arrivals.html.erb
```diff
@@ -0,0 +1,82 @@
...
+
+<div class="govuk-!-margin-bottom-4">
+  <%= render "search/search_box" %>
+</div>
+
+<%= render(:partial => 'summary_subnav', :locals => {:active => :new_arrivals}) %>
+
+<h1 class="govuk-heading-xl govuk-!-margin-bottom-4">Newly arrived</h1>
+<p>
+  Case information for these prisoners will be matched with nDelius records in the
```
kjdchapman: Has @lmoney-moj had the chance to look at this content?

spec/spec_helper.rb
```diff
@@ -11,6 +11,8 @@
   # had to drop this by .01 as a result of re-structuring code and exposing a path that
   # we were not testing anyway.
   minimum_coverage 99.22
+  # 22 lines uncovered at 99.25% coverage
+  minimum_coverage 99.25
```
kjdchapman: Let's have a separate conversation about test coverage?

app/views/layouts/errors_and_contact.html.erb
KPobeeNorris: Re-added this @kjdchapman

PR #984, opened 2019-11-20
Pom 489 prd prompt ped ted in the past

starswan:


kjdchapman:
This can't be merged in its current state - it may or may not be useful as a start on work we want to do on asking POMs for more information, but we'll judge that later.

PR #985, opened 2019-11-22
Pom 520 - Stop crashing when POM is not correctly setup 

KPobeeNorris:
https://dsdmoj.atlassian.net/browse/POM-520

At present we crash if a POM is not properly set up i.e. a POM needs to
have the ALLOC_CASE_MGR role, and have a POM job type in classic NOMIS.
We therefore need to handle this more effectively, and redirect an
incorrectly set up POM to a 401 page.


app/controllers/prisons_application_controller.rb
```diff
@@ -23,4 +33,9 @@ def check_prison_access
 
     @prison = Prison.new(active_prison_id)
   end
+
+  def pom_at_active_prison?
+    user = Nomis::Elite2::UserApi.user_details(current_user)
+    StaffMember.new(user.staff_id).pom_at?(active_prison_id)
```
starswan: This feels like it should be a member variable like @prison which is available to all controllers via inheritance. Then we load it in a before action, and pom_at_active_prison? becomes @current_staff_member.pom_at?(active_prison_id)

spec/controllers/dashboard_controller_spec.rb
```diff
@@ -17,11 +17,11 @@
...
 
       before do
+        stub_auth_token
         stub_poms(prison, poms)
-        stub_sso_pom_data(prison)
+        stub_sso_pom_data(prison, 'alice')
         stub_signed_in_pom(1, 'Alice')
-        stub_request(:get, "https://gateway.t3.nomis-api.hmpps.dsd.io/elite2api/api/users/").
-          with(headers: { 'Authorization' => 'Bearer token' }).
+        stub_request(:get, "https://gateway.t3.nomis-api.hmpps.dsd.io/elite2api/api/users/alice").
```
starswan: I think you should now be able to delete this line - it should already be done by the stub_sso_pom_data

spec/controllers/tasks_controller_spec.rb
```diff
@@ -16,7 +16,7 @@
   let(:next_week) { Time.zone.today + 7.days }
 
   before do
-    stub_sso_pom_data(prison)
+    stub_sso_pom_data(prison, username)
```
starswan: now I'm slightly confused - I think we should only have 1 'login as POM' method but we seem to have 2 (stub_sso_pom_data and stub_signed_in_pom)

PR #994, opened 2019-11-27
POM-530 - Remove Ross as a POM on T3

KPobeeNorris:
Ross has left the team and as such we need to remove him as a POM on T3.
This PR replaces Ross with a generic POM that has been created to avoid
the issue of having to re-write tests each time a team member leaves.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2019_11_20_095310) do
+ActiveRecord::Schema.define(version: 2019_11_26_081703) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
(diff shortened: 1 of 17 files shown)


db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2019_11_20_095310) do
+ActiveRecord::Schema.define(version: 2019_11_26_081703) do
```
starswan: oops - I found this in my stuff, and made a PR to fix it

spec/features/coworking_feature_spec.rb
```diff
@@ -2,11 +2,11 @@
 
 feature 'Co-working' do
   let!(:nomis_offender_id) { 'G4273GI' }
-  let!(:probation_pom) do
+  let!(:prison_pom) do
```
starswan: name should probably just be :pom to show we don't care

PR #1025, opened 2019-12-13
Add sentence start date & responsibility override to debugging page

KPobeeNorris:


spec/features/debugging_feature_spec.rb
```diff
@@ -11,7 +11,7 @@
     fill_in 'offender_no', with: nomis_offender_id
     click_on('search-button')
 
-    expect(page).to have_css('tbody tr', count: 29)
+    expect(page).to have_css('tbody tr', count: 34)
```
starswan: Think it would have been much better to move this stuff towards a controller spec. This change doesn't actually show that the code works sadly.

PR #1029, opened 2019-12-16
Pom 574 - Fix POM responsibility calculation

KPobeeNorris:
We have been incorrectly calculating POM responsibility as needed to
take into account the length of time left to serve at the start of the
offender's sentence, and ensure that responsibility correctly flips to
supporting at the handover date.


spec/services/handover_date_service_spec.rb
```diff
@@ -3,32 +3,36 @@
...
+      let(:offender) {
+        OpenStruct.new indeterminate_sentence?: indeterminate,
+                       nps_case?: true,
+                       automatic_release_date: release_date,
+                       tariff_date: tariff_date
+      }
 
-        it 'is not set' do
-          expect(described_class.handover(offender).start_date).to be_nil
+      let(:release_date) { Date.new(2020, 8, 30) }
```
maysakanoni: Just trying to understand what some of these dates mean. Does it matter that this is in the future?

app/services/responsibility_service.rb
```diff
@@ -12,12 +12,17 @@ def to_s
   COWORKING = 'Co-Working'
   NPS = 'NPS'
 
+  # We currently don't have access to the date of the parole board decision, which means that we cannot correctly
+  # calculate responsibility for NPS indeterminate cases with parole eligibility where the TED is in the past.
+  # A decision has been made to display a notice so staff can check whether they need to override their case or not;
+  # this is until we get access to this data.
   def self.calculate_pom_responsibility(offender)
```
kjdchapman: Was this the gnarly "if statement" @KPobeeNorris mentioned? It is quite hard to follow as a naive reader!

Normally I'd suggest pairing to see if we can refactor it a bit, but I'd prefer to focus on getting this working and merged in, and catch it the next time we're in this area.

spec/helpers/override_helper_spec.rb
```diff
@@ -21,15 +21,26 @@
 
   describe '#complex_reason_label' do
     context 'when a prison POM' do
-      let(:subject) { OffenderPresenter.new(OpenStruct.new(immigration_case?: true), nil) }
+      # Immigration cases are always recommended to Prison POMs
```
kjdchapman: Could this be expressed with `describe`/`context`/`it` naming instead? I worry that comments won't stay up-to-date as the code changes further, which might confuse future readers.

spec/services/allocation_service_spec.rb
```diff
@@ -11,7 +11,7 @@
   describe '#allocate_secondary', :queueing do
     let(:kath_id) { 485_637 }
     let(:ross_id) { 485_926 }
-    let(:nomis_offender_id) { 'G7806VO' }
+    let(:nomis_offender_id) { 'G4273GI' }
```
kjdchapman: What made this spec change along with the change we're trying to make?

app/services/handover_date_service.rb
```diff
@@ -18,19 +20,32 @@ def self.handover(offender)
...
   def self.crc_handover_date(offender)
-    [
-        offender.home_detention_curfew_eligibility_date,
-        offender.conditional_release_date
+    crd_ard = [
+        offender.conditional_release_date,
+        offender.automatic_release_date
     ].compact.map { |date| date - 12.weeks }.min
+
+    if offender.home_detention_curfew_actual_date.present?
```
maysakanoni: whats the difference between `home_detention_curfew_eligibility_date` and `home_detention_curfew_actual_date`. When would one be entered over the other?

app/models/nomis/offender_base.rb
```diff
@@ -1,6 +1,7 @@
 module Nomis
   class OffenderBase
     delegate :home_detention_curfew_eligibility_date,
+             :home_detention_curfew_actual_date,
```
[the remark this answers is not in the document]
KPobeeNorris: @kjdchapman - this is the name of the field that we get back from our API call.  When calculating dates we look for the 'home_detention_curfew_actual_date' and if it's not present then we use the 'home_detention_curfew_eligibility_date'; as these two dates are linked there shouldn't be too many places where one is seen without the other.

app/services/handover_date_service.rb
```diff
@@ -51,38 +66,36 @@ def self.nps_handover_date(offender)
...
   def self.indeterminate_responsibility_date(offender)
     [
-        offender.parole_eligibility_date,
-        offender.tariff_date
+      offender.parole_eligibility_date,
+      offender.tariff_date
     ].compact.map { |date| date - 8.months }.min
   end
 
   # There are a couple of places where we need .5 of a month - which
```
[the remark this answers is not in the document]
KPobeeNorris: We have moved this to be closer to the next method as it's relevant to this

app/services/responsibility_service.rb
KPobeeNorris: Pairing on a refactor sounds good!

spec/helpers/override_helper_spec.rb
KPobeeNorris: We've updated the comment to be more specific

spec/services/allocation_service_spec.rb
KPobeeNorris: We needed an offender with the right combination of release dates for this set of specs to work properly

app/services/handover_date_service.rb
```diff
@@ -18,19 +20,32 @@ def self.handover(offender)
...
+        offender.automatic_release_date
+      ].compact.min - (7.months + 15.days)
     end
   end
 
   def self.crc_handover_date(offender)
-    [
-        offender.home_detention_curfew_eligibility_date,
-        offender.conditional_release_date
+    crd_ard = [
```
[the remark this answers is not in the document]
KPobeeNorris: Done!

app/services/handover_date_service.rb
```diff
@@ -4,7 +4,9 @@ class HandoverDateService
   HandoverData = Struct.new :start_date, :handover_date, :reason
 
   def self.handover(offender)
-    if offender.earliest_release_date.nil?
+    if offender.recalled?
+      HandoverData.new nil, nil, 'Recall case - no handover date calculation'
```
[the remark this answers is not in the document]
kjdchapman: We had a conversation about this, sounded like people were up for looking at cutting down our rubocop config / aligning with [govuk's linting rules](https://github.com/alphagov/rubocop-govuk).

spec/features/allocation_information_spec.rb
```diff
@@ -42,6 +43,54 @@
...
+                                                  conditional_release_date: "2020-11-02",
+                                                  automatic_release_date: "2020-11-02",
+                                                  hdced: "2020-11-02",
+                                                  release_date: "2023-01-01")
+
+        visit prison_allocation_path('LEI', nomis_offender_id: offender_no)
+      end
+
+      it "displays a POM as responsible" do
+        expect(page).to have_content('Responsible')
```
kjdchapman: Are there multiple people displayed on this page? Is this test specific enough?

spec/features/allocation_information_spec.rb
```diff
@@ -146,4 +195,34 @@ def create_allocation(offender_no)
...
   end
+
+private
+
+  def stub_api_calls_for_prison_allocation_path(sentence_start_date:, conditional_release_date:, automatic_release_date:, hdced:, release_date:)
+    stub_request(:post, "https://gateway.t3.nomis-api.hmpps.dsd.io/auth/oauth/token?grant_type=client_credentials").
+            to_return(status: 200, body: {}.to_json, headers: {})
+
+    stub_request(:get, "https://gateway.t3.nomis-api.hmpps.dsd.io/elite2api/api/prisoners/#{offender_no}").
+      to_return(status: 200, body: [{ offenderNo: offender_no, title: "MR", firstName: "OZULLIRN", middleNames: "TESSE", lastName: "ABBELLA", dateOfBirth: "1980-08-15", gender: "Male", sexCode: "M", nationalities: "sxiVsxi", currentlyInPrison: "Y", latestBookingId: "1153753", latestLocationId: "LEI", latestLocation: "Leeds(HMP)", internalLocation: "LEI-B-4-021", pncNumber: "97/39395W", croNumber: "043735/97V", ethnicity: "White:Eng./Welsh/Scot./N.Irish/British", birthCountry: "England", religion: "EfJSmIEfJSm", convictedStatus: "Convicted", imprisonmentStatus: "SENT03", receptionDate: "2016-11-26", maritalStatus: "sjedbztIaJRRzRZVIYadsjedbztIaJRRzRZVIYa" }].to_json)
```
kjdchapman: This is a large enough data structure that I think it's worth sticking it somewhere else - either as a variable somewhere in this file, or even in a flat file fixture.

That said, this problem is repeated across the app, so may be better to try and address once we have more context from working on it for longer.

spec/helpers/override_helper_spec.rb
```diff
@@ -21,15 +21,27 @@
 
   describe '#complex_reason_label' do
     context 'when a prison POM' do
-      let(:subject) { OffenderPresenter.new(OpenStruct.new(immigration_case?: true), nil) }
+      # we need to set up this test to return a Prison POM recommendation; we are using an
```
kjdchapman: Is there a way to express this information through the context blocks rather than comments? Comments are more likely to rot, I think.

spec/rails_helper.rb
```diff
@@ -58,6 +58,10 @@
     DatabaseCleaner.clean
   end
 
+  config.before(:each) do
```
kjdchapman: I'm a little worried about the speed implication of this... but probably more worried about the implications of it not being the case for all these tests! 

Big fan of this being its own commit, nice one.

spec/features/allocation_information_spec.rb
maysakanoni: No, only a single person. It would show up as either supporting or responsible.

spec/features/allocation_information_spec.rb
maysakanoni: Yeah, I agree- depending on our priorities right now, I'm happy to do either.

spec/helpers/override_helper_spec.rb
maysakanoni: we could have this, 
image
however were losing some of the context that immigration cases always evaluate to prison pom. not sure how important this information is

kjdchapman: I like that better! I'm not sure what the `context "when a prison POM"` adds by itself as well. I'm guessing the other branch of this (on line 33, "when a probation POM") is setting up an offender that's expected to be recommended to a probation POM?

You have more context than I do, but the ideal structure to me looks like:
```ruby
describe '#complex_reason_label' do
  context 'with an immigration case, which requires a prison POM' do
    # SNIP: stuff setting up the immigration case
    it 'notes the prisoner is not appropriate for a probation POM' do
      expect(subject.complex_reason_label).to eq('Prisoner assessed as not suitable for a prison officer POM')
    end
  end

  context 'with a case that is suitable for a probation POM' do
    # SNIP: stuff setting up that other offender

    it "notes that the prisoner can be assigned to a probation POM" do
      expect(subject.complex_reason_label).to eq('Prisoner assessed as suitable for a prison officer POM despite tiering calculation')
    end
  end
end
```

PR #1032, opened 2019-12-18
Refactor 2 phase construction

starswan:
Getting rid of 2-phase construction (new followed by attribute assignment) should allow us to convert a lot of attr_accessors into attr_readers and thus make our Offender models have greater encapsulation.

It also removes the SentenceDetail constructor used by tests, which hid a subtle bug around sentenceExpiryDate where we never read that field from NOMIS

### DIFF
--- a/app/models/hmpps_api/sentence_detail.rb
+++ b/app/models/hmpps_api/sentence_detail.rb
@@ -11,14 +11,10 @@ module HmppsApi
                 :licence_expiry_date,
                 :sentence_start_date,
                 :tariff_date,
-                :sentence_expiry_date,
                 :actual_parole_date
 
-    def initialize(fields = {})
-      fields.each do |k, v|
-        instance_variable_set("@#{k}", v)
-      end
-    end
+    # Note - this is hiding a defect - we never get sentence_expiry_date from NOMIS (but maybe we should?)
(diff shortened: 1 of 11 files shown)


kjdchapman:
Given this is a stale branch that requires some amount of effort to understand, I'm closing it for now. We can definitely re-open it if the team feels it has value / they have more context than I do!

starswan:
IMHO this is quite a useful improvement

app/models/hmpps_api/sentence_detail.rb
```diff
@@ -11,14 +11,10 @@ class SentenceDetail
...
-                :sentence_expiry_date,
                 :actual_parole_date
 
-    def initialize(fields = {})
-      fields.each do |k, v|
-        instance_variable_set("@#{k}", v)
-      end
-    end
+    # Note - this is hiding a defect - we never get sentence_expiry_date from NOMIS (but maybe we should?)
+    attr_accessor :sentence_expiry_date
```
ollietreend, line 17: If we never receive sentence expiry date in real life, why do we test against it? 🤔

This definitely seems a bit odd, so it's good to have this comment here. But it does make me wonder if we should just get rid of it if genuinely not used.

ollietreend (approved):

PR #1033, opened 2019-12-18
remove earliest_release_date from sentence details

starswan:
Comments welcomed - not convinced that this necessarily helps. The issue seems more around release_date - although recent comments from Louise Ball do seem to back up the use of the PED/TED/release_date field so maybe this isn't necessary?


KPobeeNorris:
Following clarification of responsibility and handover date calculations we are in the process of amending the various release dates we use, and are planning to remove 'earliest release date' as we use different data depending on offender's situations.  Therefore we are going to close this PR as this is no longer relevant.

PR #1066, opened 2020-02-03
Return 'unknown', not error, if dates are missing

kjdchapman:
Notably, this is a change in what this method returns, so if a component higher up the stack is expecting a Date (and this isn't covered by acceptance specs), this may not fix the issue.


kjdchapman:
We've decided to revert instead of attacking the problem this way (as there's more unknowns, and we can have better 'default' behaviour than this)

PR #1068, opened 2020-02-04
Pom 574

KPobeeNorris:
We have been incorrectly calculating POM responsibility as needed to take into account the length of time left to serve at the start of the offender's sentence, and ensure that responsibility correctly flips to supporting at the handover date.


app/services/responsibility_service.rb
```diff
@@ -105,7 +105,7 @@ def welsh_prepolicy_rules(offender)
     end
 
     def release_date_gt_15_mths_at_policy_date?(offender)
```
maysakanoni: I think we should rename this method

app/services/responsibility_service.rb
```diff
@@ -150,7 +151,7 @@ def hub_or_private?(offender)
     end
 
     def release_date_gt_mths_at_policy_date?(offender, threshold)
```
maysakanoni: I think we should rename this too

app/services/responsibility_service.rb
maysakanoni: `release_date_after_policy_start_date_plus_15_months`

PR #1073, opened 2020-02-06
Expose post recall release dates

KPobeeNorris:
Whilst investigating issues with POM-574 where we stopped using
Elite2's 'release_date' field in favour of the underlying dates, we
found that cases with post recall dates are not showing properly.
This exposes these dates within the app so we can investigate further


KPobeeNorris:
@kjdchapman - at the moment we're hoping to expose this date to check if it's useful as part of the work on POM-574; if we can make of it then there will certainly be tests added to any amendments to the calculations etc. :)

PR #1107, opened 2020-03-06
Update dates for 'earliest release date'

KPobeeNorris:
Since starting MOIC we have iterated in regards to the different dates
we use to determine if an offender is sentenced, calculating POM
responsibility and Handover dates.  However, we haven't updated the
dates we display to reflect these changes, and it has led to some
confusion for users.

This PR focusses on removing all references to 'release date' as other
than using this to determine if an offender is sentenced we do not use
it.  In addition, we have updated the dates we use when calculating
responsibility and handover, so the 'earliest release date' method now
matches these.

There are offenders where one or more of the dates we use to calculate
the 'earliest release date' falls into the past, and yet have another
date set in the future.  It has been agreed that a sensible approach to
earliest_release_date is to iterate over the dates we get back and
determine if they're in the future or the past.  If the offender has any
dates in the future we will choose the earliest of those, and if they
only have dates in the past we will then default to the earliest of
these.  This should avoid us displaying release related dates that have
already past, but where an offender also has other dates in the future.

We are looking to replace the release date with the sentence start date
on some of the pages within the service.  There are times when offenders
sentences haven't been entered correctly on NOMIS and this can lead to
their sentence start date being wrong/absent.  The hope is that by
exposing this data to SPO/POMs that they can fix this data themselves
rather than contacting us via Zendesk.

### DIFF
--- a/app/models/nomis/sentence_detail.rb
+++ b/app/models/nomis/sentence_detail.rb
@@ -29,7 +29,10 @@ module Nomis
 
     def earliest_release_date
       dates = [
-          release_date,
+          automatic_release_date,
+          conditional_release_date,
+          home_detention_curfew_actual_date,
+          home_detention_curfew_eligibility_date,
           parole_eligibility_date,
           tariff_date
       ].compact
--- a/app/models/nomis/sentence_detail.rb
+++ b/app/models/nomis/sentence_detail.rb
@@ -38,7 +38,18 @@ module Nomis
       ].compact
       return nil if dates.empty?
 
-      dates.min.to_date
+      past_dates = []
+      future_dates = []
+
+      dates.each do |date|
+        date >= Time.zone.today ? future_dates << date : past_dates << date
+      end
+
+      if future_dates.present?
+        future_dates.min.to_date
+      else
+        past_dates.min.to_date
+      end
     end
 
     def full_name
--- a/app/models/nomis/sentence_detail.rb
+++ b/app/models/nomis/sentence_detail.rb
@@ -42,14 +42,14 @@ module Nomis
       future_dates = []
 
       dates.each do |date|
-        date >= Time.zone.today ? future_dates << date : past_dates << date
+        if date >= Time.zone.today
+          future_dates << date
+        else
+          past_dates << date
+        end
       end
 
-      if future_dates.present?
-        future_dates.min.to_date
-      else
-        past_dates.min.to_date
-      end
+      future_dates.present? ? future_dates.min.to_date : past_dates.min.to_date
     end
 
     def full_name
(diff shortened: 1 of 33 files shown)


app/models/nomis/sentence_detail.rb
```diff
@@ -29,13 +29,27 @@ def conditional_release_date
...
           tariff_date
       ].compact
       return nil if dates.empty?
 
-      dates.min.to_date
+      past_dates = []
+      future_dates = []
+
+      dates.each do |date|
+        date >= Time.zone.today ? future_dates << date : past_dates << date
```
kjdchapman: I found this a little hard to parse - I'll happily use ternary operators for assignment:
```ruby
some_variable = condition ? first_possibility : second_possibility
```

but I'll rarely use it for operations unless they're all methods
```ruby
# happy with this
condition ? do_thing : do_second_thing

# starting to lose track!
some + maths ? comparisons > thing : otherwise << thing_two
```

My first thought is to split this up a little (not sure if this is exactly right, but feels closer):
```ruby
date_has_passed = date >= Time.zone.today
dates_to_compare = date_has_passed ? future_dates : past_dates
dates_to_compare << date
```

Or maybe just a straight `if` is cleaner?
```ruby
if date >= Time.zone.today 
  future_dates << date
else
  past_dates << date
end
```

spec/features/pom_caseload_feature_spec.rb
```diff
@@ -159,11 +159,11 @@
 
     it 'can be sorted by release date' do
       page.all('th')[2].find('a').click
-      within '.offender_row_4' do
-        expect(page).to have_content('Allix, Aobmethani')
+      within '.offender_row_6' do
```
kjdchapman: Are these tests too brittle? Do they provide value?

app/views/summary/unallocated.html.erb
```diff
@@ -67,7 +67,7 @@
         </td>
 
         <td aria-label="Earliest release<br/>date" class="govuk-table__cell">
-          <%= format_date(offender.earliest_release_date) %>
+          <%= format_date(offender.earliest_release_date, replacement: 'Unknown') %>
```
kjdchapman: I like these "Unknown" labels - would it be good to have a test covering at least one of these pages, though? 

It may just be a case of adding an expectation to `when_there_is_missing_information_spec.rb`?

KPobeeNorris: Expectation added to `when_nomis_is_missing_information_spec.rb` to cover the "Unknown" labels

app/models/nomis/sentence_detail.rb
KPobeeNorris: Refactored this section of the method; hopefully making it a little easier to read

PR #1108, opened 2020-03-06
POM-637 Get POM name from NOMIS

KPobeeNorris:
We have received a report from a user advising that one of their POM’s
has changed their name, it’s been updated in NOMIS but it’s still
showing up as the old name on DPS.  A quick investigation unearthed that
when DPS calls the MOIC API we are sending back the POM name in the
allocation record, and not from NOMIS, so it won’t show the correct name
if the POM has changed their name.  We need to update this, and other
places where we display the POM name in MOIC to fetch the name from
NOMIS to ensure it’s accurate/up to date.

I have also updated the prisoner page to get the POM name from NOMIS too.

### DIFF
--- a/app/controllers/prisoners_controller.rb
+++ b/app/controllers/prisoners_controller.rb
@@ -15,10 +15,13 @@ class PrisonersController < PrisonsApplicationController
     @tasks = PomTasks.new.for_offender(@prisoner)
     @allocation = Allocation.find_by(nomis_offender_id: @prisoner.offender_no)
 
-    @primary_pom_name = helpers.fetch_pom_name(@allocation.primary_pom_nomis_id).titleize if @allocation.present?
+    if @allocation.present?
+      @primary_pom_name = helpers.fetch_pom_name(@allocation.primary_pom_nomis_id).
+          titleize
+    end
 
     if @allocation.present? && @allocation.secondary_pom_name.present?
-      @secondary_pom_name = fetch_pom_name(@allocation.secondary_pom_nomis_id).titleize
+      @secondary_pom_name = helpers.fetch_pom_name(@allocation.secondary_pom_nomis_id).titleize
     end
 
     @pom_responsibility = pom_responsibility
--- a/app/controllers/prisoners_controller.rb
+++ b/app/controllers/prisoners_controller.rb
@@ -15,10 +15,10 @@ class PrisonersController < PrisonsApplicationController
     @tasks = PomTasks.new.for_offender(@prisoner)
     @allocation = Allocation.find_by(nomis_offender_id: @prisoner.offender_no)
 
-    @primary_pom_name = fetch_pom_name(@allocation.primary_pom_nomis_id) if @allocation.present?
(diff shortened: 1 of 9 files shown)


spec/features/when_nomis_is_missing_information_spec.rb
```diff
@@ -83,6 +83,9 @@
           latestBookingId: booking_id
         }]
 
+        stub_request(:get, "#{stub_api_host}/staff/#{staff_id}").
+            to_return(status: 200, body: { staffId: staff_id, firstName: "TEST", lastName: "MOIC" }.to_json)
```
kjdchapman: nit: consistent indentation
```suggestion
          to_return(status: 200, body: { staffId: staff_id, firstName: "TEST", lastName: "MOIC" }.to_json)
```

app/controllers/prisoners_controller.rb
```diff
@@ -38,6 +44,12 @@ def image
 
 private
 
+  def fetch_pom_name(staff_id)
+    pom_firstname, pom_secondname =
+      PrisonOffenderManagerService.get_pom_name(staff_id)
+    "#{pom_secondname.titleize}, #{pom_firstname.titleize}"
```
kjdchapman: This formatting rule is in multiple places - is there a good way to DRY it up? (might be beyond the scope of this PR)

app/controllers/prisoners_controller.rb
```diff
@@ -15,6 +15,12 @@ def show
     @tasks = PomTasks.new.for_offender(@prisoner)
     @allocation = Allocation.find_by(nomis_offender_id: @prisoner.offender_no)
 
+    @primary_pom_name = fetch_pom_name(@allocation.primary_pom_nomis_id) if @allocation.present?
```
kjdchapman: Introducing more instance variables for the two names (`@primary...` and `@secondary...`) suggests to me that we're doing too much in this controller. Is there a way to avoid this?

spec/features/when_nomis_is_missing_information_spec.rb
KPobeeNorris: Indentation fixed

app/controllers/prisoners_controller.rb
KPobeeNorris: Moved this method to application_helper.rb so it can be used across controllers, not 100% sure this was what you were thinking here?

kjdchapman: That makes it live in a single place, which is good!

My worry with putting it in `application_helper` is that it's quite general - if we do this with lots of things we end up with all sorts of functionality mixed together. That can make it hard to read/understand in the future.

I like to associate things like this with a particular model, so I have some `PrisonerOffenderManager` object, and I can ask it for its `.name`, making it the responsibility of that class to do any formatting/retrieving/orchestrating*. Because of the structure of the code right now, this would probably take too much work for this PR, though?

*(if one of these tasks became a large enough job, I'd happy split it further into presenters/repositories/whatever)

KPobeeNorris: Moved the method to the pom_helper, so it's still available centrally, but in a helper specifically related to POMs

PR #1109, opened 2020-03-09
Appease Rubocop by renaming Cop and updating config to include new Cops

KPobeeNorris:


KPobeeNorris:
We had a recent-ish Dependabot update for Rubocop so imagine these warnings started appearing around that time.  This doesn't currently cause Rubocop to fail, but thought it would be good to reduce the noise during the runs

PR #1110, opened 2020-03-09
Fix responsibility override email bug

KPobeeNorris:
We have been seeing Sentry alerts
(https://sentry.service.dsd.io/mojds/live1-allocation-manager-produ/issues/40529/?referrer=slack)
 related to missing email addresses for the responsibility override
feature; where an SPO can override responsibility to the community.
Investigation into this bug uncovered a bug where if an offender is
linked to an LDU, but that LDU has an empty string for an email address
then it still allows the SPO to complete the override.  However, this
will clearly not result in an email being sent and so this PR adds a
specific check for the presence of an email to prevent this happening in
the future.

### DIFF
--- a/app/controllers/responsibilities_controller.rb
+++ b/app/controllers/responsibilities_controller.rb
@@ -2,7 +2,7 @@
 
 class ResponsibilitiesController < PrisonsApplicationController
   def new
-    if ldu_email_address(params[:nomis_offender_id])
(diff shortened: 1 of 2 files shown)


kjdchapman:
Would it be good to cover this with a test?

KPobeeNorris:
Spec added to cover the scenario of an LDU having an empty string in the email_address field

PR #1111, opened 2020-03-12
Don't adjust handover date for MAPPA if it's long before release dates

kjdchapman:
A case was spotted where an offender is due to be released some time away (longer than 4.5 months, which is the general amount of time handover happens for), but is displaying *today* as their handover date. This also means their responsibility is showing as community rather than custody.

This was due to a 'MAPPA' flag being set in nDelius, which our service takes to mean "this person is being released imminently, so hand them over now". In actuality, this flag can be set far in advance of a handover process (even if it "shouldn't" be set), so this can lead our service to display confusing/incorrect information.

We suspect this is something that will happen a fair amount, so changing our service to respond sensibly is a good idea.

[The behaviour for home detention dates already looks like it's correct to me](https://github.com/ministryofjustice/offender-management-allocation-manager/blob/master/app/services/handover_date_service.rb#L105-L112), as it takes the release dates as well.

### DIFF
--- a/app/services/handover_date_service.rb
+++ b/app/services/handover_date_service.rb
@@ -93,8 +93,13 @@ private
     ].compact.map { |date| date - 8.months }.min
   end
 
-  def self.mappa_23_responsibility_date(_offender)
-    Time.zone.today
+  def self.mappa_23_responsibility_date(offender)
+    earliest_date = [
+      offender.conditional_release_date,
+      offender.automatic_release_date
+    ].compact.map { |date| date - (4.months + 15.days) }.min
+
+    [Time.zone.today, earliest_date].max
   end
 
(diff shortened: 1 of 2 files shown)


kjdchapman:
With a little time / conversation spent on this, I can't see a better way to structure the test doubles for now, so I'd be happy to merge this - comments still very welcome though.

ollietreend:
This looks good to me @kjdchapman. The business logic and meaning of MAPPA and associated TLAs aren't clear to me yet – so I trust that this is correct. But as for your code changes – they make sense!

As for the multiple `let` blocks that you pointed out as code smell in the spec – I suppose it could be DRYer to assign them using an array & loop, but I prefer the explicitness of simply listing them all as you've done.

PR #1118, opened 2020-03-25
POM-566: SPO to see cases coming up for handover at their prison

ollietreend:
This Pull Request adds a new 'Handover cases' summary page under the 'Allocate a prisoner to an offender manager' section. This page is accessible to SPOs, and displays all offenders in the currently selected prison who are approaching handover.

This new page will show offenders whose:
- handover start date is within 30 days, or
- handover is in progress, but has not completed – i.e. the handover start date has passed, but the responsibility hasn't yet switched to the community

### DIFF
--- a/app/controllers/summary_controller.rb
+++ b/app/controllers/summary_controller.rb
@@ -79,15 +79,22 @@ private
   end
 
   def sort_params(summary_type)
-    if params['sort'].blank?
-      return [:sentence_start_date, :asc] unless summary_type == :allocated
-
-      return [nil, nil]
-    end
+    return default_sort_params(summary_type) if params['sort'].blank?
 
     parts = params['sort'].split.map { |s| s.downcase.to_sym }
     return [parts[0], :asc] if parts.count == 1
 
     parts
+  end
+
+  def default_sort_params(summary_type)
+    case summary_type
+    when :allocated
+      [nil, nil]
+    when :handovers
+      [:handover_start_date, :asc]
+    else
+      [:sentence_start_date, :asc]
+    end
   end
 end
--- a/app/controllers/summary_controller.rb
+++ b/app/controllers/summary_controller.rb
@@ -48,6 +48,14 @@ class SummaryController < PrisonsApplicationController
     @summary = SummaryPresenter.new summary
   end
 
+  def handovers
+    summary = create_summary(:handovers)
+    items = offenders(:handovers, summary.handovers)
+
+    @offenders = Kaminari.paginate_array(items).page(page)
+    @summary = SummaryPresenter.new summary
+  end
+
 private
 
   def offenders(summary_type, offenders)
--- a/app/controllers/summary_controller.rb
+++ b/app/controllers/summary_controller.rb
@@ -79,12 +79,16 @@ private
   end
 
   def sort_params(summary_type)
-    return default_sort_params(summary_type) if params['sort'].blank?
+    sort_params = params['sort'] || ''
+    parts = sort_params.split.map { |s| s.downcase.to_sym }
 
-    parts = params['sort'].split.map { |s| s.downcase.to_sym }
-    return [parts[0], :asc] if parts.count == 1
-
-    parts
+    if parts.blank?
+      default_sort_params(summary_type)
+    elsif parts.second.blank?
+      parts + [:asc]
+    else
+      parts
+    end
   end
 
   def default_sort_params(summary_type)
--- a/app/services/summary_service.rb
+++ b/app/services/summary_service.rb
@@ -31,8 +31,7 @@ class SummaryService
     offenders = prison.offenders
     active_allocations_hash = AllocationService.active_allocations(offenders.map(&:offender_no), prison.code)
 
-    # Add POM allocation details to offender objects. We use this information in 'allocated' and 'handovers' views.
-    add_allocated_poms(offenders, active_allocations_hash)
+    add_allocated_poms_and_coms(offenders, active_allocations_hash)
 
     # We need arrival dates for all offenders and all summary types because it is used to
     # detect new arrivals and so we would not be able to count them without knowing their
@@ -123,12 +122,18 @@ private
     end
   end
 
-  def self.add_allocated_poms(offenders, active_allocations_hash)
+  def self.add_allocated_poms_and_coms(offenders, active_allocations_hash)
     offenders.each do |offender|
-      alloc = active_allocations_hash[offender.offender_no]
       next unless active_allocations_hash.key?(offender.offender_no)
+
+      alloc = active_allocations_hash[offender.offender_no]
+
+      # Add POM details
       offender.allocated_pom_name = restructure_pom_name(alloc.primary_pom_name)
       offender.allocation_date = (alloc.primary_pom_allocated_at || alloc.updated_at)&.to_date
+
+      # Add COM details
+      offender.allocated_com_name = alloc.com_name
     end
   end
 
--- a/app/services/summary_service.rb
+++ b/app/services/summary_service.rb
@@ -31,6 +31,9 @@ class SummaryService
     offenders = prison.offenders
     active_allocations_hash = AllocationService.active_allocations(offenders.map(&:offender_no), prison.code)
 
+    # Add POM allocation details to offender objects. We use this information in 'allocated' and 'handovers' views.
+    add_allocated_poms(offenders, active_allocations_hash)
+
     # We need arrival dates for all offenders and all summary types because it is used to
     # detect new arrivals and so we would not be able to count them without knowing their
     # prison_arrival_date
@@ -55,12 +58,6 @@ class SummaryService
       if approaching_handover?(offender)
         buckets[:handovers].items << offender
       end
-    end
-
-    # For the allocated and handover offenders, we need to provide the allocated POM's name
-    [:allocated, :handovers].each do |bucket_name|
-      bucket_offenders = buckets[bucket_name].items
-      add_allocated_poms(bucket_offenders, active_allocations_hash)
     end
 
     Summary.new(summary_type, allocated: buckets[:allocated],
--- a/app/services/summary_service.rb
+++ b/app/services/summary_service.rb
@@ -52,8 +52,9 @@ class SummaryService
         buckets[:pending].items << offender
       end
 
-      # TODO: Add filtering logic for upcoming handover cases. All offenders are included for now.
-      buckets[:handovers].items << offender
+      if approaching_handover?(offender)
+        buckets[:handovers].items << offender
+      end
     end
 
     # For the allocated and handover offenders, we need to provide the allocated POM's name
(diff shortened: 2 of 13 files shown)


app/controllers/summary_controller.rb
```diff
@@ -71,15 +79,22 @@ def page
   end
 
   def sort_params(summary_type)
-    if params['sort'].blank?
-      return [:sentence_start_date, :asc] unless summary_type == :allocated
-
-      return [nil, nil]
-    end
+    return default_sort_params(summary_type) if params['sort'].blank?
```
kjdchapman: This is a pre-existing problem with this method, but I find it hard to parse the control flow here. It seems like this part isn't well-covered by tests, though, so I would understand not wanting to change it.

That said, my thoughts expressed in code, trying to express the policy of "do something different depending on which params are provided":
```ruby
def sort_params(summary_type)
  sort_params = params['sort'] || ''
  parts = sort_params.split.map { |s| s.downcase.to_sym }

  if parts.blank?
    default_sort_params(summary_type)
  elsif parts.second.blank?
    parts + [:asc]
  else
    parts
  end
end
```

app/controllers/summary_controller.rb
```diff
@@ -71,15 +79,22 @@ def page
...
 
     parts = params['sort'].split.map { |s| s.downcase.to_sym }
     return [parts[0], :asc] if parts.count == 1
 
     parts
   end
+
+  def default_sort_params(summary_type)
```
kjdchapman: I like splitting this off into a method, nice one.

app/services/summary_service.rb
```diff
@@ -22,12 +24,16 @@ def self.summary(summary_type, prison)
...
                 pending: Bucket.new(sortable_fields),
-                new_arrivals: Bucket.new(sortable_fields)
+                new_arrivals: Bucket.new(sortable_fields),
+                handovers: Bucket.new(sortable_fields)
     }
 
     offenders = prison.offenders
     active_allocations_hash = AllocationService.active_allocations(offenders.map(&:offender_no), prison.code)
 
+    # Add POM & COM allocation details to offenders. We use this information in 'allocated' and 'handovers' views.
```
kjdchapman: I'd be happy to drop this comment, the method name is good enough to express it, I think.

app/services/summary_service.rb
```diff
@@ -48,20 +54,14 @@ def self.summary(summary_type, prison)
...
       end
-    end
 
-    # For the allocated offenders, we need to provide the allocated POM's
-    # name
-    buckets[:allocated].items.each do |offender|
-      alloc = active_allocations_hash[offender.offender_no]
-      offender.allocated_pom_name = restructure_pom_name(alloc.primary_pom_name)
-      offender.allocation_date = (alloc.primary_pom_allocated_at || alloc.updated_at)&.to_date
+      # Add to 'Handover cases' bucket if offender is approaching their handover date
```
kjdchapman: I'd be happy to drop this comment, too, the associated code is clear.

app/services/summary_service.rb
```diff
@@ -77,6 +77,22 @@ def self.new_arrival?(offender)
...
+
+    return false if start_date.nil?
+
+    if start_date.future?
+      start_date.between?(today, thirty_days_time)
+    else
+      today.between?(start_date, handover_date)
+    end
+  end
+
```
kjdchapman: I think this method is re-expressing some of the logic found in the [handover date service](https://github.com/ministryofjustice/offender-management-allocation-manager/blob/master/app/services/handover_date_service.rb). 

This method is probably clearer than our existing implementation! But I'm worried that we'll fail to change both in tandem, in the future, especially given the test coverage here.

Is there any good way to bring them together without extensive refactoring of the `handover_date_service`? I'm not sure, but bringing it up in case we have any good ideas! :sweat_smile:

app/services/summary_service.rb
```diff
@@ -85,6 +101,10 @@ def self.sort_fields_for_new_arrivals
     [:last_name, :prison_arrival_date, :earliest_release_date]
   end
 
+  def self.sort_fields_for_handovers
```
kjdchapman: These `sorting` behaviours being in the `summary_service` as well as the `summary_controller` smells to me.

(I appreciate we're following an existing pattern, so I'm not saying change it now!)

app/services/summary_service.rb
```diff
@@ -100,6 +120,21 @@ def self.add_arrival_dates(offenders)
     end
   end
 
+  def self.add_allocated_poms_and_coms(offenders, active_allocations_hash)
+    offenders.each do |offender|
+      next unless active_allocations_hash.key?(offender.offender_no)
```
kjdchapman: This "loop guard clause" is  :100: ruby goodness! :tada:

app/services/summary_service.rb
```diff
@@ -100,6 +120,21 @@ def self.add_arrival_dates(offenders)
...
+
+      alloc = active_allocations_hash[offender.offender_no]
+
+      # Add POM details
+      offender.allocated_pom_name = restructure_pom_name(alloc.primary_pom_name)
+      offender.allocation_date = (alloc.primary_pom_allocated_at || alloc.updated_at)&.to_date
+
+      # Add COM details
+      offender.allocated_com_name = alloc.com_name
```
kjdchapman: I'm not familiar with this part of the domain, please forgive my innocent question: Do COM names not need the same formatting as POM names?

app/views/dashboard/index.html.erb
```diff
@@ -53,6 +53,14 @@
       <p class="govuk-body">Prisoners whose case information has not yet been matched with nDelius.</p>
     </div>
   </div>
+  <div class="govuk-grid-column-one-third">
+    <div class="wrapper">
+      <h1 class="govuk-heading-s">
+        <%= link_to "See all handover cases", prison_summary_handovers_path(@prison.code), class:"govuk-link" %>
+      </h1>
+      <p class="govuk-body">All cases that are within the handover period or start the handover process to the community in the next 30 days.</p>
```
kjdchapman: I think this sentence is too long, and people will have trouble parsing it (I'm not even sure if the title/helper text is the right way to frame these).

```suggestion
      <p class="govuk-body">Cases that need handing over to the community soon.</p>
```

(I'm not a content designer, just suffering from hubris. It's possible our actual content designer looked at this, too. Hmm.)

app/views/summary/_summary_subnav.html.erb
```diff
@@ -27,6 +27,13 @@
         Newly arrived (<%= @summary.new_arrivals_total %>)
       </a>
     </li>
+
+    <li class="moj-sub-navigation__item">
+      <a class="moj-sub-navigation__link" <%= 'aria-current=page' if active == :handovers %>
+         href="<%= prison_summary_handovers_path(@prison.code) %>">
```
kjdchapman: nit: indenting
```suggestion
        href="<%= prison_summary_handovers_path(@prison.code) %>">
```

app/views/summary/handovers.html.erb
```diff
@@ -0,0 +1,100 @@
...
+
+<div class="govuk-!-margin-bottom-4">
+  <%= render "search/search_box" %>
+</div>
+
+<%= render(:partial => 'summary_subnav', :locals => {:active => :handovers}) %>
+
+<h1 class="govuk-heading-xl govuk-!-margin-bottom-4">Handover cases</h1>
+<p>
+  All cases that are within the handover period or start the handover process to the community in the next 30 days.
```
kjdchapman: Similar thought to above, I'll wait to hear if @lmoney-moj 's had the chance to look at these rather than rabbiting on at poor @ollietreend !

spec/features/spo_handover_cases_feature_spec.rb
```diff
@@ -0,0 +1,58 @@
+require "rails_helper"
+
+feature "SPO viewing upcoming handover cases" do
+  let(:prison) { 'LEI' }
+
+  context 'when signed in as an SPO', vcr: { cassette_name: :spo_handover_cases_feature } do
+    before do
+      allow_any_instance_of(Nomis::OffenderBase).to receive_messages(
```
kjdchapman: Overriding constants is nasty, but I think this is the best way to do this given the context. Nice find!

app/services/summary_service.rb
ollietreend: That could be the case, although my understanding was that the handover date service had a slightly different responsibility – being more concerned with calculating when the handover should happen, i.e. to calculate the handover dates themselves. (Maybe it should be the 'responsibility date calculator service'? 😅)

The "is this offender approaching handover?" question currently comes with the baked-in assumption that "approaching handover" means "handovers starting within the next 30 days". I'm not sure that "in the next 30 days" is a global definition of when a handover is approaching, or more specific to this 'handover cases' SPO summary page.

For example, POMs have a [disturbingly similar page](https://github.com/ministryofjustice/offender-management-allocation-manager/blob/master/app/controllers/caseload_controller.rb#L31-L35), to see offenders which are approaching handover. However it [doesn't follow exactly the same criteria](https://github.com/ministryofjustice/offender-management-allocation-manager/blob/master/app/controllers/caseload_controller.rb#L39-L53): it doesn't include offenders who are already in their handover period.

I definitely agree that this could be cleaned up – especially given the similarities between POM/COM functionality. I'm not sure I have any good ideas for moving towards that goal in this PR 😅 but open to suggestions.

app/services/summary_service.rb
ollietreend: Sorting and pagination on the whole feels quite bespoke. I think we both assumed that a gem was being used to achieve this – but it turns out that's not the case.

I also noticed that the pagination/sort functionality in the view layer are disconnected from the controller layer – for example, the sorting helper [assumes `:last_name` is the default](https://github.com/ministryofjustice/offender-management-allocation-manager/blob/POM-566/app/helpers/sort_helper.rb#L8) sort column, despite [the controller having other opinions](https://github.com/ministryofjustice/offender-management-allocation-manager/blob/POM-566/app/controllers/summary_controller.rb#L90-L98).

Definitely something to look at when we can!

app/services/summary_service.rb
ollietreend: Good question. I'm not familiar with how these COM names are stored, so this is also a bit unclear to me. I haven't yet been able to get my hands on an offender object which has an COM assigned (I'm guessing this sort of data probably lives mostly on preprod or prod...?)

That said – our subject matter expert didn't flag any issues with how names are displaying when reviewing it. It seems like they might already have been 'restructured' elsewhere, or perhaps come in like that from the nDelius import.

app/views/dashboard/index.html.erb
ollietreend: This content has been changed by our content designer – new commit on its way.

app/services/summary_service.rb
kjdchapman: I think the `handover_date_service` deals with both concepts - both "approaching" handover and the actual date responsibility switches from POM to COM. The [data it returns](https://github.com/ministryofjustice/offender-management-allocation-manager/blob/master/app/services/handover_date_service.rb#L4) refers to these as "start date" and "handover date" respectively.

The 30-day thing though, I think is new, and specific to this page. Argh. I think you're right that moving towards it in this PR isn't good, but I do think that those two core classes - `handover_date_service` and `responsibility_service` - need some love soon.

app/services/summary_service.rb
kjdchapman: We'll soon notice this if it breaks, so np.

kjdchapman:
Oops, I forgot to mention first time round - how do you feel about 50/72 commit messages? ([GOV style guide](https://github.com/alphagov/styleguides/blob/master/git.md#example-1))

app/controllers/summary_controller.rb
ollietreend: Implemented ⬆️ thanks for this

PR #1119, opened 2020-03-26
Lint changed files w/GOV.UK rubocop

kjdchapman:
This PR introduces a [pre-commit git hook](https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks) that lints your changed files, using the [GOV.UK rubocop rules](https://github.com/alphagov/rubocop-govuk). 

The build still runs against the 'old' linting rules, so you can ignore this completely, if you like.

If you want to use this, run `make setup` in the root of the project directory. 
This will install the other existing pre-commit hooks (around `git-crypt`) as well.

(You can also install this hook, but ignore it for any given commit with [the `--no-verify` flag](https://git-scm.com/docs/git-commit)).

### DIFF
--- a/.rubocop-govuk.yml
+++ b/.rubocop-govuk.yml
@@ -1,0 +1,4 @@
+inherit_gem:
+  rubocop-govuk:
+    - config/default.yml
+    - config/rails.yml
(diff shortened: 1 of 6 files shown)


kjdchapman:
You have (quite correctly) called me on my laziness :grin:  - fixed in https://github.com/ministryofjustice/offender-management-allocation-manager/pull/1119/commits/1e15e6d87b3421edffdc5881eb5d34a4c92fb51d

PR #1120, opened 2020-03-30
Show community information to SPO users

ollietreend:
Community information is now visible to SPO users on the 'allocation information' and 'new allocation' pages. This closes Jira ticket POM-650.

Previously, community information was only visible to the allocated POM user. However, the decision was made for SPOs to be able to see this information to allow them to cover for POMs who are unable to work.

Since there is so much cross-over between these pages and the POM's 'view prisoner' page, I've reused the 'Community information' partial from that page (`prisoners/community_information`). Eventually, an effort should be made to reconcile all the different prisoner/allocation views, since there's currently a lot of duplication. That task is captured in Jira ticket POM-366, and is a much larger piece of work than this ticket.


kjdchapman:
Would be good to cover this with a basic acceptance spec ("smoke test"?) if we can - but I don't know how feasible that is.

PR #1132, opened 2020-04-08
Fix failing tests

maysakanoni:
Currently, some test that use VCR cassettes are failing on master. This is because allot of the test rely on external data. As we knew why this data changed, it was sufficient amend the tests to reflect this change.

More specifically, some tests relating to the number of prison offender managers (POM) were failing, this was because we had a new joiner who we added as a POM.

There were also test failures with regards to the responsibility calculations, this was due to an error in testing and not a bug in the code.

### DIFF
--- a/app/services/responsibility_service.rb
+++ b/app/services/responsibility_service.rb
@@ -25,7 +25,7 @@ class ResponsibilityService
       RESPONSIBLE
     elsif offender.indeterminate_sentence? && (offender.tariff_date.nil? ||
        offender.tariff_date < Time.zone.today)
-       
+
       RESPONSIBLE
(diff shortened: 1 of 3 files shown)


app/services/responsibility_service.rb
```diff
@@ -25,6 +25,7 @@ def self.calculate_pom_responsibility(offender)
       RESPONSIBLE
     elsif offender.indeterminate_sentence? && (offender.tariff_date.nil? ||
        offender.tariff_date < Time.zone.today)
+
```
maysakanoni, line 28: I added the extra line in the wrong place 🤦

PR #1135, opened 2020-04-27
Add a feature spec for the Delius import job

maysakanoni:
This PR covers the happy path as things are currently working with the Delius import Job.  The Delius import requires some significant work so that the teams can be mapped to the correct LDU. Currently, this is not entirely correct so there are plans to use this import to fix this issue. This feature test (which is more rigours than the existing unit tests) is required so that the refactors can be made with confidence that we're not regressing by breaking existing functionality.

Co-authored-by: Ollie Treend <ollie@treend.com>


spec/features/delius_import_feature_spec.rb
```diff
@@ -1,143 +1,63 @@
...
-  let(:team) { create(:team, local_divisional_unit: ldu) }
-
-  before do
-    test_strategy = Flipflop::FeatureSet.current.test!
-    test_strategy.switch!(:auto_delius_import, true)
-  end
+require 'support/lib/mock_imap'
+require 'support/lib/mock_mail'
+
+RSpec.describe DeliusImportJob do
```
maysakanoni, line 5: It doesnt feel correct to use `RSpec.describe DeliusImportJob`  in a feature spec, however the stubs were not working correctly when this was not used. Is this rspec thing?

ollietreend: To add some context – we wanted to define the spec as something like this:

```ruby
describe 'the Delius import' do
end
```

(or even using `feature` instead of `describe`)

But for some reason, this gives us an error relating to our stub of the `Net::IMAP` const:

```
TypeError:
       superclass mismatch for class IMAP
```

However, those superclass mismatch errors go away if we pass the `DeliusImportJob` class itself to `describe`, rather than passing a string:

```ruby
describe DeliusImportJob do
end
```

Ideally we'd be able to pass a string to `describe`/`feature`, since this test exercises a much larger area of the codebase than just the `DeliusImportJob` class. It feels wrong as-is, but it's the only way we could get the stubbed `Net::IMAP` class to work.

@kjdchapman We're hoping you might know why this is happening 🤞😬

kjdchapman: I'm not sure quite why this is happening, but as some extra information that may help...

`describe Class` is "call the #describe method and pass it the Class constant as the first parameter", under the hood. This allows RSpec to do a few clever things - mainly it provides an [implicitly described subject](https://relishapp.com/rspec/rspec-core/v/3-5/docs/subject/implicitly-defined-subject).

I'm surprised this is interacting with your test badly, since it _should_ just be setting metadata. Some routes to investigate:
- Could it be something to do with file load order / imports, as it may change when the DeliusImportJob constant is imported?
- Can you replicate the problem in other contexts? Does a test around a non-job class cause the same problem (even if it stubs the `IMAP` class)

kjdchapman: [This StackOverflow answer](https://stackoverflow.com/questions/46472997/typeerror-superclass-mismatch-for-class-task) helped me to understand the root error (I think it's a clue that it's `IMAP` that's throwing it):

> TypeError: superclass mismatch happens when you reopen a class and try to redefine the superclass for it. This is a way to force this error:
> ```ruby
> class A; end
> class B < A; end
> class C; end
> class B < C; end # raises error
> ```

maysakanoni: It seems that this issue is resolved by requiring the delius import job directly in the spec file.

spec/features/delius_import_job_feature_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+
+    stub_const("Mail", MockMailMessage)
+
+    ENV['DELIUS_EMAIL_FOLDER'] = 'delius_import_feature'
+    ENV['DELIUS_XLSX_PASSWORD'] = 'secret'
+
+    create(:team, code: 'A')
+  end
+
+  it "imports the Delius spreadsheet and creates case information" do
```
kjdchapman, line 55: Is there a way to flip this test around so it checks for presence rather than absence of content?

I've found when I use `not_to have_content` in my assertions, they'll sometimes misleadingly pass - if I hit an error page, for example.

maysakanoni: Yeah, were now testing that its on the correct page and that the offender appears in the make allocations page.

PR #1136, opened 2020-04-29
POM-657 Expose new release date fields

Katy600:
[https://dsdmoj.atlassian.net/jira/software/projects/POM/boards/336?selectedIssue=POM-657](url)

**This adds 2 new date fields to the Prisoner Profile page:** 

- Post recall release date (DD/MM/YYYY)
- Sentence/ license end date (DD/MM/YYYY)

Each of these fields needs to display the most relevant date for the POM.

**For post recall release date this could be 3 potential dates:** 

1. post_recall_release_override_date
2. post_recall_release_date
3. actual_parole_date
image

**For Sentence/ license end date**
The meaningful date is which ever comes first.


app/models/nomis/sentence_detail.rb
```diff
@@ -27,6 +29,14 @@ def conditional_release_date
...
     end
 
+    def recall_release_date
+      if @actual_parole_date.presence && @actual_parole_date < @post_recall_release_override_date
+        @actual_parole_date
+      else
+        @post_recall_release_override_date.presence || @post_recall_release_date
+      end
+    end
+
```
kjdchapman, line 39: I found this conditional a little hard to read. I played with it a little locally to try and understand why.

- I think checking if the actual parole date's presence works better as a guard clause
- That shape forces you to repeat the `else` branch comparison, which made me want in a separate branch

I ended up here, if any of it's useful:
```ruby
def recall_release_date
  return post_recall_release_date if @actual_parole_date.blank?

  if @post_recall_release_override_date > @actual_parole_date
    @actual_parole_date
  else
    post_recall_release_date
  end
end

# is there a better name for this method?
def post_recall_release_date
  @post_recall_release_override_date.presence || @post_recall_release_date
end

```

app/models/nomis/sentence_detail.rb
```diff
@@ -27,6 +29,14 @@ def conditional_release_date
       @conditional_release_override_date.presence || @conditional_release_date
     end
 
+    def recall_release_date
+      if @actual_parole_date.presence && @actual_parole_date < @post_recall_release_override_date
```
kjdchapman, line 33: I think this comparison raises an error if `@actual_parole_date` is present but `@post_recall_release_override_date` is nil. 

It'd be good to test that case, and modify this to handle it gracefully.

app/models/nomis/offender_base.rb
```diff
@@ -1,11 +1,12 @@
 module Nomis
   class OffenderBase
     delegate :home_detention_curfew_eligibility_date,
-             :post_recall_release_date, :post_recall_release_override_date,
+             :post_recall_release_date,
```
kjdchapman, line 4: Is removing this override date intentional? I notice we refer to it in some of the stuff added later?

app/models/nomis/sentence_detail.rb
kjdchapman: (although see my other comment - is `override_date` still meant to be part of this component?)

spec/models/nomis/sentence_detail_spec.rb
```diff
@@ -1,53 +1,117 @@
...
       end
 
-      it 'uses original' do
+      it "uses original" do
         expect(subject.conditional_release_date).to eq(date)
       end
     end
   end
+
+  describe "#recall_release_date" do
```
kjdchapman, line 54: I like what these tests cover. I think using more cascading contexts and lets could make them easier to read, though.

I played with this a little to try and demonstrate what I mean. Happy to pair on it (we don't need to make changes in this PR, necessarily!):
```ruby
describe "#recall_release_date" do
  before do
    subject.actual_parole_date = parole_date
    subject.post_recall_release_date = post_recall_release_date
    subject.post_recall_release_override_date = override_date
  end

  context 'when there are no dates' do
    let(:parole_date) { nil }
    let(:post_recall_release_date) { nil }
    let(:override_date) { nil }

    it 'returns nil' do
      expect(subject.recall_release_date).to be_nil
    end
  end

  context 'when there is no post-recall release date' do
    let(:post_recall_release_date) { nil }

    context 'and a parole date' do
      let(:parole_date) { 12.weeks.from_now }

      context 'and no post-recall override date' do
        let(:override_date) { nil }

        it 'returns the parole date' do
          expect(subject.recall_release_date).to eq(parole_date)
        end
      end

      context 'and a later post-recall override date' do
        let(:override_date) { parole_date + 4.weeks }

        it 'returns the parole date' do
          expect(subject.recall_release_date).to eq(parole_date)
        end
      end

      context 'and an earlier post-recall override date' do
        let(:override_date) { parole_date - 4.weeks }

        it 'returns the override date' do
          expect(subject.recall_release_date).to eq(override_date)
        end
      end
    end
  end
end
```

kjdchapman: (note that first test fails, exposing an error I've talked about in another comment)

app/models/nomis/sentence_detail.rb
Katy600: Yes this is so much clearer.

app/models/nomis/offender_base.rb
Katy600: Oh dear. I think I thought I had introduced post_recall_release_override_date from the Swagger Nomis API. When I created the method recall_release_date I remember adding actual_parole_date to the delegate list in OffenderBase without really understanding what the delegate did. And then I realised that I didn't need it there if it was only going to be used in the new method, recall_release_date, in the SentenceDetail class, so I removed it. I also moved it out of attr_accessor to attr_writer in SentenceDetail, as I thought it only needed to be accessible to the SentenceDetail class itself, as it was only being used in my method recall_release_date. I think because post_recall_release_override_date was also part of my recall_release_date method, I did the same for it - I thought that I had introduced it to the code! 
I have just searched post_recall_release_override_date in the code base and it is only being used by me though! That is perhaps what motivated me to remove it from OffenderBase!

app/models/nomis/sentence_detail.rb
```diff
@@ -27,6 +30,20 @@ def conditional_release_date
       @conditional_release_override_date.presence || @conditional_release_date
     end
 
+    def recall_release_date
```
ollietreend, line 33: Are we right to be calling this the 'recall release date'? I note that we're reading from fields '**post** recall release date'. And in the view, we present this as 'prisoner recall release date'.

I'm wondering whether both names ('post recall release date', and 'prisoner recall release date') are actually both expansions of acronym 'PRRD'..? 🤔 I'm wondering whether the name 'recall release date' is a truthful description of what this field actually represents, or if it adds to the existing mix of names.

Naming things is hard! 🙈

app/models/nomis/sentence_detail.rb
```diff
@@ -27,6 +30,20 @@ def conditional_release_date
...
+      return decide_which_recall_release_date if @actual_parole_date.blank?
+      return @actual_parole_date if recall_release_override_date_is_earlier?
+      decide_which_recall_release_date
+    end
+  
+    def recall_release_override_date_is_earlier?
+      @post_recall_release_override_date.blank? || @post_recall_release_override_date > @actual_parole_date
+    end
+
+    def decide_which_recall_release_date
```
ollietreend, line 43: I have a similar thought here:

I'm wondering whether this is a descriptive name of what the method gives us. It's true that this method does *decide* which value to return, but ultimately what the method gives us is the 'true' post recall release date (either the existing one, or the overridden one).

I'm wondering whether a more descriptive name could be something like `actual_post_recall_release_date`, or `true_post_recall_release_date` 🤔

Again: naming things is hard!

app/models/nomis/sentence_detail.rb
```diff
@@ -27,6 +30,20 @@ def conditional_release_date
...
     end
 
+    def recall_release_date
+      return decide_which_recall_release_date if @actual_parole_date.blank?
+      return @actual_parole_date if recall_release_override_date_is_earlier?
+      decide_which_recall_release_date
+    end
+  
+    def recall_release_override_date_is_earlier?
+      @post_recall_release_override_date.blank? || @post_recall_release_override_date > @actual_parole_date
```
ollietreend, line 40: I think we're missing a check against – 'is post recall release date earlier than actual parole date'?

PR #1137, opened 2020-04-29
Enforce single quotes in both linting rulesets

kjdchapman:
As a followup to b7a7b9, this commit modifies the linting the git hook uses slightly. The base is still the default GOV.UK rubocop config, but with an added rule to enforce double-quoting.

This quoting rule is what's in the rubocop ruleset that is checked in the build pipeline. This change means that the build's rubocop and the git hook's rubocop won't ask us to do two conflicting things!

I had hoped that generally, the GOV.UK ruleset would be a superset of our existing custom ruleset, so this wouldn't happen often. If it does happen many times, this may not be the right strategy.


kjdchapman:
Thanks @ollietreend for catching I'd mixed up single/double quotes in PR title/commit! :facepalm:

PR #1142, opened 2020-05-01
POM-685: Update ‘Active’ Team → LDU relationships in the nightly Delius import

ollietreend:
Jira ticket: [POM-685](https://dsdmoj.atlassian.net/browse/POM-685)
Co-authored-by: @maysakanoni

---

The Delius import spreadsheet is the 'source of truth' for the names of active teams and their associations to LDUs. This PR updates the Delius import job to automatically update the names and associations between Teams and LDUs in our database to reflect those received in the Delius extract.

We've created a new service object, `UpdateTeamNameAndLduService`, which updates the team names and associations to LDUs in our database. It will not create new Team or LDU records in our database.

We've also added to the `DeliusImportJob`, which is the job that runs nightly to perform the import. The change involves performing a second pass over the Delius import spreadsheet:

1. **NEW**: Update ‘Active’ Team → LDU relationships
2. Update `DeliusData` to trigger updates on `CaseInformation` records

Since we're now performing two passes over the spreadsheet, we took the opportunity to DRY-up the looping functionality so that we don't end up needing to repeat that. This looping functionality is now handled by the new `for_each_record` method in the `DeliusImportJob` class.

### DIFF
--- a/app/jobs/delius_import_job.rb
+++ b/app/jobs/delius_import_job.rb
@@ -110,19 +110,7 @@ private
   end
 
   def update_team_names_and_ldus(processor)
-    processor.each_with_index do |row, index|
-      # skip header row in row[0]
-      next if index == 0
-
-      record = {}
-
-      # For each row, map the column to the appropriate column name
-      # as the existing column names are not very hash/symbol friendly
-      row.each_with_index do |val, idx|
-        key = FIELDS[idx]
-        record[key] = val
-      end
-
+    for_each_record(processor) do |record|
       UpdateTeamNameAndLduService.update(
         team_code: record[:team_code],
         team_name: record[:team],
@@ -133,6 +121,18 @@ private
 
   def upsert_delius_data_records(processor)
     total = 0
+
+    for_each_record(processor) do |record|
+      if record[:noms_no].present?
+        DeliusDataService.upsert(record)
+        total += 1
+      end
+    end
+
+    Rails.logger.info("[DELIUS] #{total} records attempted upsert")
+  end
+
+  def for_each_record(processor)
     processor.each_with_index do |row, index|
       # skip header row in row[0]
       next if index == 0
@@ -146,12 +146,7 @@ private
         record[key] = val
       end
 
-      if record[:noms_no].present?
-        DeliusDataService.upsert(record)
-        total += 1
-      end
+      yield(record)
     end
-
-    Rails.logger.info("[DELIUS] #{total} records attempted upsert")
   end
 end
--- a/app/jobs/delius_import_job.rb
+++ b/app/jobs/delius_import_job.rb
@@ -104,9 +104,35 @@ private
 
   def process_decrypted_file(filename)
     Rails.logger.info('[DELIUS] Processing decrypted file')
-
-    total = 0
     processor = Delius::Processor.new(filename)
+    update_team_names_and_ldus(processor)
+    upsert_delius_data_records(processor)
+  end
+
+  def update_team_names_and_ldus(processor)
+    processor.each_with_index do |row, index|
+      # skip header row in row[0]
+      next if index == 0
+
+      record = {}
+
+      # For each row, map the column to the appropriate column name
+      # as the existing column names are not very hash/symbol friendly
+      row.each_with_index do |val, idx|
+        key = FIELDS[idx]
+        record[key] = val
+      end
+
+      UpdateTeamNameAndLduService.update(
+        team_code: record[:team_code],
+        team_name: record[:team],
+        ldu_code: record[:ldu_code]
+      )
+    end
+  end
+
+  def upsert_delius_data_records(processor)
+    total = 0
     processor.each_with_index do |row, index|
       # skip header row in row[0]
       next if index == 0
--- a/app/services/update_team_name_and_ldu_service.rb
+++ b/app/services/update_team_name_and_ldu_service.rb
@@ -1,0 +1,36 @@
+class UpdateTeamNameAndLduService
+  def self.update(team_code:, team_name:, ldu_code:)
+    new(team_code: team_code, team_name: team_name, ldu_code: ldu_code).update
+  end
+
+  def initialize(team_code:, team_name:, ldu_code:)
+    @team = Team.find_by(code: team_code)
+    @team_name = team_name
+    @ldu_code = ldu_code
+  end
+
+  def update
+    return unless name_needs_updating? || ldu_needs_updating?
+
+    if name_needs_updating?
+      @team.name = @team_name
+    end
+
+    if ldu_needs_updating?
+      ldu = LocalDivisionalUnit.find_by(code: @ldu_code)
+      @team.local_divisional_unit = ldu
+    end
+
+    @team.save
+  end
+
+private
+
+  def name_needs_updating?
+    @team.name != @team_name
+  end
+
+  def ldu_needs_updating?
+    @team.local_divisional_unit&.code != @ldu_code
+  end
+end
--- a/app/services/update_team_name_and_ldu_service.rb
+++ b/app/services/update_team_name_and_ldu_service.rb
@@ -14,19 +14,9 @@ class UpdateTeamNameAndLduService
   def update
     return if @team.nil?
 
-    return unless name_needs_updating? || ldu_needs_updating?
-
-    if name_needs_updating?
-      @team.name = @team_name
-    end
-
-    if ldu_needs_updating?
-      ldu = LocalDivisionalUnit.find_by(code: @ldu_code)
-      @team.local_divisional_unit = ldu
-      return Rails.logger.error("Couldn't find LDU with code #{@ldu_code}") if ldu.nil?
-    end
-
-    @team.save
+    update_name if name_needs_updating?
+    update_ldu if ldu_needs_updating?
+    @team.save if @team.changed?
   end
 
 private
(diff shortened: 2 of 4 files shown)


app/services/update_team_name_and_ldu_service.rb
```diff
@@ -0,0 +1,42 @@
...
+
+    Rails.logger.error("Couldn't find team with code #{team_code}") if @team.nil?
+  end
+
+  def update
+    return if @team.nil?
+
+    update_name if name_needs_updating?
+    update_ldu if ldu_needs_updating?
+    @team.save if @team.changed?
```
kjdchapman, line 19: Are you checking the name in this way because you're concerned about the extra time of updating records that haven't changed? Or because you wanted the same pattern for name / ldu?

The ActiveRecord dirty-tracker is pretty clever, you can safely update that attributes and `.changed?` will give you what you want, I think. I think it may be clever enough for the relation as well, but that may be a bit nicer long-form.

info.tier
 => "A" 
info.tier = "B"
 => "B" 
info.changed?
 => true 
info.tier = "A"
 => "A" 
info.changed?
 => false 
```

File: app/services/update_team_name_and_ldu_service.rb [hunk 2]
```diff
@@ -0,0 +1,42 @@
...
+  def update_name
+    @team.name = @team_name
+  end
+
+  def ldu_needs_updating?
+    @team.local_divisional_unit&.code != @ldu_code
+  end
+
+  def update_ldu
+    ldu = LocalDivisionalUnit.find_by(code: @ldu_code)
```
kjdchapman, line 37: What do you think about pulling this into the initializer? I think it might be clearer to have the "if any of this data looks suspect, just don't do anything" errors into one place.

I think that may also make the updating code clearer in turn, but I'd need to play with it a little.

spec/features/delius_import_job_feature_spec.rb
```diff
@@ -49,23 +49,52 @@
 
     ENV['DELIUS_EMAIL_FOLDER'] = 'delius_import_feature'
     ENV['DELIUS_XLSX_PASSWORD'] = 'secret'
+  end
+
+  context "when the team is associated with an LDU" do
+    before do
+      create(:team, code: 'A')
+    end
```
kjdchapman, line 57: nit: can one-line short before blocks for a little easier reading
```suggestion
    before { create(:team, code: 'A') }
```

spec/features/delius_import_job_feature_spec.rb
```diff
@@ -49,23 +49,52 @@
 
     ENV['DELIUS_EMAIL_FOLDER'] = 'delius_import_feature'
     ENV['DELIUS_XLSX_PASSWORD'] = 'secret'
+  end
+
+  context "when the team is associated with an LDU" do
```
kjdchapman, line 54: I love that we've got a test in this style now!

spec/features/delius_import_job_feature_spec.rb
```diff
@@ -49,23 +49,52 @@
...
+    it "imports the Delius spreadsheet and creates case information" do
+      visit prison_summary_pending_path(prison)
+      expect(page).to have_content("Add missing information")
+      expect(page).to have_content(offender_no)
 
-    create(:team, code: 'A')
+      DeliusImportJob.perform_now
+
+      reload_page
+      expect(page).to have_content("Add missing information")
```
kjdchapman, line 67: I think this is a good assertion to use here - it's general enough that it shouldn't be too brittle - but I'm wondering how to make it more obvious to a reader what we're checking. 

You need prior knowledge of other bits of the app to "get" this test, not sure how much of a problem that is. 

I also have no concrete suggestions for improving it :shrug: so not sure how useful this is!

kjdchapman: On further reflection, maybe it could be helped with more structure, using context/it blocks to describe what the different stages/assertions are?

spec/services/update_team_name_and_ldu_service_spec.rb
```diff
@@ -0,0 +1,78 @@
+require 'rails_helper'
+
+RSpec.describe UpdateTeamNameAndLduService do
+  context "when the team has the wrong name" do
```
kjdchapman, line 4: I think you can reduce the duplication by using `let` here.

I couldn't resist playing with this, here's where I got to. You definitely don't have to use this verbatim, just to explain my thoughts!

```ruby
describe UpdateTeamNameAndLduService do
  context "with an existing LDU and team" do
    before do
      create(:local_divisional_unit, code: 'LDU1')
      team = build(:team, code: team_code)
      team.local_divisional_unit = local_divisional_unit
      team.save(validate: false)
    end

    context "when the team is not associated with the LDU" do
      let(:team_code) { 'TEAM1'}
      let(:local_divisional_unit) { nil }

      it "repairs the association" do
        described_class.update(team_code: team_code, ldu_code: 'LDU1')
        team = Team.find_by(code: team_code)
        expect(team.local_divisional_unit.code).to eq('LDU1')
      end
    end

    context "when the team is associated with a different LDU" do
      let(:team_code) { 'TEAM1'}
      let(:local_divisional_unit) { create(:local_divisional_unit, code: 'other') }

      it "associates the team with the new LDU" do
        described_class.update(team_code: team_code, ldu_code: 'LDU1')
        team = Team.find_by(code: 'TEAM1')
        expect(team.local_divisional_unit.code).to eq('LDU1')
      end
    end

    context "when the team is already associated to the correct LDU" do
      let(:team_code) { 'TEAM1'}
      let(:local_divisional_unit) { create(:local_divisional_unit) }

      it "doesn't update the team record" do
        described_class.update(team_code: team_code, ldu_code: local_divisional_unit.code)
        expect_any_instance_of(Team).not_to receive(:save)
        described_class.update(team_code: 'TEAM1', ldu_code: 'LDU1')
      end
    end
  end
end
```

kjdchapman: (note that this doesn't cover all of the tests)

app/jobs/delius_import_job.rb
```diff
@@ -104,9 +104,35 @@ def process_attachment(attachment_bytes)
 
   def process_decrypted_file(filename)
     Rails.logger.info('[DELIUS] Processing decrypted file')
+    processor = Delius::Processor.new(filename)
```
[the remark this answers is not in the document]
ollietreend: @kjdchapman The `Delius::Processor` class opens a XLSX file and provides an `Enumerable` interface to loop through rows of the spreadsheet. So as for naming, a more descriptive name would probably be `XlsxReader` – it's not unique to Delius, so the module namespacing here is misleading.

It looks like one of the key considerations when originally implementing `Delius::Processor` in 2d07dc7f was that it should stream rows from the XLSX file, rather than attempt to load the entire file into memory. However this seems to be is something that established XLSX-reading gems can do, too ([roo](https://github.com/roo-rb/roo#excel-xlsx-and-xlsm-support), [creek](https://github.com/pythonicrubyist/creek)).

So If I were going for a full refactor, I'd replace this functionality with a gem – I'd rather us not have to be aware of the intricacies of reading XLSX files, nor have to maintain that as part of our codebase. But I didn't need to get too close to it for this PR, so have left it as is.

app/jobs/delius_import_job.rb
```diff
@@ -104,9 +104,35 @@ def process_attachment(attachment_bytes)
...
+    update_team_names_and_ldus(processor)
+    upsert_delius_data_records(processor)
+  end
+
+  def update_team_names_and_ldus(processor)
```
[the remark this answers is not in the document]
ollietreend: Thanks! 🙂

app/services/update_team_name_and_ldu_service.rb
```diff
@@ -0,0 +1,42 @@
+class UpdateTeamNameAndLduService
+  def self.update(team_code:, team_name:, ldu_code:)
```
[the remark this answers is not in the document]
ollietreend: @kjdchapman I like your separation of initialisation time and execution time concerns.

I do agree with your concerns about the approach I've taken becoming hard to understand over time – I was hesitant t use it.

The primary motive behind this approach was to strike a balance between:
- maintaining consistency with the rest of the application – most service objects seem to provide their interface via class methods (for example, the DeliusDataService, HandoverDateService, OffenderService, etc.)
- allowing the `UpdateTeamNameAndLduService` itself to be expressive in its implementation – avoiding cluttering things by having to pass parameters around to private methods within the service

So this allowed us not to 'break' the existing model of using services via class methods, but still to do nice things within the class. I'd prefer to leave this as-is for now, but I do appreciate your concerns about its long term usage, and I hope my reasoning seems sensible enough.

app/services/update_team_name_and_ldu_service.rb
ollietreend: My implementation prior to 8dc0a40 wasn't actually using the `.changed?` method – so the existence of `name_needs_updating?` is somewhat of a hang-over from that. Although now it's there, I do like the consistency of using the same pattern for both name and LDU.

It's very cool to see how smart the ActiveRecord dirty tracking is 🤓 I didn't know that!

app/services/update_team_name_and_ldu_service.rb
ollietreend: That's a fair point – I'd like the explicitness of having those 'find' errors in the same place.

The reasoning for this approach was to avoid performing (potentially) unnecessary database queries. In practice, I expect most teams will already have the correct LDU (that's the ideal scenario 😬🤞) so won't need updating – my intention was to avoid looking up the 'target' LDU until it's clear that we need to update the existing association.

Perhaps that's an over-optimisation at this stage 🤷 but I'd prefer to leave it as-is for now, so long as you're happy for me to do so.

spec/features/delius_import_job_feature_spec.rb
ollietreend: I think our feelings about the structure of this file might change once we've got more contexts in – coming shortly in [POM-686](https://dsdmoj.atlassian.net/browse/POM-686) is functionality to map shadow teams to active teams. It might be worth having this discussion again at that point.

PR #1146, opened 2020-05-05
POM-670 additional fields for earliest release date

starswan:
This change adds 4 new fields to the earliest_release_date calculation (if the 5 major ones are all missing). It uses the same logic as the other dates, except that this has been tweaked slightly so that if the only dates available are in the past, then the **latest** date is chosen rather than the oldest as it is more likely to be correct.

### DIFF
--- a/app/models/nomis/sentence_detail.rb
+++ b/app/models/nomis/sentence_detail.rb
@@ -10,7 +10,9 @@ module Nomis
                 :release_date,
                 :licence_expiry_date,
                 :sentence_start_date,
-                :tariff_date
+                :tariff_date,
+                :sentence_expiry_date,
+                :actual_parole_date
 
     def initialize(fields = {})
       fields.each do |k, v|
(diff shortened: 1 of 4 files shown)


app/models/nomis/sentence_detail.rb
```diff
@@ -8,16 +8,25 @@ class SentenceDetail
...
 
     attr_writer :automatic_release_date,
                 :automatic_release_override_date,
                 :conditional_release_date,
-                :conditional_release_override_date
+                :conditional_release_override_date,
+                :post_recall_release_date,
+                :post_recall_release_override_date
+
+    def initialize(fields = {})
```
kjdchapman, line 25: Would it be good to use this in the production code that initializes this object? ([on line 49 of this same file, below](https://github.com/ministryofjustice/offender-management-allocation-manager/pull/1146/files#diff-af383291a16f3cffa1cb19f273fdb104R49))

I think it's good to pick one strategy (I don't have strong opinions around custom initializers vs `#tap`)

starswan: I think the other PR changes this as you suggest - It was part of a refactoring, so I didn't include it here

PR #1149, opened 2020-05-06
POM-658 hide handover dates when COM responsible 

starswan:
all handover data is now N/A if the offender has already been handed over to the community

### DIFF
--- a/app/models/nomis/offender_base.rb
+++ b/app/models/nomis/offender_base.rb
@@ -84,7 +84,7 @@ module Nomis
     end
 
     def pom_responsibility
-      ResponsibilityService.calculate_pom_responsibility(self)
+      @pom_responsibility ||= ResponsibilityService.calculate_pom_responsibility(self)
     end
 
     def sentence_start_date
@@ -163,7 +163,11 @@ module Nomis
   private
 
     def handover
-      @handover ||= HandoverDateService.handover(self)
+      @handover ||= if pom_responsibility&.custody?
+                      HandoverDateService.handover(self)
+                    else
+                      HandoverDateService::NO_HANDOVER_DATE
+                    end
     end
   end
 end
--- a/app/services/responsibility_service.rb
+++ b/app/services/responsibility_service.rb
@@ -182,23 +182,20 @@ private
 
   class CrcRules
     def self.responsibility(offender)
-      if release_date_gt_12_weeks?(offender)
+      # CRC can look at HDC date, NPS is not supposed to
+      earliest_release_date =
+        offender.home_detention_curfew_actual_date.presence ||
+          [offender.automatic_release_date,
+           offender.conditional_release_date,
+           offender.home_detention_curfew_eligibility_date].compact.min
+
+      return nil if earliest_release_date.nil?
+
+      if earliest_release_date > DateTime.now.utc.to_date + 12.weeks
         RESPONSIBLE
       else
         SUPPORTING
       end
-    end
-
-  private
-
-    # CRC can look at HDC date, NPS is not supposed to
-    def self.release_date_gt_12_weeks?(offender)
-      earliest_release_date =
-        offender.home_detention_curfew_actual_date.presence ||
-            [offender.automatic_release_date,
-             offender.conditional_release_date,
-             offender.home_detention_curfew_eligibility_date].compact.min
-      earliest_release_date > DateTime.now.utc.to_date + 12.weeks
     end
   end
 
--- a/app/services/responsibility_service.rb
+++ b/app/services/responsibility_service.rb
@@ -35,7 +35,7 @@ class ResponsibilityService
 private
 
   def self.standard_rules(offender)
-    if nps_case?(offender) || offender.indeterminate_sentence?
+    if offender.nps_case? || offender.indeterminate_sentence?
       nps_rules(offender)
     else
       crc_rules(offender)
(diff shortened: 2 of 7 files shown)


spec/factories/offenders.rb
```diff
@@ -0,0 +1,14 @@
...
+    inprisonment_status do 'SENT03' end
+    sentence do
+      Nomis::SentenceDetail.new.tap { |s|
+        s.automatic_release_date = Time.zone.today + 1.year
+        s.sentence_start_date = Time.zone.today
+      }
+    end
+    prison_id do 'LEI' end
+    case_allocation do 'NPS' end
+    mappa_level { 0 }
```
ollietreend, line 12: Could these block declarations be made consistent? I think the preferred approach is `{...}` curly braces for single line blocks, and `do...end` for multiline.

```suggestion
    prison_id { 'LEI' }
    case_allocation { 'NPS' }
    mappa_level { 0 }
```

spec/factories/offenders.rb
```diff
@@ -0,0 +1,14 @@
+FactoryBot.define do
+  factory :offender, class: 'Nomis::Offender' do
+    inprisonment_status do 'SENT03' end
+    sentence do
+      Nomis::SentenceDetail.new.tap { |s|
```
ollietreend, line 5: I note that this `.tap` multiline block uses `{...}` curly braces style, although I've seen this same style used elsewhere specifically for `.tap` blocks.

It looks like both our rubocop configs (the main `.rubocop.yml` and the aspirational `.rubocop-govuk.yml`) don't seem to have an opinion on `do...end` vs `{...}` – so maybe this is a non-issue.

app/services/responsibility_service.rb
```diff
@@ -182,24 +182,21 @@ def self.open_prison_nps_offender?(offender)
...
     def self.responsibility(offender)
-      if release_date_gt_12_weeks?(offender)
+      # CRC can look at HDC date, NPS is not supposed to
+      earliest_release_date =
+        offender.home_detention_curfew_actual_date.presence ||
+          [offender.automatic_release_date,
+           offender.conditional_release_date,
+           offender.home_detention_curfew_eligibility_date].compact.min
+
+      return nil if earliest_release_date.nil?
```
ollietreend, line 192: In the case that we return `nil`, what does this ultimately mean – is it that the POM is not responsible because the COM is?

I wonder whether it would be clearer/more descriptive to have this case as another class constant, alongside the `RESPONSIBLE`, `SUPPORTING`, `COWORKING` and `NPS` constants defined on lines 10-13.

If this represents what I think it does, then I guess it could be a constant called `NOT_RESPONSIBLE` or `COM_RESPONSIBLE`?

I also wonder whether the return value itself should/could be a new `Responsiblity` object with a `:custody?` value of `false`, rather than the implicit `false` garnered from the `nil` response.

I think this could make this code a little more explicit in what it's trying to calculate. What do you think @starswan?

spec/models/nomis/offender_spec.rb
```diff
@@ -0,0 +1,21 @@
+require 'rails_helper'
```
ollietreend, line 1: On the subject of tests – it might be nice to also cover this in a feature test somewhere it ends up being shown to users. Hopefully there's already a feature test which could be added to.

What do you think @starswan?

app/models/nomis/offender_base.rb
```diff
@@ -167,7 +167,11 @@ def load_case_information(record)
   private
 
     def handover
-      @handover ||= HandoverDateService.handover(self)
+      @handover ||= if pom_responsibility&.custody?
+                      HandoverDateService.handover(self)
+                    else
+                      HandoverDateService::NO_HANDOVER_DATE
+                    end
```
ollietreend, line 174: Is there a reason why we don't make this `if/else` statement the responsibility of the `HandoverDateService`? Is the intention to avoid causing a dependency between the two services (i.e. the call to `pom_responsibility&.custody?`)? If so, that makes sense.

spec/factories/offenders.rb
starswan: I think rubocop -a is doing this - agree with your suggestion above

app/services/responsibility_service.rb
starswan: This module got changed at some point to allow a return of nil for 'dont know' which looks like a receipe for disaster (I think the null object pattern might have been a better choice). I didn't address this challenge in this PR.

app/models/nomis/offender_base.rb
starswan: Yes. It does feel like we have one service in 2 bits here - responsibility and handover are clesrly 2 sides of the same coin (but that wasn't clear for quite some time). It would probably make sense to put them closer together somehow (to reflect their cohesive nature)

spec/models/nomis/offender_spec.rb
starswan: I didn't feel it was needed in this case, as all this does is change the scenarios where 'N/A' is shown to the user. Our current tests are dominated by feature tests which is less than ideal, as often they break in unfathomable ways which makes testing/debugging much harder than it needs to be.

app/services/responsibility_service.rb
```diff
@@ -182,24 +182,21 @@ def self.open_prison_nps_offender?(offender)
...
+
+      if earliest_release_date > DateTime.now.utc.to_date + 12.weeks
```
ollietreend, line 194: Can we split this refactor into its own commit? As discussed, I don't think it's related to the primary change in this commit – i.e. displaying N/A for handover dates where the COM is already responsible

PR #1153, opened 2020-05-13
Add rake task for seeing missing handover emails

kjdchapman:
This PR includes a rake task, which can be run with:
`rake handover_email_summary_by_prison [prison_code]`

Notably, this rake task cannot give information across the entire estate, only for an individual prison. I expect it's possible to get a view across the state in one task with some fault-tolerance / batch-processing work.

Also, this rake task outputs the raw counts for each category, not any IDs to investigate. It should be easy to change this in the service if more detail is needed.

This PR is draft, because these counts are not battle-tested yet. I expect there is at least one problem - I haven't yet seen a way to differentiate between:
- This record is not marked as 'NPS' because it doesn't have the
relevant Delius record stored against it yet.
- This record is genuinely not marked as 'NPS'.

### DIFF
--- a/app/services/view_handover_email_addresses_summary.rb
+++ b/app/services/view_handover_email_addresses_summary.rb
@@ -10,13 +10,19 @@ class ViewHandoverEmailAddressesSummary
   def offender_category(offender)
     case_info = CaseInformation.find_by(nomis_offender_id: offender.offender_no)
 
-    return :missing_delius_record if missing_delius_record?(case_info)
-    return :missing_team_link if missing_team_link?(case_info)
-    return :missing_team_information if missing_team_information?(case_info)
-    return :missing_local_delivery_unit if missing_local_delivery_unit?(case_info)
-    return :missing_local_delivery_unit_email if missing_email?(case_info)
-
-    :has_email_address
+    if missing_delius_record?(case_info)
+      :missing_delius_record
+    elsif missing_team_link?(case_info)
+      :missing_team_link
+    elsif missing_team_information?(case_info)
+      :missing_team_information
+    elsif missing_local_delivery_unit?(case_info)
+      :missing_local_delivery_unit
+    elsif missing_email?(case_info)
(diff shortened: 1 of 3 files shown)


ollietreend:
@starswan I think the original intention was for this service to be used to build a HTML dashboard view for users – but that didn't work due to page load speed concerns. (Lots of slow queries to build the output.)

In practice, this service will probably also be useful within the context of a rails console – i.e. where we could use the service directly in the console rather than dropping out to bash and using it via the rake task. So I'm not sure about putting it in a `services/rake_tasks` directory – I guess that would be a strong indicator that the code really belongs in the rake task itself.

But I do agree with your concerns about the services directory becoming an uncategorised dumping ground of 'miscellaneous' things.

Do you think it could go into a `services/reports` directory instead? I'm not sure what the appetite is for other services which also produce report-like outputs – so I guess the risk is it becomes a directory with just one service in. But `reports` feels like a more descriptive category for this service than `rake_tasks`.

starswan:
> @starswan I think the original intention was for this service to be used to build a HTML dashboard view for users – but that didn't work due to page load speed concerns. (Lots of slow queries to build the output.)
> 
> In practice, this service will probably also be useful within the context of a rails console – i.e. where we could use the service directly in the console rather than dropping out to bash and using it via the rake task. So I'm not sure about putting it in a `services/rake_tasks` directory – I guess that would be a strong indicator that the code really belongs in the rake task itself.
> 
> But I do agree with your concerns about the services directory becoming an uncategorised dumping ground of 'miscellaneous' things.
> 
> Do you think it could go into a `services/reports` directory instead? I'm not sure what the appetite is for other services which also produce report-like outputs – so I guess the risk is it becomes a directory with just one service in. But `reports` feels like a more descriptive category for this service than `rake_tasks`.

I think just a comment would suffice for now - at least that way it will be explicitly obvious what this code is for, rather than trying to categorise it.

PR #1154, opened 2020-05-14
Update Team Shadow code from Delius Import

maysakanoni:
Currently, there are lots of offender import failures as a result of not being able to find the Team and corresponding Ldu for an Offender. This pr contains one of the fixes which may reduce the number import failures. It aims to read through the Delius import row by row, check for shadow records (which can be identified , read [commit message](1dd7e4e7663cc9f8c9d29adea0c035c66ff1ae65)) and update the shadow codes which correspond to a given Team. When the service later attempts to `process` this offender, or create a `CaseInformation`, it will attempt to find its Team by looking at its code, and for shadow records, this will be the shadow code which was just updated. Therefore, it will be able to match to a Team and therefore an Ldu.


app/services/update_shadow_team_association_service.rb
```diff
@@ -0,0 +1,37 @@
...
+    retrieve_team if @team_name.match('(?i)OMIC.').present?
+
+    return Rails.logger.error("Couldn't find team with name #{@team_name}") if @team.nil?
+
+    update_shadow_code if incorrect_team_shadow_code
+  end
+
+private
+
+  def retrieve_team
```
maysakanoni, line 22: This can probably be allot cleaner

app/jobs/delius_import_job.rb
```diff
@@ -149,4 +163,14 @@ def for_each_record(processor)
...
   end
+
+  def active_record?(ldu_code)
+    return false if ldu_code.match('N\d{2}OMIC').present?
+
+    true
+  end
+
+  def shadow_record?(ldu_code)
+    ldu_code.match('N\d{2}OMIC').present?
```
ollietreend, line 174: I think we could improve the readability of this line by using the [`.match?` method](https://ruby-doc.org/core-2.6.3/Regexp.html#method-i-match-3F) in place of `.match().presence?`:

```suggestion
    ldu_code.match?('N\d{2}OMIC')
```

The [`match` method](https://ruby-doc.org/core-2.6.3/Regexp.html#method-i-match) returns a `MatchData` object which gives us the results of the regex match. This is useful, for example, when we want to extract something from the original string. But in our case, we're not interested in extracting anything from the original string – we just want to know whether it matches or not.

So for that, we can use the `match?` method. This will simply give us a boolean `true` or `false` response, rather than a `MatchData` object that we then need to inspect to see if it has 'presence' or if it's empty.

Both approaches ultimately achieve the same thing, so this is mostly just a matter of preference. No need to change if you're happy as it is.

app/services/update_shadow_team_association_service.rb
ollietreend: @maysakanoni This does what we need it to do – so that's the first hurdle.

I think a regular expression could help reduce the number of lines and code repetition. This is an example where we'd want to extract data from the original string using a regular expression (using the `match` method) rather than just test whether it matches (with `.match?`).

I'm not sure how familiar you are with regular expressions, but thought I'd share this in case it's useful. I like using a tool called [RegExr](https://regexr.com/) when writing regular expressions – it has built in documentation, and an 'explain' feature which will describe exactly what the regular expression is doing and why.

I've put together an example of a regex that could be useful to when trying to extract an 'active' team name from the 'shadow' team. Have a look if you're interested: https://regexr.com/54kbn

In RegExr, the blue lines are the ones which have matched against the regular expression. If you hover over each highlighted line, you'll see a `Group #1` value – this is what we'd want to _extract_ from the original string using Ruby's `match` method.

Also note that this is a case-insensitive match, so will also be able to extract an active team name from a really weird shadow team (for example: `oMiC - Something` or `OmIc - Something`)

ollietreend:
> I've approved this PR although there's one minor issue that I've noticed and that is in UpdateShadowTeamAssociationService, line 26. We don’t need `if match.nil?` after `Rails.logger.error` as we are already in an if match.nil? statement. I didn't want to block this PR though for such a minor thing.

Great spot @theKHutDeveloper – thank you! 😄 I'll that fix before merging it in.

PR #1158, opened 2020-05-25
convert SentenceDetail#post-recall-override-date into a hidden property

starswan:
This change matches the others in SentenceDetail. The presence of an 'OverrideDate' in some API fields should be encapsulated within the API-focussed SentenceDetail class, and should not leak into the application. This has already been done for automatic_release and conditional_release dates.

### DIFF
--- a/app/models/nomis/sentence_detail.rb
+++ b/app/models/nomis/sentence_detail.rb
@@ -8,8 +8,6 @@ module Nomis
                   :home_detention_curfew_eligibility_date,
                   :home_detention_curfew_actual_date,
                   :parole_eligibility_date,
-                  :post_recall_release_date,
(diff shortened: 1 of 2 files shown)


starswan:
> I'm rejecting this purely on the grounds that @Katy600 has implemented effectively the same thing in #1136. That PR is almost ready to go, so would rather not introduce a change which would have to be rebased against.
> 
> It turns out `post_recall_release_date` is not only driven by the override field, but also by the existence of an `actual_parole_date` value – so Katy's PR addresses that, as well as displaying that date to the end user.

My intent here is to remove the block between Katy and my stories that has held everything up since I was away. My story is ready to go, and I didn't want create a dependency just because of this trivial change

PR #1160, opened 2020-05-26
Govuk Rubocop config

starswan:
The current multiple rubocop configs are a bit confusing. Although the spirit was well-intentioned, the current setup results in rubocop noise in almost every PR (especially refactorings) making it hard to follow the actual work involved. 
This change brings the GovUk rubocop style into the project centre-stage, with our local exceptions intact. Some of these could be removed if they don't make sense, but it seems sensible to use judgement rather than slavishly agreeing with all the 'standards' in the GovUk set - many of these are just opinions.


ollietreend:
Our rubocop config file is getting quite long. It looks like there are some rules in which the GOV.UK rubocop config agrees with us – so wondering if we could use this as an opportunity to cut back our own config by reducing duplicate rule definitions.

For example, a lot of the `Metrics/` rules are disabled altogether by the GOV.UK rules: https://github.com/alphagov/rubocop-govuk/blob/master/config/metrics.yml

In our config file, we override the (previous) defaults to make them more lenient, but didn't go so far as to disable them. At this point, is it worth just accepting the GOV.UK rules here and allowing those rules to be disabled? If so, it looks like we could safely lose all our `Metrics/` rules in favour of those from GOV.UK.

The same goes for `Layout/AccessModifierIndentation` which is configured the same as us in the GOV.UK rules. So we could remove it from ours and just use the definition from GOV.UK.

I'm sure there will be other cross-over between us and GOV.UK – although it could take a while to go through every rule and check! But it'd be a shame not to get the low hanging fruit and trim our rules a little, rather than just importing more on top of what we already have.

PR #1162, opened 2020-05-28
Add offender counts to teams page in Active Admin

ollietreend:
Jira ticket: [POM-717](https://dsdmoj.atlassian.net/browse/POM-717)

---

This pull request adds an 'offenders' column to the 'Teams' page in Active Admin:

localhost_3000_admin_teams_order=id_desc(iPad)

This column shows the number of offenders ('case information' records) assigned to each team.

This is an enabler for us cleaning up our team data, as it'll allow us to see which teams are in use, and which ones are redundant.

### Database changes

I've used a [counter cache](https://guides.rubyonrails.org/association_basics.html#options-for-belongs-to-counter-cache) to maintain a count of the number of offenders in each team. This adds a new `case_information_count` field to the `teams` database.

A database migration will create this new column and seed it for existing team records.

### Other changes

I've also fixed a couple of small but related things:

- **Active Admin:** The "Local Divisional Units" dropdown/select list in the "Filter" sidebar panel on the Teams page is now sorted alphabetically. Previously this was unsorted, so was very difficult to find the desired LDU if using this filter.
- **Database index:** I've added a database index on the `teams.shadow_code` column. We frequently find teams by their shadow code in our application, so we should really have an index for that read pattern.
- **Code coverage:** I've configured Simplecov to ignore files in the `app/admin/` directory, as testing the functionality of our Active Admin is not a priority. We do, however, still have a feature test to cover user access permissions in [`spec/features/active_admin_feature_spec.rb`](https://github.com/ministryofjustice/offender-management-allocation-manager/blob/master/spec/features/active_admin_feature_spec.rb)

### DIFF
--- a/app/admin/teams.rb
+++ b/app/admin/teams.rb
@@ -13,6 +13,8 @@ ActiveAdmin.register Team do
   #   permitted << :other if params[:action] == 'create' && current_user.admin?
   #   permitted
   # end
+
+  # Columns for index page
   index do
     selectable_column
     id_column
@@ -27,4 +29,11 @@ ActiveAdmin.register Team do
     column 'Updated at', :updated_at
     actions
   end
+
+  # Filter fields
+  filter :name
+  filter :code
+  filter :shadow_code
+  filter :local_divisional_unit, collection: proc { LocalDivisionalUnit.order(:name) }
+  filter :case_information_count, label: 'Offender count'
 end
--- a/app/admin/teams.rb
+++ b/app/admin/teams.rb
@@ -13,4 +13,18 @@ ActiveAdmin.register Team do
   #   permitted << :other if params[:action] == 'create' && current_user.admin?
   #   permitted
   # end
+  index do
+    selectable_column
+    id_column
+    column :name
+    column :code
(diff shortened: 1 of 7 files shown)


app/admin/teams.rb
```diff
@@ -13,4 +13,27 @@
...
+  # Columns for index page
+  index do
+    selectable_column
+    id_column
+    column :name
+    column :code
+    column 'Shadow code', :shadow_code
+    column 'LDU', :local_divisional_unit
+    column 'Offenders', sortable: :case_information_count do |team|
+      team.case_information.size
```
starswan, line 26: I think this should be case_information_count, otherwise we will get an N+1 query?

ollietreend: @starswan somewhat counterintuitively, it seems that `.size` is the method you need to use to take advantage of the counter cache – `.count` performs an SQL count regardless of counter cache.

I was seeing N+1 queries in development, which is when I went back to [the Rails guide](https://guides.rubyonrails.org/association_basics.html#options-for-belongs-to-counter-cache) and noticed it said to use `.size` for counter cache.

PR #1171, opened 2020-06-02
Pom 511 delius manual update

theKHutDeveloper:
### Context:
When we have auto matching switched on for Delius we know that there will be some instances where we cannot make a match. So instead of unmatched cases being stuck in the missing information page, we will allow the SPO to manually enter the following information:

_1. whether the case will be managed in the Scottish or Northern Irish community. We know that these cases don’t exist in nDelius, so there is nothing to match against
2. the provider details (CRC or NPS)
3. the prisoner tier
4. the team name_

### Screenshots
- Initial view: SPO needs to identify whether the offender is Scottish, Welsh or from Northern Ireland
Screen Shot 2020-05-25 at 20 22 53

- **Scottish or Northern Irish offender**

The journey ends, SPO taken back to the 'Add missing information' page.
The offender will be ready to be allocated to a POM. The offender's case information is updated to reflect the following:

    welsh_offender: 'No'
    probation_service: 'Scotland' or 'Northern Ireland'
    tier: 'N/A'
    case_allocation: 'N/A'
    team: nil

- **English or Welsh offender**

**NOTE:** All cases that are not Scottish, Northern Irish or Welsh will be considered 'English' even if they are foreign nationals

- SPO taken to second page where they provide additional information
Screen Shot 2020-05-25 at 20 23 14

When the SPO provides this information, the journey ends and the offender's case information is updated to reflect the following:

    welsh_offender: 'Yes' (if probation_service is 'Wales'), otherwise 'No'
    probation_service: 'Wales' or 'England'
    tier: selected by SPO 
    case_allocation: selected by SPO
    team: selected by SPO

- An email gets sent to the LDU and the SPO in the background using a Notify template. The SPO will also get a flash notification in the app.

Screen Shot 2020-05-25 at 20 23 32

- If we know that the LDU or the SPO doesn't have an email address, then the SPO receives a flash alert in the app, advising them of any issues.

Screen Shot 2020-05-25 at 20 23 47

This PR also covers editing / updating the case information. For example if the SPO originally set the offender as Scottish but later changed this to English.


app/controllers/case_information_controller.rb
```diff
@@ -97,10 +110,121 @@ def nomis_offender_id_from_url
...
+  def send_email
+    return unless @case_info.probation_service == 'England' || @case_info.probation_service == 'Wales'
+
+    prepare_email
+  end
+
+  def prepare_email
+    spo = Nomis::Elite2::UserApi.user_details(current_user).email_address.try(:first)
+    ldu = @case_info.team.try(:local_divisional_unit)
+    emails = [ldu.try(:email_address), spo]
```
starswan, line 197: would be more idiomatic to use 'compact' or reject(&:blank?) to avoid the next on line 203

config/initializers/assets.rb
```diff
@@ -17,4 +17,4 @@
 # Precompile additional assets.
 # application.js, application.css, and all non-JS/CSS in the app/assets
 # folder are already added.
-# Rails.application.config.assets.precompile += %w( admin.js admin.css )
```
starswan, line 20: Don't think this change is needed. admin.js and admin.css are needed for ActiveAdmin stuff, the new JS file should already be in the asset pipeline due to its location

spec/features/active_admin_feature_spec.rb
```diff
@@ -31,7 +31,7 @@
   context 'when spo' do
     before do
       signin_spo_user
-      ci = create(:case_information, team: nil)
+      ci = create(:case_information, probation_service: 'Scotland', team: nil)
```
starswan, line 34: Is this change necessary?

spec/features/case_information_feature_spec.rb
```diff
@@ -1,34 +1,458 @@
...
+         vcr: { cassette_name: :case_information_update_location_validation_errors } do
+        signin_user
+        visit new_prison_case_information_path('LEI', nomis_offender_id)
+
+        choose_country(country: 'England')
+        click_button 'Continue'
+        second_page_form(case_alloc: 'crc', tier: 'd', team_option: 'NPS - England')
+        click_button 'Continue'
+
+        visit edit_prison_case_information_path('LEI', nomis_offender_id)
```
starswan, line 327: do we need to visit a new path here, or can we click a button to get there?

spec/features/case_information_feature_spec.rb
```diff
@@ -1,34 +1,458 @@
...
+      it 'does not send email when updating team has the same ldu', :raven_intercept_exception,
+         vcr: { cassette_name: :case_information_update_email_not_triggered_when_updated_team_has_same_ldu } do
+        choose_country(country: 'England')
+        click_button 'Continue'
+        second_page_form(case_alloc: 'nps', tier: 'b', team_option: 'NPS - England')
+        click_button 'Continue'
+
+        visit edit_prison_case_information_path('LEI', nomis_offender_id)
+
+        page.execute_script("document.getElementsByName('input-autocomplete')[0].value='NPS - England 3'")
```
starswan, line 393: As this is a feature spec, IMHO we should set js: true and then perform the actual clicks if possible.

spec/features/case_information_feature_spec.rb
```diff
@@ -1,34 +1,458 @@
...
+        end
+
+        expect(page).not_to have_css(".alert")
+        expectations(probation_service: 'England', tier: 'A', team: 'NPS - England 3', case_allocation: 'NPS')
+      end
+    end
   end
 
   it "clicking back link after viewing prisoner's case information, returns back the same paginated page",
+     :raven_intercept_exception,
```
starswan, line 455: not sure about these raven_intercept_exception tags. I think we should only use them very sparsely in the whole codebase if at all.

app/assets/javascripts/application.js
```diff
@@ -14,4 +14,5 @@
 //= require turbolinks
 //= require 'govuk/all'
 //= require 'moj/all'
+//= require 'accessible-autocomplete.min'
```
starswan, line 17: would be good to record the origin and version of this 3rd-party component in comments and/or a filename. We can use a non-minified version of such components, as the asset pipeline minifies things anyway, which is sometimes useful if a bugfix is released then it can be diffed against the original.

app/models/case_information.rb
```diff
@@ -23,22 +25,67 @@ def latest_early_allocation
...
 
-  validates :welsh_offender, inclusion: {
-    in: %w[Yes No],
-    allow_nil: false,
-    message: 'Select yes if the prisoner’s last known address was in Wales'
-  }
-  validates :tier, inclusion: { in: %w[A B C D], message: 'Select the prisoner’s tier' }
+  validates :tier, inclusion: { in: %w[A B C D N/A], message: "Select the prisoner's tier" }
 
   validates :case_allocation, inclusion: {
```
starswan, line 39: I wonder if it might have been better to allow nil rather than the special value 'N/A'? (here in in the tier field)

app/models/case_information.rb
```diff
@@ -23,22 +25,67 @@ def latest_early_allocation
   validates :manual_entry, inclusion: { in: [true, false], allow_nil: false }
   validates :nomis_offender_id, presence: true, uniqueness: true
 
-  validates :local_divisional_unit, :team, presence: true, unless: ->{ manual_entry }
+  validates :team,
```
starswan, line 28: I'm unsure about this validation. I think it needs a comment to explain why team is mandatory when we're not entering manually (i.e. from nDelius) and the probation service is (not?) Scot/NI - I'm finding it hard to understand the 'unless' inverted logic in this case. (Actually I think the logic is slightly wrong - I think the intention is to validate the team when manually entering English Welsh cases)

app/models/case_information.rb
```diff
@@ -23,22 +25,67 @@ def latest_early_allocation
...
+    false
+  end
+
+  def save_scottish_or_ni
+    self.tier = 'N/A'
+    self.case_allocation = 'N/A'
+    self.team = nil
+  end
+
+  def english_or_welsh?
```
starswan, line 76: IMHO this is better expressed as just: 
last_known_location == 'No' || probation_service == 'Wales' || probation_service == 'England'

config/initializers/assets.rb
```diff
@@ -17,4 +17,4 @@
...
+Rails.application.config.assets.precompile += %w( accessible-autocomplete.min.js )
```
starswan, line 20: Don't think this is required, as the file should already be in the right place to be picked up by the asset pipeline?

spec/controllers/tasks_controller_spec.rb
```diff
@@ -114,8 +114,12 @@
     it 'can show offenders needing nDelius updates' do
       stub_offender(offender_no, booking_number: 754_206)
```
starswan, line 116: unsure about this change - it seems unrelated to the story. Concerned that we are changing the sense of a test, and it passes by accident than by design.

It is possible to have a nil team when probation service is NI or Scotland, so maybe that might have been a better change to make here?

spec/features/active_admin_feature_spec.rb
starswan, line 34: It might be nice to turn this into a trait 'nilteam' so that it doesn't look weird that we are picking (validly) probation_service Scotland for no obvious reason

app/models/case_information.rb
starswan: I think tier should still be mandatory unless the offender is Scotland/NI ?

app/assets/javascripts/application.js
ollietreend: Ideally we'd use the JavaScript package manager (NPM or Yarn) to handle this dependency, rather than checking the file in to our repo.

It looks like we have `package.json` and `yarn.lock` files already set up – do you know if the project is already using this?

starswan: we do use NPM packages already for the GOV.UK styles, so I think it should be possible?

app/controllers/case_information_controller.rb
```diff
@@ -10,7 +10,50 @@ def new
...
+                                       'Yes'
+                                     end
+    @team_name = Team.find_by(id: @case_info.team_id)&.name
+  end
+
+  def update
+    @case_info = CaseInformation.find_by(nomis_offender_id: case_information_params[:nomis_offender_id])
+    @prisoner = prisoner(case_information_params[:nomis_offender_id])
+
+    # we only send email if the ldu is different from previous
```
starswan, line 47: IMHO there is a more idiomatic way of implementing this. 
If the controller just updates model attributes (via model[:attr] = value) rather than calling update, then the changes will not be saved (yet). Then the controller can look at the 'team_was' attribute to determine the old team value (and hence the old LDU value)
P.S. it is a little un-idiomatic to have a method ending with a ? peform a side effect.

app/assets/javascripts/application.js
ollietreend: The Sprockets setup isn't quite as slick as Webpacker – it looks like the JS module directory needs to be registered as an include path in `config/initializers/assets.rb`:

https://github.com/ministryofjustice/offender-management-allocation-manager/blob/e22f0e37c670cb3b3b8b375e9b34f2a6be36e0a6/config/initializers/assets.rb#L8-L15

The `node_modules` directory is already there, so that might already be enough.

app/models/case_information.rb
ollietreend: It looks like the other side-effect of this change is that we lose the validation on `:local_divisional_unit` needing to be present.

Currently, the nightly Delius import fails if the offender is assigned to a team which doesn't have an LDU – this is covered by error code `DeliusImportError::MISSING_LDU`.

If I'm following this correctly, dropping the validation here will mean that is _will_ be possible to assign offenders (either manually or automatically) to a team which doesn't have an LDU.

Is that something we want to do?

As an aside: we may want to adjust the 'Teams' autocomplete list to filter out teams which don't have an LDU – so a manual entry can never trigger the 'missing LDU' error type.

app/models/case_information.rb
ollietreend: I tend to agree – it feels like `nil` might be a more meaningful value for 'tier' and 'case allocation' when they aren't set, and then display that `nil` value as a human-friendly `"N/A"` on the page.

app/models/case_information.rb
```diff
@@ -23,22 +25,67 @@ def latest_early_allocation
...
+              in: %w[Yes No],
+              allow_nil: true,
+              message: "Select yes if the prisoner's last known address was in Northern Ireland, Scotland or Wales"
+            }, if: -> { manual_entry }
+
+  def scottish_or_ni?
+    return true if last_known_location == 'Yes' &&
+      (probation_service == 'Scotland' || probation_service == 'Northern Ireland')
+
+    false
```
ollietreend, line 67: How would you feel about removing the explicit `return`s (guard clauses) in favour of something like this?

```suggestion
    last_known_location == 'Yes' &&
      (probation_service == 'Scotland' || probation_service == 'Northern Ireland')
```

I feel like the guard clauses might be causing unnecessary noise in these methods. This would also make the implementations more consistent with the `english?` method.

app/models/case_information.rb
```diff
@@ -23,22 +25,67 @@ def latest_early_allocation
...
+
+  def english?
+    last_known_location == 'No'
+  end
+
+  def stage2_filled?
+    return false if tier.nil? || team.nil? || case_allocation.nil?
+
+    true
+  end
```
ollietreend, line 90: I think I've answered my own question here. Are `nil` values in the 'tier' and 'case allocation' fields reserved to mean "The user hasn't completed this field", rather than "The user chose N/A"?

Also – I wonder if this could be easier to read if the logic is flipped:
```suggestion
  def stage2_filled?
    tier.present? && team.present? && case_allocation.present?
  end
```

app/controllers/case_information_controller.rb
```diff
@@ -97,10 +110,121 @@ def nomis_offender_id_from_url
     params.require(:nomis_offender_id)
   end
 
+  def stage1_errors
+    if case_information_params[:last_known_location].nil?
```
starswan, line 114: really unsure about this 'removing' of errors. Not even sure if its 'allowed' by rails - my understanding is that model errors are mutable so that you can add to them, unsure what effect removing errors would have. 
It's not clear to me why this code is here and what problems it addresses? Maybe its related to the validations rules not being as precise as we need?

PR #1186, opened 2020-06-08
Fix broken tests which relied on ex-team member

theKHutDeveloper:
Replace signed_in user with MOIC_POM, instead of relying on team members so that the majority of tests continue to work when team member leaves

### DIFF
--- a/spec/controllers/allocations_controller_spec.rb
+++ b/spec/controllers/allocations_controller_spec.rb
@@ -96,7 +96,7 @@ RSpec.describe AllocationsController, :versioning, type: :controller do
       stub_request(:get, "https://gateway.t3.nomis-api.hmpps.dsd.io/elite2api/api/staff/1").
         to_return(status: 200, body: {}.to_json, headers: {})
 
-      stub_request(:get, "https://gateway.t3.nomis-api.hmpps.dsd.io/elite2api/api/users/PK000223").
+      stub_request(:get, "https://gateway.t3.nomis-api.hmpps.dsd.io/elite2api/api/users/MOIC_POM").
         to_return(status: 200, body: { staffId: 3 }.to_json, headers: {})
 
       stub_request(:get, "https://gateway.t3.nomis-api.hmpps.dsd.io/elite2api/api/staff/4").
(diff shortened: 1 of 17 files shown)


spec/controllers/allocations_controller_spec.rb
```diff
@@ -167,7 +167,26 @@
     end
 
     context 'without DeliusDataJob' do
-      render_views
```
starswan, line 170: think this was removed by accident?

starswan: don't think this line should have been deleted

theKHutDeveloper: I've put it back

ollietreend: I can see that this allocation fixture `before` block has been moved one level up into the parent `context 'without DeliusDataJob'` block, where it used to be inside the child `context 'with an allocation'` block.

It looks like we create a different allocation fixture in the other child context block `context 'with a different allocation'`, so I don't think it uses the first allocation fixture.

I'm wondering if it makes more sense to keep it in the `context 'with an allocation'` block. Or is there are reason for it being moved up into the parent?

spec/services/prison_offender_manager_service_spec.rb
```diff
@@ -38,7 +38,7 @@
          vcr: { cassette_name: :pom_service_get_poms_list } do
         expect(subject).to be_kind_of(Enumerable)
         # 1 POM in T3 (Toby Retallick) is marked inactive, so expect one less active one
```
starswan, line 40: silly question, but why has this test changed? Did you remove Toby R as well?

ollietreend:
@theKHutDeveloper it looks like you and Ross are hardcoded in the `PrisonOffenderManagerService` spec in the `#get_pom_name` and `#get_user_name` tests:
https://github.com/ministryofjustice/offender-management-allocation-manager/blob/8097da0478501fecfe6c72ea4fd590f4c8a4739d/spec/services/prison_offender_manager_service_spec.rb#L12-L27

PR #1195, opened 2020-06-11
Pom725 calculate determinate/indeterminate sentence type from nomis API

starswan:
It appears that certain offenders don't appear on our service. We have a fixed list of sentence types hard-coded, which we use to decide whether the sentence is determinate or otherwise.
This PR attempts to use the NOMIS Elite2 API for this purpose, although from experimentation it doesn't always work as expected, so we are falling back to the hard-coded list if we don't get a sensible answer from NOMIS


starswan:
This looks like the wrong solution to this problem. MO-199 replaces this ticket with a better simpler versikon using the PrisonSearchAPI

PR #1199, opened 2020-06-15
POM-739 Delius import bugfix

ollietreend:
Jira ticket: [POM-739](https://dsdmoj.atlassian.net/browse/POM-739)

This pull request fixes a bug where cell values from the Delius extract XLSX spreadsheet were being read incorrectly if they contained an ampersand (`&`).

In particular, this bug was causing problems with reading team names from the Delius spreadsheet which contained ampersands. For example, a team named `"N&S-Bury St Edmunds"` was being read as just `"N"` – the name was truncated just before the ampersand.

### Replace custom XLSX reader with `roo` gem

The `Delius::Processor` component that read the XLSX files was custom-built and hard to understand if you're not familiar with the internals of the XLSX file format. Ideally our service shouldn't be concerned with the particulars of reading XLSX files – this is a solved problem, and not one that we're interested in solving again.

One of the key considerations when the custom XLSX parser was originally implemented in [`2d07dc7`](https://github.com/ministryofjustice/offender-management-allocation-manager/commit/2d07dc7f9efbbf5e6a2b48b85b568a95f381c504) was that it should 'stream' the rows of the spreadsheet, rather than loading the entire file into memory. This is due to the fact that we're likely to be reading a ~10 MB spreadsheet containing ~100k rows of data, so loading the entire file at once would exhaust the available memory.

I've replaced our custom XLSX-parsing functionality with [a gem called `roo`](https://github.com/roo-rb/roo#readme).

Roo is a popular library capable of reading XLSX files. Crucially, it has a feature which allows it to [stream rows from a XLSX spreadsheet](https://github.com/roo-rb/roo#excel-xlsx-and-xlsm-support), meaning we can continue processing large spreadsheets without exhausting the available memory.

### Refactor responsibility of the 'reader' class

I've replaced the class `Delius::Processor` (custom XLSX reader) with a new class `Delius::ExtractFileReader`. This new class has a slightly different responsibility to the one it replaces.

`Delius::Processor` was only responsible for reading rows of the spreadsheet into arrays of strings. It didn't transform those values into `field name` => `cell value` Hash objects – this became the responsibility of whatever was consuming the data.

As part of this refactor, I've made the new `Delius::ExtractFileReader` responsible for performing that transformation. Rather than returning arrays of cell values, it returns Hash objects mapped to field names.

Each Hash returned from `Delius::ExtractFileReader` takes the form:

```
{
  crn: 'crn code',
  pnc_no: 'pnc num',
  noms_no: 'G3356GT',
  fullname: 'Bobby Pin',
  tier: 'A1',
  roh_cds: 'risk of harm',
  offender_manager: 'ROSS JONES',
  org_private_ind: 'N',
  org: 'NPS',
  provider: 'NPS',
  provider_code: 'C1234',
  ldu: 'NPS Bolton',
  ldu_code: 'N014403',
  team: 'Team 1',
  team_code: 'A',
  mappa: 'Y',
  mappa_levels: '1',
  date_of_birth: '19/03/1985'
}
```

This has allowed me to DRY up the code which is responsible for that Array => Hash transformation. It was being being performed in three different areas of the codebase – this now happens in just one place.

### Test coverage

I've included tests to cover specific known cases where LDU and team names contain ampersands.


ollietreend:
I've marked this PR as 'draft' because it still needs some work. I've added a comment to [the Jira ticket POM-739](https://dsdmoj.atlassian.net/browse/POM-739) documenting status and potential next steps.

TL;DR:

I've run this in preprod with a real Delius extract spreadsheet. It looks like there's a memory leak somewhere, as kubernetes kills the container before the job completes.

https://github.com/ministryofjustice/offender-management-allocation-manager/blob/97cc19268b2e4c6075df02293c2145dfdea9efe1/app/jobs/delius_import_job.rb#L95-L101

In my test on preprod, it completed the first and second pass of the sheet, but fails about half way through the third pass:

1. ✅ `update_team_names_and_ldus` completed successfully
2. ✅ `update_shadow_team_associations` completed successfully
3. ❌ `upsert_delius_data_records` – memory ran out about half way through this method

Next steps:
- Investigate memory usage of `roo`. Could garbage collection help?
- Investigate other XLSX-reader gems instead of roo. I think [creek](https://github.com/pythonicrubyist/creek) could be a good candidate.

ollietreend:
I'm closing this Pull Request as it's no longer needed:

- The bug in the existing XLSX reader was fixed in #1218 – which was easier than resolving the memory leak in this PR.
- In the not-too-distant future, we want to move away from the XLSX Delius import and over to the Community API – at which point, this XLSX reader won't be needed anyway.

PR #1202, opened 2020-06-17
use relative rather than absolute dates in test

starswan:
We have a feature test that used absolute dates, hence it was setup to fail at some point in the future - which happened to be today (17-Jun-2020) - due to their handover date being reached and the responsibility rules changing as a result.


starswan:
> Looks good. I can see the dates you supply to `stub_api_calls_for_prison_allocation_path` are now `Date` objects rather than strings – presumably this doesn't matter?

Seemed ok - had a quick look, and stub_api_calls_for_prison_allocation_path calls to_json on its arguments at the end, so I think it's fine.

PR #1209, opened 2020-06-19
 Add data fields to debugging page

Katy600:
This commit expose two additional date fields in the ‘offence’ table on Debugging page: post recall release and License expiry date.

As part of this ticket it created a new spec/views for the debugging page, adding tests for existing code: prisoner’s full name and tariff-date, to ensure the test data is set up correctly. It then tests for post_recall_release_date and license_expiry_date - to display on this page. Nokogiri has been used to help make these tests more precise - specifying where we expect the dates to appear on the page.


spec/views/debugging/debugging.html.erb_spec.rb
```diff
@@ -0,0 +1,75 @@
+require 'rails_helper'
+
+RSpec.describe "debugging/debugging", type: :view do
```
starswan, line 3: It would be nce to 'DRY' up this to improve readability as its testing the same page

spec/factories/offenders.rb
```diff
@@ -44,4 +44,29 @@ def initialize(data)
       }
     }
   end
+
```
starswan, line 47: we don't need this factory any more - you can use 'offender'instead

spec/views/debugging/debugging.html.erb_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+        automatic_release_date: Time.zone.today + 10.months)
+      prison = build(:prison)
+
+      assign(:offender, OffenderPresenter.new(offender, nil))
+      assign(:prison, prison)
+
+      render
+      page = Nokogiri::HTML(rendered)
+      full_name = page.css('#prisoner-information').css('#name').first
+      expect(full_name.text).to match(/Dory, John/)
```
starswan, line 18: eq?

PR #1217, opened 2020-06-26
POM-764 Stop using the 'release date' in MOIC filter

theKHutDeveloper:
We are no longer using 'release_date' as a means to determine if an offender
is sentenced and are no longer collecting this information from NOMIS.

If the offender has one at least one of the following, then we consider them to
be sentenced:

Conditional release date (CRD)
Automatic release date (ARD)
Home detention curfew actual date (HDCAD)
Post recall release date (PRRD)
Sentence licence expiry date (SLED)
Parole eligibility date (PED)
Home detention curfew eligibility date (HDCED)
Tariff end date (TED)
Offender has an indeterminate sentence


spec/features/when_nomis_is_missing_information_spec.rb
```diff
@@ -42,7 +42,7 @@
           stub_bookings = [{
             bookingId: 1,
             sentenceDetail: {
-              releaseDate: 30.years.from_now.iso8601,
+            releaseDate: 30.years.from_now.iso8601,
```
starswan, line 45: this is no longer valid - hence the (unintended) change below. I think this needs to be conditionalReleaseDate as above?

app/models/nomis/offender_base.rb
```diff
@@ -27,15 +27,23 @@ def convicted?
...
+      # 5) Sentence licence expiry date (SLED) or
+      # 6) Parole eligibility date (PED) or
+      # 7) Home detention curfew eligibility date (HDCED) or
+      # 8) Tariff end date (TED)
+      # 9) Or have an indeterminate sentence
+
       return false if sentence&.sentence_start_date.blank?
 
-      sentence.release_date.present? ||
+      sentence.conditional_release_date.present? ||
```
starswan, line 42: The ticket suggested that we should check certain dates based on the sentence type. It would be really nice to do this, as it would start to put some real 'domain logic' into the codebase and start to make it a bit more encapsulated. It might be a pre-requirement that 'sentenceType' gets pushed into 'sentenceDetail' but that in itself sounds like a good thing. What do you think?

spec/features/when_nomis_is_missing_information_spec.rb
theKHutDeveloper: This was to show that we do not have any of the required dates the page would not crash, admittedly I could have just removed the 'releaseDate' and the result would have been the same. It would probably be clearer to the reader. I'll fix that

app/models/nomis/offender_base.rb
theKHutDeveloper: Not sure the best way to do that

starswan: I think there is a simpler solution than the one I proposed above - we can simply look at @sentence_type (e.g. @sentence_type.recall?) to decide which dates are valid in which scenarios

PR #1243, opened 2020-07-15
Pom511 Part1b - stop using welsh_offender column

starswan:
This PR stops using the old welsh_offender column, as this data can now be calculated by using the new probation_service column. Before this PR can be deployed, the rake task backfill:probation_service needs to be run so that all the case_information records have probation_service populated


ollietreend:
Closing as these changes are included in #1263, which is the main PR for POM-511. If we want to partially deploy any of this feature, we should pull fresh PRs out of commits in that branch.

starswan:
RE-open as we still need this for preprod

PR #1245, opened 2020-07-15
POM-757 change caseload route to include staff_id

Katy600:
This pull request changes the caseload route to include staff_id

### DIFF
--- a/app/controllers/caseload_controller.rb
+++ b/app/controllers/caseload_controller.rb
@@ -1,15 +1,14 @@
 # frozen_string_literal: true
 
 class CaseloadController < PrisonsApplicationController
-  before_action :ensure_pom
-  before_action :load_pom
+  before_action :ensure_signed_in_pom_is_this_pom, :load_pom
 
   breadcrumb -> { 'Your caseload' },
-             -> { prison_caseload_index_path(active_prison_id) }
+             -> { prison_staff_caseload_index_path(active_prison_id, staff_id) }
   breadcrumb -> { 'New cases' },
-             -> { new_prison_caseload_path(active_prison_id) }, only: [:new]
+             -> { new_prison_staff_caseload_path(active_prison_id, staff_id) }, only: [:new]
   breadcrumb -> { 'Cases close to handover' },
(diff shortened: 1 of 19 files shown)


app/controllers/caseload_controller.rb
```diff
@@ -84,13 +83,21 @@ def filter_allocations(allocations)
...
-    @pom = PrisonOffenderManagerService.get_signed_in_pom_details(
-      current_user,
-      active_prison_id
-    )
+    @pom = StaffMember.new(params[:staff_id])
+  end
+
+  def ensure_signed_in_pom_is_this_pom
+    user = Nomis::Elite2::UserApi.user_details(current_user)
+    unless staff_id == user.staff_id || current_user_is_spo?
```
starswan, line 91: think you forgot to check that we are actually a POM?

spec/services/allocation_service_spec.rb
```diff
@@ -47,7 +47,7 @@
             "nomis_offender_id" => "G4273GI",
             "pom_email" => "pom@digital.justice.gov.uk",
             "pom_name" => "Moic",
-            "url" => "http://localhost:3000/prisons/LEI/caseload"
+            "url" => "http://localhost:3000/prisons/LEI/staff/485926/caseload"
```
starswan, line 50: would be nice to show that this is the correct POM (primary) and secondary below

spec/services/email_service_spec.rb
```diff
@@ -152,7 +152,7 @@
         offender_name: "Ahmonis, Imanjah",
         offender_no: "G2911GD",
         prison: 'HMP Leeds',
-        url: 'http://localhost:3000/prisons/LEI/caseload'
```
starswan, line 155: ditto here - it's not obvious that we're headed for the correct place

app/controllers/caseload_controller.rb
Katy600: @starswan  Oh that is a good point. I think I had assumed that that the signed in staff member could only be a pom or an spo.

PR #1247, opened 2020-07-16
Pom 746 debugging page fields

Katy600:
This commit exposes two additional date fields in the ‘offence’ table on Debugging page


spec/features/debugging_feature_spec.rb
```diff
@@ -72,7 +72,7 @@
     end
   end
 
-  context 'when debugging at a prison level', vcr: { cassette_name: :debugging_prison_level } do
+  context 'when debugging at a prison level' do
```
starswan, line 75: oops

PR #1254, opened 2020-07-20
POM767 pre-refactor: refactor tests to closer approximate the real world

starswan:
After reviewing the PR for POM-767, it had a poor signal-to-noise ratio, making it difficult even for the author to review it. Hence I have pulled out the noise into this refactiring PR to help with a potential discussion around solutions to POM-767

### DIFF
--- a/app/controllers/prisoners_controller.rb
+++ b/app/controllers/prisoners_controller.rb
@@ -24,7 +24,7 @@ class PrisonersController < PrisonsApplicationController
       @secondary_pom_name = helpers.fetch_pom_name(@allocation.secondary_pom_nomis_id).titleize
     end
 
-    @pom_responsibility = pom_responsibility
+    @pom_responsibility = @prisoner.pom_responsibility
     @keyworker = Nomis::Keyworker::KeyworkerApi.get_keyworker(
       active_prison_id, @prisoner.offender_no
     )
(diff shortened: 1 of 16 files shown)


app/models/allocation.rb
```diff
@@ -3,6 +3,14 @@
 class Allocation < ApplicationRecord
   has_paper_trail ignore: [:com_name]
 
+  has_one :case_information,
+          primary_key: :nomis_offender_id,
+          foreign_key: :nomis_offender_id,
+          inverse_of: :allocation,
+          # allocation(histories) are never destroyed in the normal cycle,
+          # they are kept in case offender returns to the system
+          dependent: :restrict_with_exception
```
ollietreend, line 12: This would also prevent us from deleting the associated `CaseInformation` record. Is that the intended behaviour?

@starswan I think I remember you mentioning that `CaseInformation` records are deleted as part of movements – i.e. when an offender moves out of prison, their record is deleted.

starswan: It's meant to stop the case_information record cascade deleting the allocation. Allocations are never deleted in the normal cycle (as it says in the comment)

ollietreend: @starswan Maybe I'm thinking 'in the wrong direction'. I thought it worked the other way around.

[From the docs](https://guides.rubyonrails.org/association_basics.html#options-for-has-one-dependent) for `:dependent` on `:has_one` associations:

> Controls what happens to the associated object **when its owner is destroyed**:
> - `:restrict_with_exception` causes an `ActiveRecord::DeleteRestrictionError` exception to be raised if there is an associated record
> - `:restrict_with_error` causes **an error to be added to the owner** if there is an associated object

(bold added by me)

I read this as: "The owner record cannot be deleted if it has any associated records"

Note that `:restrict_with_error` (which isn't what we're using, but it works in the same 'direction') says it adds an error **"to the owner"** record.

In this case:

- "owner" = CaseInformation
- "associated" = Allocation

So I've interpreted that as "The CaseInformation record cannot be deleted if it has any associated Allocations"

Isn't that the opposite of what we're expecting? i.e. we want to be able to delete CaseInformation records but not the associated Allocations

starswan: There's a test for it. Yes the owner cannot be deleted (i.e. the allocation) if there are any case_information records.

ollietreend: Yes – my mistake. I was getting it the owner/associated records the wrong way around.

PR #1256, opened 2020-07-20
Pom511 autocomplete via yarn

starswan:


ollietreend:
@starswan It looks like Heroku is choking when trying to compile JS assets. Not sure why that same issue isn't flagged up in the Circle tests – maybe it doesn't attempt a full production build of JS assets.

See build log here: https://dashboard.heroku.com/apps/moic-pom511-autocomplet-cljwxv/activity/builds/1a94fbac-2a4d-456c-bc32-97157cbb8d2e

The line which stood out to me:

```
Uglifier::Error: Unexpected token: name (element). To use ES6 syntax, harmony mode must be enabled with Uglifier.new(:harmony => true).
```

It looks like it could be something to do with ES6 syntax being used, but Sprockets not knowing how to compile it...? Or being misconfigured?

PR #1262, opened 2020-07-22
POM-511 last step - remove welsh_offender column

starswan:
This is the very last step of POM-511 - removing the old welsh_offender column from the database table.


ollietreend:
Closing as these changes are included in #1263, which is the main PR for POM-511. If we want to partially deploy any of this feature, we should pull fresh PRs out of commits in that branch.

PR #1263, opened 2020-07-22
Pom511 update team manually

starswan:
POM-511 (nuff said)
I have split this story into a number of commits - the original work is the first commit, and all subsequent commits should be logical stages towards teh end goal.
I'm pretty sure there's still some work to be done on this story - so please pore over it with a fine tooth-comb, and make it the best we can be.


db/seeds.rb
```diff
@@ -76,19 +76,21 @@
...
 
 CaseInformation.find_or_create_by!(nomis_offender_id: 'G7658UL') do |info|
-    info.tier = 'A'
-    info.case_allocation = 'NPS'
-    info.welsh_offender = 'Yes'
-    info.manual_entry =  true
-    info.team_id = team1.id
+  info.tier = 'A'
+  info.case_allocation = 'NPS'
+  info.welsh_offender = 'Yes'
```
ollietreend, line 81: Shouldn't `welsh_offender` be removed from these seeds, given that we've now added the `probation_service` field?

spec/factories/case_information.rb
```diff
@@ -28,6 +24,15 @@
...
 
-    crn { Faker::Alphanumeric.alpha(number: 10) }
+    crn do Faker::Alphanumeric.alpha(number: 10) end
+
+    probation_service do
+      'Wales'
+    end
+
+    trait :no_team do
+      probation_service { 'Scotland' }
```
ollietreend, line 34: Is 'Scotland' an allowed value for the `probation_service` field?

The CaseInformation model has a validation rule to only allow 'England' and 'Wales'.

So unless I'm missing something, I think the `:no_team` trait will create a CaseInformation record which could never exist in the wild.

ollietreend: Actually – scratch that. I see that the additional allowed values ('Scotland', 'Northern Ireland') were added to the validation rule in the next commit.

app/controllers/case_information_controller.rb
```diff
@@ -97,10 +122,75 @@ def nomis_offender_id_from_url
...
+    render :new
+  end
+
+  def handle_stage2_rendering
+    @prisoner = prisoner(case_information_params[:nomis_offender_id])
+    render :missing_info
+  end
+
+  def team_identifier
+    Team.find_by(name: params['input-autocomplete'])&.id
```
ollietreend, line 136: It looks like the value in the autocomplete field comes back to us as a team name, and we're using that identify the team – i.e. we find the team by its name, and convert that into its respective ID.

**This is only safe to do if the team names are unique.** Otherwise, we won't be able to identify which team was selected given a duplicate team name.

But it looks like the name uniqueness constraint on team name **only applies to NPS teams**:
https://github.com/ministryofjustice/offender-management-allocation-manager/blob/fa312de2842c019129848e21d5d142fcb5b42fc0/app/models/team.rb#L8

**There's a chance that we have duplicate CRC team names.**

Whilst we won't need to send out automatic handover emails for CRC offenders, it's important to note that we may be capturing the wrong data as it goes in at this end. So CRC team data might not be reliable if needed in the future.

It looks like a crack that things could fall down and come back to bite us in the future.

Are we okay with knowingly taking this on?

app/controllers/case_information_controller.rb
```diff
@@ -5,12 +5,67 @@ class CaseInformationController < PrisonsApplicationController
...
+      @case_info = CaseInformation.new(case_information_params.merge(team_id: team_identifier, manual_entry: true))
+      unless @case_info.valid?
+        handle_stage2_rendering
+      end
+    end
+    if @last_location.valid? && @case_info.valid?
+      @case_info.save!
+      send_email
+      redirect_to prison_summary_pending_path(active_prison_id, sort: params[:sort], page: params[:page])
+    end
```
ollietreend, line 40: I think there are could be a few code smells in this method which would be nice to clear up.

A few things which stand out to me:

- Naming between the steps/stages of the form is inconsistent. We have two stages – sometimes called 'stage2' (e.g. method `handle_stage2_rendering`), and sometimes referred to by name (e.g. `params[:stage] == 'last_location'`).
  - Stage 1 is referred to as:
    - 'last location' – i.e. we receive a param `params[:stage] == 'last_location'`
    - 'address page' – i.e. `display_address_page` method
    - The view for this stage is `new.html.erb` – which isn't immediately obvious that it's for this stage
  - Stage 2 is referred to as:
    - 'missing info' – i.e. `render :missing_info`
    - 'stage 2' – i.e. `handle_stage2_rendering`
    - The view for this stage is `missing_info.html.rb` – out of context, this might read like the name of an error page. Perhaps `add_missing_info` would be a more explicit name.
- I find the control flow in this method quite hard to follow. I wonder if there's any way we could simplify this.
- It's not immediately obvious which stage of the form comes first when we refer to them by name. I'm not sure if this could become clearer through simplified control flow, or maybe just a comment in the method.
- Smell: It feels a bit hacky that the final save (line 37) relies upon always having a truthy `@last_location.valid?` – so in stage 2 (line 30) we actually have to create a fake LastLocationForm object which returns true. I wonder if a change in control flow could help here, too.

This could be worth a conversation to talk it through – or I'd also be happy to take a crack at seeing if I can make any improvements myself.

ollietreend:
It looks like there's a bug in the validation when running through this as a user would.

When moving from stage 1 (last known address) to stage 2 (add missing information), validation errors already display on the page even when it's the first time you've seen the stage 2 form.

Validation bug

app/presenters/offender_presenter.rb
```diff
@@ -8,9 +8,9 @@ class OffenderPresenter
            :full_name_ordered, :full_name, :main_offence,
            :sentence_start_date, :team, :prison_id,
            :home_detention_curfew_eligibility_date, :home_detention_curfew_actual_date,
-           :tariff_date,
+           :tariff_date, :probation_service,
            :date_of_birth, :release_date, :parole_eligibility_date,
-           :welsh_offender, :case_allocation, :earliest_release_date,
+           :welsh_offender, :case_allocation, :probation_service, :earliest_release_date,
```
ollietreend, line 13: We've included `:probation_service` twice in this list – only needed once.

app/services/case_information_service.rb
```diff
@@ -7,4 +7,16 @@ def self.get_case_information(offender_ids)
...
+    return false if team_ids.blank?
+
+    return false if team_ids.last.nil? # from eng/wales to scot/ni
+
+    return true if team_ids.first.nil? # from scot/ni to eng/wales
+
+    old_ldu = Team.find_by(id: team_ids.first).try(:local_divisional_unit)
+    new_ldu = Team.find_by(id: team_ids.last).try(:local_divisional_unit)
+    old_ldu != new_ldu
+  end
```
ollietreend, line 21: This method feels very similar to some of the 'dity' methods which come with ActiveRecord models. (relevant docs: [here](https://api.rubyonrails.org/classes/ActiveModel/Dirty.html) and [here](https://api.rubyonrails.org/classes/ActiveRecord/AttributeMethods/Dirty.html))

It makes me think that this method would probably be more at home in the `CaseInformation` model, rather than in the `CaseInformationService` object.

I'm thinking that we'd be able to call something like `@case_info.was_ldu_changed?` after having called `@case_info.update`.

app/controllers/case_information_controller.rb
starswan: Agreed. This is something I had considered. Maybe having a 'stage1' method in the routing, so that we hit that rather than 'create' from the first form?

app/controllers/case_information_controller.rb
starswan: There is one more commit which I think I need to add to this branch - I was uncertain about it as I couldn't work out how to implement the 'Remove' button (but I think it can just go anyway) which limits the team to those with email addresses. I guess there's something weird here about CRC cases that we have missed - good spot

ollietreend: 17f6eb1 seems to have done the trick 👍 now there's no ambiguity about duplicate team names – we simply receive the ID.

app/services/case_information_service.rb
starswan: +1

starswan:
> It looks like there's a bug in the validation when running through this as a user would.
> 
> When moving from stage 1 (last known address) to stage 2 (add missing information), validation errors already display on the page even when it's the first time you've seen the stage 2 form.
> 
> Validation bug

Just fixed this

ollietreend:
I've just pushed changes to the new/edit journey so they're now unified.

One thing to note with this change:

- I've removed the 'Remove' option under the team autocomplete. The JavaScript for this was broken, and I didn't see it mentioned in the POM-511 ticket/design, so I removed it rather than fix it.
- However, there's probably a reason why that existed. So it may need to go back in.
- If it's no longer needed, I forgot to remove the CSS for it in `case_info.scss`, so that still needs to be done.

app/views/case_information/_probation_data_form.html.erb
```diff
@@ -55,7 +55,7 @@
       <span class="govuk-hint">Enter the full name of the Team, for example 'NPS Croydon'</span>
 
       <%= form.select(:team_id,
-                      Team.with_ldu.map { |team| [team.name, team.id] },
+                      Team.with_ldu.map { |team| [team.name, team.id] }.sort_by { |team| team.first },
```
ollietreend, line 58: @theKHutDeveloper Sorry to be a pest – but it would be more efficient for us to do this sort at the database level rather than in-memory with Ruby.

i.e. instead of using `.sort_by` on the array, we could use `.order` on the ActiveRecord query:

```suggestion
                      Team.with_ldu.order(name: :asc).map { |team| [team.name, team.id] },
```

The difference might be negligible for small result sets, but here we know we'll be dealing with over 1000 team records, so would prefer we didn't sort that in memory.

ollietreend:
I have a feeling the database migration in this branch might be missed whenever we do deploy this to staging/production.

Since we have a 'newer' database migration in `master` now – even when we rebase, the migration here will have an older timestamp. Which _I think_ means Rails won't notice the 'older' migration from this PR, because it's got an older timestamp.

i.e. in `master` right now:
https://github.com/ministryofjustice/offender-management-allocation-manager/blob/6b6909110828fe36a350c4ee80a13783443cd50b/db/schema.rb#L13

but in this branch:
https://github.com/ministryofjustice/offender-management-allocation-manager/blob/f8f02676928bcb308894c3662a7a84284157e9a3/db/schema.rb#L13

Note that the version in `master` has a newer version timestamp.

I imagine this DB migration won't rebase/merge as cleanly as our code changes.

@starswan is that right?

PR #1269, opened 2020-07-31
POM-767 / POM-718 CRC Handover cases

starswan:
This PR brings back the CRC handover story, and also fixes the defect whereby 'unknown' cases were being treated as CRC cases from a handover pespective


app/models/nomis/offender_base.rb
```diff
@@ -10,6 +10,8 @@ class OffenderBase
 
     delegate :indeterminate_sentence?, to: :@sentence_type
```
ollietreend, line 11: Looks like this line needs removing? In favour of the one added just below.

PR #1271, opened 2020-07-31
POM-654: Banner to share new feature updates with users

theKHutDeveloper:
This pull request is in response to a desire to share new features and updates with our users.

This will result in: 

a static ’What’s new page’
a link in the footer to the ‘What’s new’ page
a YAML file that will hold the following information needed to display a service notification banner
a yellow notification alert banner with information for the user
a close button so the user can dismiss the notification

-     id: this should be a unique identifier as this will be used to hide the notification should a user dismiss it
-     start_date: the first day that the service notification should be displayed on the service
-     duration: the number of days that the service notification should be displayed on the service
-     role: the type of user that should be shown this service notification
-     text: the text that should be displayed on the service notification banner

Any service notification relevant to the user based on their role (Allocation manager and/or Allocation case manager) will be displayed until the user removes it OR the duration for the notification to be displayed has passed. The service notification will be visible on all views except static pages.

We are utilising javascript and local storage to permanently hide notifications if the user clicks on the close button.

**What's new page**
Screen Shot 2020-07-31 at 09 16 38

**Footer link to What's new page**
Screen Shot 2020-07-31 at 09 16 22

**Service notification banner**
Screen Shot 2020-07-30 at 13 58 27


app/controllers/pages_controller.rb
```diff
@@ -40,6 +40,8 @@ def help; end
 
   def contact; end
 
+  def whats_new; end
```
starswan, line 43: IMHO a 'service notification' should be a model-type object. This action should populate it, and the view should display it?

app/assets/config/manifest.js
```diff
@@ -1,3 +1,4 @@
 //= link_tree ../images
+//= link @ministryofjustice/frontend/moj/assets/images/icon-close-cross-black.svg
```
starswan, line 2: we shouldn't need this?

app/assets/javascripts/service_banner.js
```diff
@@ -0,0 +1,20 @@
+document.addEventListener("turbolinks:load", function() {
+    let buttons = document.getElementsByClassName('close-service-notification');
```
starswan, line 2: The sprockets pipeline seems to have a problem with ES6. Need to either convert or tweak the pipeline to make it work. IN POM-511 I just converted to ES5.

app/assets/javascripts/service_banner.js
```diff
@@ -0,0 +1,20 @@
...
+
+    for (let item of buttons) {
+        if(localStorage.getItem(item.id)) {
```
starswan, line 5: am I missing something? localStorage? Do all our browsers support it?

spec/services/service_notifications_service_spec.rb
```diff
@@ -0,0 +1,115 @@
...
+              "id" => "my-new-feature",
+              "start_date" => start_date,
+              "role" => ["POM"],
+              "duration" => 30,
+              "text" => "This is a test banner"
+            }
+          ]
+        }
+
+      allow(YAML).to receive(:load).with(File.read(filename)).and_return(data)
```
starswan, line 74: For me this is a code smell. TDD is meant to help with structure. This says that we are welded to the filesystem and YAMl loading. IMHO we should have an iterable of notifications (which can then be passed a YAML.load by someone else) thus making this mock redundant.

starswan: see above comment about structure - this could be SystemNotification.future (where the future() method is responsible for the filtering)

app/controllers/application_controller.rb
```diff
@@ -23,6 +23,11 @@ def current_user_is_spo?
     sso_identity.current_user_is_spo?
   end
 
+  def service_notifications
```
starswan, line 26: IMHO this should follow the pattern for PrisonsApplicationController#load_staff_id making it a before action rather than a helper method

app/assets/javascripts/service_banner.js
theKHutDeveloper: We want the data to remain longer than just a 'session'

app/assets/javascripts/service_banner.js
theKHutDeveloper: Done

app/assets/config/manifest.js
theKHutDeveloper: Doesn't work without it

app/controllers/pages_controller.rb
theKHutDeveloper: No, because the YAML just holds enough information about the banner. Discussed it with @ollietreend and we decided to separate the 'What's new' page data from the banner. The 'What's new' page is to be updated manually (hard-coded)

app/assets/javascripts/service_banner.js
```diff
@@ -0,0 +1,20 @@
+document.addEventListener("turbolinks:load", function() {
+    var buttons = document.getElementsByClassName('close-service-notification');
+
+    for (var i = 0; i < buttons.length; i++) {
+        if (localStorage.getItem(buttons[i].id)) {
```
ollietreend, line 5: It might be a good idea to prefix the notification IDs so they're namespaced. Whilst we don't use JavaScript `localStorage` for anything else right now, we don't want to pollute the global namespace with notification IDs and then later find that those IDs clash with other JS features we want to release.

So for example, given a service notification with ID `"early-allocations-released"`, we could use a `localStorage` key of something like `"ServiceNotifications.early-allocations-released"`.

app/views/layouts/_service_notification_banner.html.erb
```diff
@@ -0,0 +1,13 @@
+<% unless @service_notifications.empty? %>
+  <% @service_notifications.each do |service_alert| %>
+      <div class="govuk-width-container app-site-width-container">
+        <br>
```
ollietreend, line 4: The `<br>` here it causing the main content to get bumped down the page. Because it's not inside the `<div class="service_banner">`, the `<br>` still shows when the adjacent `<div>` is hidden.

The effect is more pronounced when multiple notifications are showing and have been hidden. Whilst this is a pretty extreme example (it's never going to happen in real life!), here's a screenshot where the page is showing 10 notifications which have all been hidden – so you're seeing 10 `<br>`s pushing the page down.

Screenshot 2020-08-07 at 12 26 20

Ideally hidden banners would take up no extra space.

### Using `<br>` vs. CSS for spacing

If the `<br>` is used here to make sure there's spacing between the service banner and whatever it's close to, it might make more sense to use CSS `margin` to space it out. Using `<br>` for this feels a bit hacky, because it's not 'semantic HTML' – i.e. the `<br>` isn't actually representing a line-break in some content, but is instead being used for styling/spacing purposes.

Applying a CSS margin to the `.service_banner` element might actually solve both problems at once – CSS margins won't apply on hidden elements, and it'll be more semantic than using `<br>` HTML elements.

app/views/pages/whats_new.html.erb
```diff
@@ -0,0 +1,17 @@
+<div class="govuk-grid-column govuk-body">
+  <div class="govuk-breadcrumbs">
+    <ul class="govuk-breadcrumbs__list">
+      <li class="govuk-breadcrumbs__list-item">
+        <a class="govuk-breadcrumbs__link" href="/">Home</a>
+      </li>
+      <li class="govuk-breadcrumbs__list-item" aria-current="page">Whats-new</li>
```
ollietreend, line 7: A typo in the breadcrumb text:

```suggestion
      <li class="govuk-breadcrumbs__list-item" aria-current="page">What's new</li>
```

app/views/layouts/_service_notification_banner.html.erb
```diff
@@ -0,0 +1,12 @@
+<% unless @service_notifications.empty? %>
+  <% @service_notifications.each do |service_alert| %>
+      <div class="govuk-width-container app-site-width-container">
+        <div class="service_banner">
+          <button onclick="cancelThisNotification(this)" type="button" class="close-service-notification" id=<%= "ServiceNotifications." + service_alert['id'] %>>
```
ollietreend, line 5: Only a tiny thing, but the `id=""` HTML attribute is missing the quote marks here.

We're getting HTML output like this at the moment:

```html
id=ServiceNotifications.ollie-test>
```

Where really we should have:

```html
id="ServiceNotifications.ollie-test">
```

Browsers are usually pretty chill about this stuff, and it's technically still valid HTML 5 without the quotes. But just to be safe, it's better to have them.

PR #1272, opened 2020-07-31
Pom 648 type of case at a glance

Katy600:
Badges for case type

### DIFF
--- a/app/helpers/badge_helper.rb
+++ b/app/helpers/badge_helper.rb
@@ -1,0 +1,25 @@
+module BadgeHelper
+  def badge_colour(offender)
+    if offender.indeterminate_sentence?
+      'purple'
+    else
+      'blue'
+    end
+  end
+
+  def badge_text(offender)
+    if offender.indeterminate_sentence?
+      'Indeterminate'
+    else
+      'Determinate'
+    end
+  end
+
+  def badge_for_parole?(offender)
+    offender.tariff_date.present? || offender.parole_eligibility_date.present? || parole_review_date?(offender)
+  end
+
+  def parole_review_date?(offender)
+    offender.indeterminate_sentence? && offender.parole_review_date.present?
+  end
+end
--- a/app/views/shared/_offence_info.html.erb
+++ b/app/views/shared/_offence_info.html.erb
@@ -52,7 +52,7 @@
     </tr>
     <tr class="govuk-table__row">
       <td class="govuk-table__cell">Parole Review date</td>
-      <td class="govuk-table__cell table_cell__left_align">
+      <td id="parole-review-date", class="govuk-table__cell table_cell__left_align">
         <% if @prisoner.indeterminate_sentence? %>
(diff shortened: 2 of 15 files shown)


app/helpers/badge_helper.rb
```diff
@@ -0,0 +1,21 @@
...
+  def badge_text(offender)
+    if offender.indeterminate_sentence?
+      'Indeterminate'
+    else
+      'Determinate'
+    end
+  end
+
+  def badge_for_parole?(offender)
+    offender.tariff_date || offender.parole_eligibility_date
```
starswan, line 19: kt's more idiomayic to use present? for this  rathet than treat dates as booleans

app/views/shared/_offence_info.html.erb
```diff
@@ -76,5 +76,4 @@
       </td>
     </tr>
   </tbody>
-</table>
```
starswan, line 79: need to remove - oops!

spec/factories/sentence_details.rb
```diff
@@ -2,17 +2,23 @@
...
     actualParoleDate { "2011-01-28" }
     licenceExpiryDate { "2011-01-28" }
+
+    trait :parole_eligibility_date do
+      paroleEligibilityDate { "2011-01-28" }
+    end
+
+    trait :tariff_date do
+      tariffDate { "2011-01-28" }
+    end
```
ollietreend, line 21: I can see you've moved `paroleEligibilityDate` and `tariffDate` into traits, rather than being regular attributes on this factory. What was the reason for this? I'm sure it's correct – just curious!

spec/factories/case_information.rb
```diff
@@ -30,4 +30,8 @@
 
     crn { Faker::Alphanumeric.alpha(number: 10) }
   end
+
+  trait :parole_review_date do
+    parole_review_date { Date.new(2019, 01, 3).to_s }
+  end
```
ollietreend, line 36: I was going to say that this probably doesn't need to be a trait, and could be just a regular field/attribute on the factory instead.

But seeing how the other fields have been moved into traits over in the `:sentence_details` factory, I'm sure there's a reason for it!

spec/factories/sentence_details.rb
Katy600: Hi @ollietreend I moved these two fields from being regular attributes to traits, because I realised when I created my new test, that checks to see if a parole eligible badge is being displayed if parole_review_date is present, this test passed before I had even created any test data. 
I then checked my two other tests that were also to do with displaying a parole eligible badge and realised that actually these tests were also both passing irrespective of how I set up my test data. In order for a parole eligible badge to display it needs either a parole_eligible_date, a tarriff_date, or a parole_review_date. So if just one of these is a regular attribute then all these tests will automatically pass. So it felt like having them as a regular attribute was undermining my tests. But I wasn't sure if making them a trait was the best practice or if their was a preferable way of doing it.

Katy600: @ollietreend I did worry when I was changing it if it was best practice!!

PR #1279, opened 2020-08-04
use faraday for VCR hooks to allow WebMock stub calls to work

starswan:
Currently when we wish to stub out Nomis calls to T3, they fail horribly rather than giving us a nice WebMock stubbing hook. This PR remedies this situation, so that  tests split nicely into VCR ones and pure-stubbed non-VCR ones.


spec/rails_helper.rb
```diff
@@ -80,6 +80,14 @@
       ActiveJob::Base.queue_adapter = adapter
     end
   end
+
+  WebMock.disable_net_connect!(allow_localhost: true)
+
+  config.around(:each, :vcr) do |example|
```
ollietreend, line 86: The significance of the `:vcr` condition on this `around()` block wasn't immediately clear to me – I didn't realise that it only applied only to VCR examples, rather than everything. Which made me question why we couldn't just `enable_net_connect!` globally.

A small comment might be helpful (to remind me next time I see this config...):

```suggestion
  # Allow HTTP requests when VCR is used
  config.around(:each, :vcr) do |example|
```

PR #1282, opened 2020-08-05
POM-784 update gateway and API hosts 

starswan:
This PR updates the authentication URL for NOMIS to reflect a name change.
However the current code assumes that the auth URL and API urls are the same, so this adds a new config item nomis_api_host. As a result I have updated the code to reflect the new endpoint i.e. removing the leading /elite2api everywhere and we are now using the new API host name (which had to be done by Sep 30th anyway)

After merging this PR, developers will either have to add /elite2api to the front of their existing gateway host, or update thie local configuration to reflect the new host names


spec/support/helpers/api_helper.rb
```diff
@@ -2,7 +2,7 @@
 
 module ApiHelper
   T3_HOST = Rails.configuration.nomis_oauth_host
-  T3 = "#{T3_HOST}/elite2api/api"
+  T3 = "#{Rails.configuration.nomis_api_host}/api"
```
ollietreend, line 5: The naming of these variables is a bit unclear now:

- `T3_HOST` represents the oAuth host
- `T3` represents the API host

Given their names, it isn't clear that the two represent different external services. Seems to be a hangover of the fact that they used to share a hostname, but now are seen as separate services.

---

Related, but probably a wider change: I think we should re-align our naming with the way the API is named now.

- 'Elite2' is now called the 'Prison API'
- The 'T3' environment is now called 'dev'

We have hardcoded references to 'Elite2' (e.g. `Nomis::Elite2::*` classes). And the use of `T3` variable name here is another example of that.

This probably extends to the environment variable / config keys – e.g. `NOMIS_API_HOST` could become `PRISON_API_HOST`

PR #1289, opened 2020-08-19
POM-608 recalls identified using recall field from NOMIS

starswan:
This PR makes 'recall' (from the prison search API) the definitive decision-maker for recalls


ollietreend (approved):
Looks great @starswan 🎉

Probably not for this PR, but now that we have the legal status field, I do wonder which field is the 'source of truth' for immigration cases:

`@imprisonment_status == 'DET'` or `@legal_status == 'IMMIGRATION_DETAINEE'`

🤔

app/models/nomis/offender_base.rb
```diff
@@ -114,20 +114,16 @@ def load_from_json(payload)
...
       # will have been given a payload at the class level, and will call this
       # method from it's own internal from_json
-      @first_name = payload['firstName']
-      @last_name = payload['lastName']
+      @first_name = payload.fetch('firstName')
+      @last_name = payload.fetch('lastName')
       @offender_no = payload.fetch('offenderNo')
       @convicted_status = payload['convictedStatus']
-      @sentence_type = SentenceType.new(payload['imprisonmentStatus'])
+      @sentence_type = SentenceType.new(payload['imprisonmentStatus'], payload['recall'])
```
ollietreend, line 121: Part of me questions whether 'recall' status is really part of the 'sentence type', or whether that was just a convenient place to put it when we'd predicted recall status based on the sentence code.

It's a question of semantics – is the offender recalled, or are they serving a recall sentence? I think it could be the former, in which case I think it could benefit from being refactored into its own first-class attribute on `OffenderBase`.

Looking at the codebase, it seems like we always use `OffenderBase#recalled?`. We never directly call `SentenceType#recall_sentence?` except from in `OffenderBase#recalled?` when delegating. We're just passing the value of `recall` from `OffenderBase` into `SentenceType`, just so we can pull it back out again.

app/services/nomis/elite2/offender_api.rb
```diff
@@ -30,23 +30,44 @@ def self.list(prison, page = 0, page_size: 20)
...
         else
-          api_deserialiser.deserialise(Nomis::Offender, response.first)
+          recall_route = "/offenders/#{url_offender_no}"
+          recall_response = Rails.cache.fetch(recall_route,
+                                              expires_in: Rails.configuration.cache_expiry) {
+            e2_client.get(recall_route)
+          }
+          # api_deserialiser.deserialise(Nomis::Offender, response.first)
+          api_deserialiser.deserialise(Nomis::Offender,
+                                       response.first.merge('recall' => recall_response.fetch('recall', false)))
```
ollietreend, line 70: One thing to note with this implementation is that we have two different 'sources of truth' for the recall field, depending on how you get offender(s):

- `.list` – get all offenders in a prison – uses the Prison Offender Search API
- `.get_offender` – get a single offender – uses the Prison API

Are we sure it's safe and sensible to make the assumption that the two APIs will be consistent in the recall status that they give? Can we treat them both as a 'source of truth'?

I'd be tempted to say they should both be backed by the Prison Offender Search API. Given the fact that we need to perform an additional API request anyway, there shouldn't be much difference between going to the Prison Offender Search API instead of the Prison API to get the recall status.

app/models/nomis/offender_base.rb
starswan: My worry is that we already have too much code in OffenderBase. Its a bit of a GOD class pattern. I wish we had smaller simpler classes (e.g. Sentence?) for the cases when we don't take decisions based on things like first_name. SentenceType used to answer this question (in fact I have given it the immigration_case? method recently, OffenderBase was doing it). Here you're right, it's unclear which is better - especially as the API is answering the question directly

starswan: I've gone with your suggestion. SentenceType is an anomaly causwed by us binding our data model to the NOMIS API - it's really an 'imprisonmentStatus decoder' which no longer has the recall? responsibility.

app/services/nomis/elite2/offender_api.rb
```diff
@@ -111,6 +118,15 @@ def self.get_image(booking_id)
...
     private
 
+      def self.get_recall_flags(offenders)
+        search_route = '/prisoner-numbers'
+        search_key = "#{search_route}_#{Digest::SHA256.hexdigest(offenders.to_s)}"
+        Rails.cache.fetch(search_key,
+                          expires_in: Rails.configuration.cache_expiry) {
+          search_client.post(search_route, prisonerNumbers: offenders)
+        }.index_by { |o| o.fetch('prisonerNumber') }
+      end
```
ollietreend, line 128: Just my opinion, but this doesn't need to be followed:

It might be nice for the `get_recall_flags` method to return ONLY a hash of offender number => recall flag.

This would mean the method is doing exactly what it says on the tin – i.e. getting recall flags, and nothing else.

And it might also have the side effect of make the data merge that happens above (in the `#list` and `#get_offender` methods above) a little more obvious / easier to follow.

ollietreend (commented):
Just a couple more questions from me, but otherwise this is looking great.

ollietreend (approved):
Looks great 🎉

PR #1292, opened 2020-08-25
fix sentence type for CRIM_CON

starswan:
This PR fixes the sentence_type for CRIM_CON. 
Note that this was a 'undetectable' bug, purely by fluke as we check type == INDETERMINATE so having a determinate sentence recorded as determinat didn't actually cause any practical problems.


ollietreend (approved):
Nice find 👍

PR #1296, opened 2020-08-28
POM-778 add :nocov: markers when coverage is uninteresting

starswan:
Although the goal of POM-778 was to introduce the 'undercover' gem, this PR is a useful first step.

It adds :nocov: markers in appropriate places, increasing the reported coverage to be 99.87%, missing only 4 lines which appear to have been missed during their implementation and so don'r warrant being hidden from view


Katy600 (approved):
I have no feedback!

PR #1309, opened 2020-09-04
Pom825 huntercombe broken due to 400 offenders

starswan:
Our pagination code had yet another defect hiding in it - when the nunber of offenders was an exact multiple of the page size, we went for one extra request. This used to be harmless, but after the movement fetching was introduced for ROTLs, it becomes dehabilitating as the request for movements for [] returns a 400 status from the movement API

### DIFF
--- a/spec/models/prison_spec.rb
+++ b/spec/models/prison_spec.rb
@@ -2,21 +2,73 @@ require 'rails_helper'
 
 RSpec.describe Prison, type: :model do
   describe '#offenders' do
-    let(:offenders) { described_class.new("LEI").offenders }
+    subject { described_class.new("LEI").offenders }
 
     it "get first page of offenders for a specific prison",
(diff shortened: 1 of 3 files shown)


ollietreend (approved):

spec/models/prison_spec.rb
```diff
@@ -2,21 +2,59 @@
...
+        stub_request(:get, "#{ApiHelper::T3}/locations/description/LEI/inmates?convictedStatus=Convicted&returnCategory=true").
+          with(
+            headers: {
+              'Page-Limit' => '1',
+              'Page-Offset' => '1',
+            }).
+          to_return(
+            body: {}.to_json,
+            headers: { 'Total-Records' => '200' }
+          )
```
ollietreend, line 36: Looks like we might not need this stub?

We don't do a 'Page-Limit' => '1', 'Page-Offset' => '1' request any more.

ollietreend (approved):
Approved, but with one comment

PR #1310, opened 2020-09-04
POM-829 offender Ids have to conform to 'A0000AA' format

starswan:
It seems that Prison API dev has just changed its error behaviour around non-conformant NOMIS IDs - it was allowing the and returning 404, now it seems that it is returning a 400 error


ollietreend (approved):

starswan:
The change that caused these failures has been rolled back https://dsdmoj.atlassian.net/browse/DT-1095

PR #1314, opened 2020-09-09
POM-831 search API not always returning same data as Prison API

starswan:
It seems that the Prisoner Search API doesn't always return the data it has been asked for - even if the list of offenders has just been returned from the Prison API.
This fixes this problem, by assuming that the recall flag is false in this case

### DIFF
--- a/app/services/nomis/elite2/offender_api.rb
+++ b/app/services/nomis/elite2/offender_api.rb
@@ -118,13 +118,14 @@ module Nomis
 
     private
 
-      def self.get_recall_flags(offenders)
+      def self.get_recall_flags(offender_nos)
         search_route = '/prisoner-numbers'
-        search_key = "#{search_route}_#{Digest::SHA256.hexdigest(offenders.to_s)}"
-        Rails.cache.fetch(search_key,
-                          expires_in: Rails.configuration.cache_expiry) {
-          search_client.post(search_route, prisonerNumbers: offenders)
-        }.map { |prisoner| [prisoner.fetch('prisonerNumber'), { 'recall' => prisoner.fetch('recall', false) }] }.to_h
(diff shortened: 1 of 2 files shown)


app/services/nomis/elite2/offender_api.rb
```diff
@@ -118,13 +118,14 @@ def self.get_image(booking_id)
...
-        Rails.cache.fetch(search_key,
-                          expires_in: Rails.configuration.cache_expiry) {
-          search_client.post(search_route, prisonerNumbers: offenders)
-        }.map { |prisoner| [prisoner.fetch('prisonerNumber'), { 'recall' => prisoner.fetch('recall', false) }] }.to_h
+        search_key = "#{search_route}_#{Digest::SHA256.hexdigest(offender_nos.to_s)}"
+        search_result = Rails.cache.fetch(search_key,
+                                          expires_in: Rails.configuration.cache_expiry) {
+                          search_client.post(search_route, prisonerNumbers: offender_nos)
+                        }.index_by { |prisoner| prisoner.fetch('prisonerNumber') }
+        offender_nos.index_with { |_nomis_id| { 'recall' => search_result.fetch('recall', false) } }
```
ollietreend, line 128: Are we missing another layer of `.fetch()` here?

It looks like we're trying to `.fetch('recall')` from the top-level `search_result` Hash, which is actually keyed by offender ID.

I was expecting to see something more like this:

```suggestion
        offender_nos.index_with do |nomis_id|
          offender = search_result.fetch(nomis_id, {})
          { 'recall' => offender.fetch('recall', false) }
        end
```

So we end up performing two `.fetch` calls to traverse the nested Hashes.

Right now, I'm pretty sure the results must all be coming back `false`, because the `.fetch()` will always fail to find the key `'recall'` in `search_results`.

ollietreend (approved):

PR #1322, opened 2020-09-14
Spike: push POM name and handover dates to nDelius via Community API

ollietreend:


ollietreend:
I'm closing this PR as this spike is no longer needed. We have moved forward with implementations in #1360 and #1363.

PR #1329, opened 2020-09-17
POM-768 Persuade Faraday client to retry on some errors

starswan:
Context:

It looks like our Faraday client is setup to retry on errors, but in practice it excluded Faraday:ClientError which is basically everything. This PR changes this situation so that we do a proper backoff retry 3 times for most things (apart from 404s) which I'm hoping will improve our reliability talking to the Prison API

### DIFF
--- a/app/services/hmpps_api/client.rb
+++ b/app/services/hmpps_api/client.rb
@@ -10,7 +10,6 @@ module HmppsApi
     def initialize(root)
       @root = root
       retry_options = {
-          # increase the default number of retries from 2 to 10 as 3 doesn't seem to be enough
           max: MAX_RETRIES,
           # some useful values as per the Faraday documentation
           # https://lostisland.github.io/faraday/middleware/retry
@@ -21,8 +20,8 @@ module HmppsApi
           backoff_factor: 2,
           # we seem to get 502 and 504 statuses from a gateway this side of
           # the Prison API - so retry if we get one of those.
-          retry_statuses: [502, 504],
-          exceptions: Faraday::Request::Retry::DEFAULT_EXCEPTIONS + [Faraday::ConnectionFailed],
+          #retry_statuses: [502, 504],
+          exceptions: Faraday::Request::Retry::DEFAULT_EXCEPTIONS + [Faraday::ConnectionFailed, Faraday::ServerError],
           # Faraday by default doesn't retry on a POST - even though our POSTs are GETs in disguise.
           methods: Faraday::Request::Retry::IDEMPOTENT_METHODS + [:post],
           #retry_if: ->(_env, _exception) { true },
--- a/app/services/hmpps_api/client.rb
+++ b/app/services/hmpps_api/client.rb
@@ -19,8 +19,8 @@ module HmppsApi
           interval_randomness: 0.5,
           backoff_factor: 2,
           # we seem to get 502 and 504 statuses from a gateway this side of
-          # the Prison API - so retry if we get one of those.
-          #retry_statuses: [502, 504],
+          # the Prison API - cso retry if we get one of those.
+          retry_statuses: [502, 504],
           exceptions: Faraday::Request::Retry::DEFAULT_EXCEPTIONS + [Faraday::ConnectionFailed, Faraday::ServerError],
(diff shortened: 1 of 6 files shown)


ollietreend (approved):

app/services/nomis/client.rb
```diff
@@ -13,7 +13,11 @@ def initialize(root, log404 = true)
       @connection = Faraday.new do |faraday|
         faraday.request :retry, max: 3, interval: 0.05,
                                 interval_randomness: 0.5, backoff_factor: 2,
-                                exceptions: [Faraday::ClientError, 'Timeout::Error']
+                                #  this looks like a strange list - client errors are unlikely to go away
+                                exceptions: [Faraday::ClientError, 'Timeout::Error'],
+                                # we seem to get 502 and 504 statuses from a gateway this side of
+                                # the Prison API - so retry if we get one of those.
+                                retry_statuses: [502, 504]
```
ollietreend, line 20: Would it make sense to go with something more like this?

```suggestion
                                exceptions: [
                                  Faraday::ClientError, # retry on 4xx errors
                                  Faraday::ServerError, # retry on 5xx errors
                                  Errno::ETIMEDOUT, Timeout::Error, Faraday::TimeoutError # retry on network timeouts
                                ],
```

i.e.
- retry 4xx responses
- retry 5xx responses (which covers 502 and 504)
- the final 3 exception types (`Errno::ETIMEDOUT`, `Timeout::Error`, `Faraday::TimeoutError`) mean that we retry on network-related exceptions, as per [the default config for retries](https://lostisland.github.io/faraday/middleware/retry#specify-which-exceptions-should-trigger-a-retry:~:text=that%20will%20replace%20the%20default%20list,of%20network%2Drelated%20exceptions%3A%20Errno%3A%3AETIMEDOUT%2C%20Timeout%3A%3AError%2C%20Faraday%3A%3ATimeoutError) which is otherwise overridden when we set this option.

starswan: I'm not sure we want to retry on all 500-type responses, just those that we know come from the gateway between us and the Prison API. I've added another commit so that we only add the 502 and 504 statuses, and most other stuff is defaulted.

ollietreend (approved):

ollietreend:
@starswan it seems weird that this PR has [failed in Circle][1], due to the very HTTP 504 error that it's trying to resolve 😕

Any idea why this might be? 😬 It suggests that this PR isn't working how we want it to – or that the 504 still stands even after 10 retry attempts.

Perhaps we should increase the wait time between retries? Or just merge this in as-is in the hope that it makes these sorts of errors _less likely_ to affect us?

**Note:** I can't actually see any evidence in the [failed Circle job][1] logs to indicate that it really _did_ retry the request 10 times. I'm just assuming the config is working the way we expect.. 🤷‍♂️ 

Maybe we should implement a `retry_block` to visibly log every time a retry happens – so at least it might be more obvious? See ['Call a block on every retry'](https://lostisland.github.io/faraday/middleware/retry#call-a-block-on-every-retry:~:text=You%20can%20specify%20a%20block%20through,every%20time%20the%20request%20is%20retried.) in the docs.

[1]: https://app.circleci.com/pipelines/github/ministryofjustice/offender-management-allocation-manager/6879/workflows/61be0054-dad1-487b-b069-81e68256c49a/jobs/24696

app/services/hmpps_api/client.rb
```diff
@@ -5,16 +5,40 @@
...
-        faraday.request :retry, max: 3, interval: 0.05,
-                                interval_randomness: 0.5, backoff_factor: 2,
-                                exceptions: [Faraday::ClientError, 'Timeout::Error']
-
         faraday.options.params_encoder = Faraday::FlatParamsEncoder
-        faraday.use Faraday::Response::RaiseError
-        faraday.use Faraday::Request::Instrumentation
+        # This is adding 2 pieces of request middleware and 1 piece of response middleware
+        faraday.request :retry, retry_options
+        faraday.request :instrumentation
```
ollietreend, line 40: Any idea what the `:instrumentation` middleware is used for?

I'm interested whether we're actually making use of it or not.

starswan: It's use to display request times

PR #1330, opened 2020-09-18
Pom 805 bug allocation reallocation 2

Katy600:
When a prisoner is deallocated at a prison due to transfer the next allocation should be: ALLOCATION (not re-allocation as is now) So when there is a deallocation - the next step should be Allocation and then Re-allocation until the De-allocation of a case when the journey starts again. 

This bug was discovered and corrected in the Override Controller. However, this bug created bad data that could not be altered with the bug fix. To get around this the data has been modified at the view level to ensure that allocation is always displayed when an offender has been transferred to a new prison.

### DIFF
--- a/app/views/allocations/history.html.erb
+++ b/app/views/allocations/history.html.erb
@@ -8,7 +8,8 @@
       <h2 class="govuk-heading-m"><%= prison_title(prison) %></h2>
       <div class="moj-timeline">
           <% allocations.each_with_index do |allocation, i| %>
-            <% if allocation.event == 'allocate_primary_pom' %>
+<!-- It is size-1 because the list is displayed backwards -->
+            <% if allocation.event == 'allocate_primary_pom' || i == allocations.size-1 %>
               <%= render :partial => "allocate_primary_pom",
                          :locals => { allocation: allocation,
                                       pom_name: allocation.primary_pom_name,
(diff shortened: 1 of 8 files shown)


app/views/allocations/history.html.erb
```diff
@@ -8,7 +8,8 @@
       <h2 class="govuk-heading-m"><%= prison_title(prison) %></h2>
       <div class="moj-timeline">
           <% allocations.each_with_index do |allocation, i| %>
-            <% if allocation.event == 'allocate_primary_pom' %>
+<!-- It is size-1 because the list is displayed backwards -->
+            <% if allocation.event == 'allocate_primary_pom' || i == allocations.size-1 %>
```
ollietreend, line 12: `i == allocations.size-1` might be more readable if written as:

```ruby
allocation == allocations.last
```

i.e. "is the current allocation the last one in the array?"

This works because of the [Ruby method `.last`](https://ruby-doc.org/core-2.6.3/Array.html#method-i-last), which returns the last element of an Array.

ollietreend (approved):
@Katy600 this looks great 🎉

I've added one comment which I think might help the readability of the code – see what you think.
But I've approved this PR anyway, because it does look as it is.

app/views/allocations/history.html.erb
Katy600: @ollietreend thanks Ollie. Yes that is much more readable.

PR #1332, opened 2020-09-22
Pom571 allow SPO to remove a Responsibilty override

starswan:
Currently overrides can be added (only custody -> community) but there is no way of removing them.

This change allows an SPO to remove a 'community' override with a mandatory reason.


spec/forms/remove_responsibility_form_spec.rb
```diff
@@ -0,0 +1,14 @@
+require 'rails_helper'
+
+RSpec.describe RemoveResponsibilityForm, type: :model do
+  let(:offender_id) { 'GA1234G' }
+
+  context 'with the text' do
```
ollietreend, line 6: Typo: should read `'without the text'`

ollietreend (approved):
Looks good @starswan

PR #1333, opened 2020-09-24
Rename API classes to reflect new API names

ollietreend:
This pull request replaces references to `Nomis` and `Elite2` in code which speaks to HMPPS APIs.

### In the beginning there was 'Elite2'

Historically, we would retrieve prisoner data from NOMIS using the Elite2 API. Class and module names were built around those API names.

For example, to get the details of an offender from the API, we'd call:
```ruby
Nomis::Elite2::OffenderApi.get_offender('A1234BC')
```

### 'Elite2' is now called 'Prison API'

A recent change within HMPPS means that the 'Elite2' API is now known as the 'Prison API'.

This pull request renames classes and modules so that we'd now call:

```ruby
HmppsApi::PrisonApi::OffenderApi.get_offender('A1234BC')
```

This better reflects the hierarchy of APIs used within our service.

### Overview of changes

- `Nomis` → `HmppsApi` – this top-level name encapsulates all HMPPS APIs that our service consumes (Prison API, Keyworker API, _soon Community API_)
- `Nomis::Client` → `HmppsApi::Client` – this client is used to make authenticated requests to _any_ HMPPS API, regardless of if the data comes from NOMIS or somewhere else (e.g. it can speak to the Community API to pull data from nDelius)
- `Nomis::Elite2` → `HmppsApi::PrisonApi` – Elite2 is now called Prison API

### Driver for this change

Whilst it isn't strictly necessary for our naming conventions to match those in the 'outside world', it would be confusing for a new developer working on our codebase. The `Nomis::Elite2` name only really works if you have the context – i.e. if you know that the Prison API used to be called Elite2. Therefore, this change helps us make our codebase _more obvious_ for a new starter.

Additionally, we want to begin integrating with other APIs (in particular, the Community API). The `Nomis::` naming convention falls apart when consider that the Community API has _nothing to do_ with NOMIS, but still needs to use the shared HMPPS API client at `Nomis::Client` (now: `HmppsApi::Client`).

In summary, this change makes our internal naming conventions:
- more obvious for new starters
- more consistent with the rest of HMPPS Digital
- more true to their meaning/purpose (e.g. `HmppsApi::Client` to speak to any HMPPS API)


app/services/hmpps_api/prison_api/prison_api_base.rb
```diff
@@ -1,18 +1,18 @@
 # frozen_string_literal: true
 
-module Nomis
-  module Elite2
+module HmppsApi
+  module PrisonApi
     ApiPaginatedResponse = Struct.new(:total_pages, :data)
 
-    module Elite2Api
+    module PrisonApiBase
```
starswan, line 7: personally I wouldn't call this PrisonApiBase because 
(a) it's a mixin 
(b) The term 'base' implies some magic base class

maybe PrisonApiClient?

starswan (approved):

PR #1335, opened 2020-09-25
Add a Community API client for pushing data into nDelius

ollietreend:
This Pull Request adds a new class (`HmppsApi::CommunityApi`) which is capable of pushing POM name and handover dates into nDelius.

This functionality has been pulled from the spike PR #1322 and rebased against the `main` branch.

It is a dependency of tickets POM-81 and POM-788, both of which are concerned with pushing data into nDelius.

### Set POM name

Set the name of the POM for an offender using the `set_pom` method.

For example, setting POM name to 'John Smith' for offender 'A1234BC' in prison 'LEI':

```ruby
HmppsApi::CommunityApi.set_pom(
  offender_no: 'A1234BC',
  prison: 'LEI',
  forename: 'John',
  surname: 'Smith'
)
```

**Note:** although the offender number is unique across all prisons, the prison code is still required as part of the API call. This is an implementation detail of how the POM name is added to nDelius – it creates a 'staff member' within that prison in nDelius.

### Set handover dates

Set the handover dates for an offender using the `set_handover_dates` method.

For example, to set handover dates for offender 'A1234BC':

```ruby
HmppsApi::CommunityApi.set_handover_dates(
  offender_no: 'A1234BC',
  handover_start_date: Date.today + 1.month,
  responsibility_handover_date: Date.today + 8.months
)
```

**Note:** Passing `nil` values for these handover dates will delete them from nDelius.


app/services/hmpps_api/community_api.rb
```diff
@@ -0,0 +1,58 @@
...
+          surname: surname
+        }
+      }
+
+      # We don't use a cache here because we're not 'getting' data, but 'pushing' it
+      self.client.put(route, body)
+
+      # So long as the API call didn't error, return true
+      # The API response body isn't useful to us
+      true
```
starswan, line 27: unsure about always returning true here - implies that you can somehow check the response for true/false. Probably more idiomatic not to return anything - it will probably be 'truthy' but clients shouldn't rely on it.

app/services/hmpps_api/community_api.rb
```diff
@@ -0,0 +1,58 @@
...
+        else
+          # Create/update the date in nDelius
+          body = { date: date.strftime('%F') }
+          self.client.put(route, body)
+        end
+      end
+
+      # So long as the API calls didn't error, return true
+      # The API response bodies aren't useful to us
+      true
```
starswan, line 55: see above

app/services/hmpps_api/community_api.rb
```diff
@@ -0,0 +1,58 @@
...
+    end
+
+    def self.set_handover_dates(offender_no:, handover_start_date:, responsibility_handover_date:)
+      safe_offender_no = URI.encode_www_form_component(offender_no)
+      base_route = "/offenders/nomsNumber/#{safe_offender_no}/custody/keyDates/"
+
+      # Map dates to the correct `{typeCode}` for the Community API
+      dates = {
+        POM1: handover_start_date,
```
starswan, line 36: POM1 and POM2 are awful. IMHO we should codify them as 'nicer' constants, so that the mapping is in the constants rather than the code.

spec/services/hmpps_api/community_api_spec.rb
```diff
@@ -0,0 +1,137 @@
...
+        responsibility_handover_date: responsibility_handover_date
+      )
+    }
+
+    let(:stub_base_url) { "#{api_host}/secure/offenders/nomsNumber/#{offender_no}/custody/keyDates" }
+
+    before do
+      stub_request(:put, "#{stub_base_url}/POM1").to_return(status: 200, body: '{}')
+      stub_request(:put, "#{stub_base_url}/POM2").to_return(status: 200, body: '{}')
+      stub_request(:delete, "#{stub_base_url}/POM1").to_return(status: 200)
```
starswan, line 58: do the actual delete requests return 200 like this?

spec/services/hmpps_api/community_api_spec.rb
```diff
@@ -0,0 +1,137 @@
...
+      stub_request(:delete, "#{stub_base_url}/POM2").to_return(status: 200)
+
+      # Trigger the request
+      subject
```
starswan, line 62: I think this is a slight misuse of 'subject'. IMHO the 'subject' block should just be here explicitly in the before block

spec/services/hmpps_api/community_api_spec.rb
```diff
@@ -0,0 +1,137 @@
...
+        expect(WebMock).to have_requested(:delete, "#{stub_base_url}/POM2")
+      end
+    end
+
+    describe 'when handover start date is nil' do
+      let(:handover_start_date) { nil }
+      let(:responsibility_handover_date) { Date.parse('25/12/2021') }
+
+      it 'deletes handover start date' do
+        # Handover start date = POM1
```
starswan, line 104: I think this is a hint that we need some nice constants (see above)

spec/services/hmpps_api/community_api_spec.rb
ollietreend: Fixed

app/services/hmpps_api/community_api.rb
ollietreend: Fixed – I think.

@starswan is this: 18093f876af3f57043d9cc43a7c81ca03de748df the sort of thing you had in mind?

spec/services/hmpps_api/community_api_spec.rb
ollietreend: Yes:
- According to the docs, it should return `200` on successful deletion
- I've just tried it against the test/dev Community API and got a `200` response

starswan (approved):

PR #1354, opened 2020-10-06
Refactor error handling in the HMPPS API client

ollietreend:
I've simplified the way errors are handled within our API client (which we use to speak to HMPPS APIs, such as the Prison API, Keyworker API, etc.)

Previously, the client would rescue errors from Faraday (the HTTP client that we use), and raise every error as a generic `HmppsApi::APIError` error. This meant that the true cause of an error was obscured from us, such that we weren't able to determine which HTTP status the error was, or whether it was a 'client error' (4xx) or 'server error' (5xx).

With this refactor, I've removed the `APIError` custom error class, and instead the API client will simply allow Faraday exceptions to bubble up and be caught higher up in the code.

This has also led me to remove the `log404 = true` parameter from the `HmppsApi::Client` initialise method. Since 404 errors now 'bubble up' as `Faraday::ResourceNotFound` errors, it is the responsibility of the caller to either: rescue & handle the error, or allow it to continue bubbling up through the application. All uncaught exceptions will be recorded in Sentry.

This has allowed me to simplify the RSpec tests, because the spec tag `:raven_intercept_exception` is no longer needed. I've also been able to remove the custom exception handler `AllocationManager::ExceptionHandler` as this is no longer used.

In summary:

- Faraday errors will now be raised when an API request fails. This replaces the `HmppsApi::APIError` class.
- Remove unneeded tag `:raven_intercept_exception` from RSpec tests
- Remove unused class `AllocationManager::ExceptionHandler`
- Remove unneeded option `log404` from API client `HmppsApi::Client.new`


app/services/hmpps_api/client.rb
```diff
@@ -7,9 +7,8 @@ module HmppsApi
   class Client
     APIError = Class.new(StandardError)
```
starswan, line 8: needs to be removed now?

starswan (approved):
One little thing needs removing

app/services/hmpps_api/client.rb
ollietreend: Thanks @starswan. Removed.

PR #1358, opened 2020-10-07
POM-834 map unallocated and inactive COM names to nil

starswan:
nDelius produces quite a few 'fake' COM names in it's data extract. MOIC need to know real soon whether a COM has been allocated or not. 
This PR fixes up the mapping so that inactive and unallocated COM names are mapped to nil in MOIC


ollietreend (approved):
Looks good

PR #1360, opened 2020-10-07
Pom 81 push pom names into api

Katy600:
MOiC holds the ‘single source of truth’ for the POM name at any one time. This PR pushes any changes to the POM name into Ndelius. 
The user need is for this name to automatically populate in Ndelius throughout the sentence to inform the COM and community staff of this name. 
Once they have been allocated a case for handover purposes they will need to make contact with the POM to start the handover process. 

In order to do this:
1. Creates a callback in the allocations database so any changes to pom_name are then queued in the PushPomToDeliusJob.
2. Creates a global stub that prevents this callback to trigger, adding this as a tag on tests which make changes in the allocation database
3. The PushPomToDeliusJob checks these new allocations to see if the POM names are present or nil and passes this on to the Community Api via their unset_pom and set_pom methods
4. The unset_pom method on the community api implements the delete route for the community api


app/jobs/push_pom_to_delius_job.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+class PushPomToDeliusJob < ApplicationJob
+  queue_as :default
+
+  def perform(allocation)
+    # We can't yet 'unset' the allocated POM in nDelius
```
starswan, line 7: this comment is out of date...

app/models/allocation.rb
```diff
@@ -156,4 +159,18 @@ def prison_fix(movement_type)
             :event,
             :event_trigger,
             :prison, presence: true
+
+private
+
+  def check_push
```
starswan, line 165: This could do with a comment explaining why the push is done is 2 pieces

spec/jobs/push_pom_to_delius_job_spec.rb
```diff
@@ -0,0 +1,54 @@
+require 'rails_helper'
+
+RSpec.describe PushPomToDeliusJob, :allocation, type: :job, versioning: true do
+  let(:offender) { build(:nomis_offender) }
+  let(:offender_no) { offender.fetch(:offenderNo) }
+  let(:pom) { build(:pom) }
+
+  before do
+    stub_auth_token
+    stub_community_set_pom(offender)
```
starswan, line 10: This line (stub_community_set_pom) looks wrong at this level as it is intercepted in tests

spec/models/allocation_spec.rb
```diff
@@ -271,4 +271,59 @@
       end
     end
   end
+
+  describe 'automate pushing the primary pom to ndelius' do
+    let(:prison) { build(:prison).code }
+
+    context 'when a new allocation is created and a POM is set' do
+      before do
+        expect(PrisonOffenderManagerService).to receive(:get_pom_name).with(nomis_staff_id).and_return ['Bill', 'Jones']
```
starswan, line 280: The pom name is stored in the allocation, so we shouldn't need to fetch it again

spec/support/helpers/community_api_helper.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module CommunityApiHelper
+  COMMUNITY_API_HOST = Rails.configuration.community_api_host
+
+  def stub_community_set_pom(offender)
```
starswan, line 6: If my comment above is correct, this module now has no purpose

app/models/allocation.rb
ollietreend: We should be able to use Rails' Dirty Tracking in the `after_save` without needing to capture the dirty state in a `before_save`.

Rather than using `primary_pom_nomis_id_changed?`, we can use `saved_change_to_primary_pom_nomis_id?` to tell us whether the most recent save changed the `primary_pom_nomis_id`.

This would allow us to simplify the implementation by removing the `before_save` callback and `@do_push` instance variable, and instead perform the conditional check directly in the `after_save` callback / `push_pom_to_community_api` method.

See the docs for more info: https://api.rubyonrails.org/classes/ActiveRecord/AttributeMethods/Dirty.html#method-i-saved_change_to_attribute-3F

@Katy600 it would be good to give this a try, as it feels cleaner than using two callbacks. If not possible, then a comment to explain why it's done in 2 pieces is a good idea (as suggested by @starswan).

spec/models/allocation_spec.rb
ollietreend: @starswan The only risk I see with using the 'cached' `pom_name` field is this:

When saved into the `pom_name` field of an Allocation record, the POM name is transformed into a single string.
e.g. API response `['Bill', 'Jones']` is saved as string `"Jones, Bill"`

But the Community API requires the POM name to be supplied as a pair of `[forename, surname]` – which fits perfectly with the response we receive from the `PrisonOffenderManagerService`.

To use the `pom_name` field value, we'd need to perform the opposite transformation to split the `"Surname, Forename"` string back into an array.

It's _probably safe_ to reverse that transformation _in most cases_. But I wonder if there will ever be a case where one of the names contain a comma (`,`) which breaks this reverse transformation.

e.g. given an original name:

```ruby
["John", "Jones,Smith"]
```

our `pom_name` field would have value:

```ruby
"Jones,Smith, John"
```

and we'd struggle to reverse the transformation.

It could end up as:

```ruby
["Smith, John", "Jones"]
```

I have no idea how likely this scenario is, or if it's even possible for a POM name to contain a comma. I just want to point out the risk of us attempting to un-transform the POM name field to fit our needs here.

lib/tasks/batch_push_pom_to_delius.rake
```diff
@@ -0,0 +1,6 @@
+namespace :batch_push_pom_to_delius do
+  desc 'batch push pom to delius job'
+  task populate_delius: :environment do
+    Allocation.where.not(primary_pom_nomis_id: nil) { |a| PushPomToDeliusJob.perform_later(a) }
```
ollietreend, line 4: In production, this query would return tens of thousands of records.

We should really be using `.find_each` here, to get these records in batches. This will avoid us attempting to load all records at once and exhausting our available memory.

More info about reading records in batches: https://apidock.com/rails/ActiveRecord/Batches/find_each

You can see an example of where batching is performed in another rake task here: https://github.com/ministryofjustice/offender-management-allocation-manager/blob/main/lib/tasks/delius_trigger.rake#L9-L11

ollietreend (approved):

PR #1361, opened 2020-10-08
Use new nomis role

starswan:
The 'Elephants' team need a role to allow them to perform admin functions on the service that no ordinary user is permitted to have. This role MOIC_ADMIN wasa created some time ago, but not assigned to anyone. This PR uses that role so that we can be considered to be admins in staging where we only have access to 3 prisons.

### DIFF
--- a/app/controllers/concerns/sso_identity.rb
+++ b/app/controllers/concerns/sso_identity.rb
@@ -1,4 +1,10 @@
+# frozen_string_literal: true
+
 class SsoIdentity
+  POM_ROLE = 'ROLE_ALLOC_CASE_MGR'
+  SPO_ROLE = 'ROLE_ALLOC_MGR'
+  ADMIN_ROLE = 'ROLE_MOIC_ADMIN'
+
   def initialize(session)
     @sso_identity = session[:sso_data]
   end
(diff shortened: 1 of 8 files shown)


app/controllers/concerns/sso_identity.rb
```diff
@@ -16,17 +22,17 @@ def current_user
   end
 
   def current_user_is_spo?
-    roles.include?('ROLE_ALLOC_MGR')
+    roles.include?(SPO_ROLE)
   end
 
+  # This role is granted to the service team to give them 'admin' priviledges
+  # such as editing teams and changing switch values
   def is_global_admin?
```
ollietreend, line 30: Could this be a more consistent name for the admin method?

```suggestion
  def current_user_is_admin?
```

Given that we have the POM and SPO methods which already set out a convention of:
```
current_user_is_XXX?
```

Also note that I've suggested dropping 'global' from the name. I'm not sure if the addition of 'global' adds much more meaning to the 'admin' role name. And given the name of the constant `ADMIN_ROLE`, I'd rather we kept it consistent and simply went with 'admin'.

spec/support/helpers/features_helper.rb
```diff
@@ -2,15 +2,15 @@
 
 module FeaturesHelper
   def signin_spo_user(name = 'MOIC_POM')
-    mock_sso_response(name, ['ROLE_ALLOC_MGR'])
+    mock_sso_response(name, [SsoIdentity::SPO_ROLE])
   end
 
   def signin_global_admin_user
```
ollietreend, line 8: The same thing applies here about dropping 'global' from the name.

```suggestion
  def signin_admin_user
```

I think we can safely refer to this type of user as an "admin user".

app/controllers/concerns/sso_identity.rb
starswan: Problem is that we have an 'ensure_admin_user' method that checks if you are an SPO....

ollietreend (approved):

PR #1362, opened 2020-10-08
Pom 819 - Fix bugs and refactor OffenderPresenter

theKHutDeveloper:
The two bugs (**POM-802 and POM-803**) are in relation to Responsibility Overrides not being taken into account when calculating responsibility.

**POM-802:** When the responsibility has been overridden this does not change the POM recommendation

**_Screenshot of fix_**
Screen Shot 2020-09-16 at 13 44 31

**POM-803:** When the responsibility has been overridden prior to allocation, the 'Make Allocations' page does not reflect the change in the 'case owner' column.

**_Screenshot of fix_**
Screen Shot 2020-09-16 at 13 44 57

**Refactoring OffenderPresenter**
The current implementation of responsibility is handled in a few places. In particular, the OffenderPresenter has a method called pom_responsibility which will override the ‘calculated’ responsibility with the ‘manually overridden’ responsibility, if one is present.

Ideally we want to have a single source of truth for the responsibility calculation

In refactoring a couple of tests have been removed from `prison_allocations_spec.rb` as the functionality has been tested elsewhere.


app/models/hmpps_api/offender_base.rb
```diff
@@ -19,7 +19,7 @@ class OffenderBase
 
     attr_reader :crn,
                 :welsh_offender, :parole_review_date,
-                :ldu, :team
+                :ldu, :team, :responsibility
```
starswan, line 22: Think this change isn't needed?

app/presenters/offender_presenter.rb
```diff
@@ -6,7 +6,7 @@ class OffenderPresenter
            :full_name_ordered, :full_name, :main_offence,
            :sentence_start_date, :team, :prison_id,
            :home_detention_curfew_eligibility_date, :home_detention_curfew_actual_date,
-           :tariff_date,
+           :tariff_date, :responsibility,
```
starswan, line 9: think this might be a merge issue? Don't think it's needed

app/presenters/offender_presenter.rb
```diff
@@ -16,28 +16,16 @@ class OffenderPresenter
...
-      end
-    else
-      @offender.pom_responsibility
-    end
+    @offender.pom_responsibility
   end
 
   def responsibility_override?
-    @responsibility.present?
+    @offender.responsibility.present?
```
starswan, line 28: would probably be nicer just to forward this method to @offender

starswan (commented):
Looks good - just a couple of minor comments. Can't make up my mind whether to hit approve or not...

app/models/hmpps_api/offender_base.rb
theKHutDeveloper: @starswan Tests fail without this

app/presenters/offender_presenter.rb
theKHutDeveloper: @starswan I don't understand what you mean? Are you suggesting that I delegate the responsibility_override? method to the offender object?

spec/services/case_information_service_spec.rb
```diff
@@ -17,6 +17,13 @@
     expect(cases[caseinfo.nomis_offender_id]).to be_kind_of(CaseInformation)
   end
 
+  it 'can eager load the responsibility association' do
+    create(:responsibility, nomis_offender_id: caseinfo.nomis_offender_id)
+    cases = described_class.get_case_information([caseinfo.nomis_offender_id])
+
+    expect(cases[caseinfo.nomis_offender_id].association(:responsibility).loaded?).to eq(true)
```
ollietreend, line 24: This is really cool! I didn't know we could test that associations are eager loaded.

app/models/hmpps_api/offender_base.rb
starswan: see comment further down - we shouldn't be exposing this object, we should be exposing responsibility_override? instead

app/presenters/offender_presenter.rb
starswan: yes

starswan (approved):

PR #1363, opened 2020-10-08
POM-788: Push offender handover dates into nDelius

ollietreend:
Jira ticket: POM-788

---

This pull request adds functionality to push offender handover dates into nDelius using the Community API.

It also adds a new concept of 'calculated handover dates' – these are effectively cached calculations of offender handover dates. We use these to detect when handover dates have changed, and therefore whether we should push them to the Community API. It means we won't push handover dates into the Community API if they haven't changed since their last push.

Summary of changes:
- Added `RecalculateHandoverDateJob`: this job can be queued and will recalculate handover dates for a given offender, and push them into the Community API if they've changed
- Added rake task `rake recalculate_handover_dates`: this will run through all known offenders (all `CaseInformation` records) and queue up `RecalculateHandoverDateJob` jobs for each one
- The nightly nDelius import will recalculate and push handover dates when offender data changes
- Added new model `CalculatedHandoverDate` with database table `calculated_handover_dates`

## How to use

To push handover dates into nDelius, run the following rake task:

```
rake recalculate_handover_dates
```

This is a potentially expensive / long-running command: it will perform lots of calls to both the Prison API and Community API.

To avoid being 'noisy' to the APIs, we probably shouldn't run it too frequently.

## Error handling

Pushing dates into nDelius could fail for either of the following reasons:

### 1. The offender doesn't exist in NOMIS

**Scenario:** The nightly nDelius import has run and imported a new first-time offender, who has been sentenced but hasn't arrived in prison yet. Therefore, there is no record for them in NOMIS yet. A `RecalculateHandoverDateJob` job has been queued because a new Case Information record was created.

**Behaviour:** The `RecalculateHandoverDateJob` job cannot find a record of the offender in NOMIS. It returns a nil value and the job ends successfully. It is not re-queued.

### 2. The offender doesn't exist in nDelius

**Scenario:** The rake task has been run to recalculate handover dates for all known offenders. For some offenders, nDelius cannot find their record because the "DSS" button has not been used, meaning their NOMIS ID is not associated with their nDelius record. Alternatively, they might be Scottish or Northern Irish offenders, in which case no nDelius record will exist for them.

**Behaviour:** The `RecalculateHandoverDateJob` attempts to push to the Community API. The Community API responds with a '404 Not Found' HTTP error, which is thrown as an exception in our application. The exception is caught and logged to Sentry, and the job is marked as 'failed'. Failed jobs are [automatically retried by Sidekiq](https://github.com/mperham/sidekiq/wiki/Error-Handling).

### DIFF
--- a/app/jobs/process_delius_data_job.rb
+++ b/app/jobs/process_delius_data_job.rb
@@ -34,7 +34,7 @@ private
     # as a compromise, we always import the DeliusData into the case_information record now,
     # but only disable manual editing for prisons that are actually enabled.
     if dob_matches?(offender, delius_record)
-      process_record(delius_record)
+      process_record(offender, delius_record)
     else
       DeliusImportError.create! nomis_offender_id: delius_record.noms_no,
                                 error_type: DeliusImportError::MISMATCHED_DOB
@@ -53,10 +53,16 @@ private
     nil
   end
 
-  def process_record(delius_record)
+  def process_record(offender, delius_record)
     case_information = map_delius_to_case_info(delius_record)
 
-    unless case_information.save
+    if case_information.save
+      # Update the offender object with the new Case Information data
+      offender.load_case_information(case_information)
+
+      # Recalculate the offender's handover dates
+      CalculatedHandoverDate.recalculate_for(offender)
+    else
       case_information.errors.each do |field, _errors|
         DeliusImportError.create! nomis_offender_id: delius_record.noms_no,
                                   error_type: error_type(field)
@@ -69,13 +75,13 @@ private
       team = map_team(delius_record.team_code)
       case_info.assign_attributes(
         manual_entry: false,
-      com_name: map_com_name(delius_record.offender_manager),
-      crn: delius_record.crn,
-      tier: map_tier(delius_record.tier),
-    team: team,
-    case_allocation: delius_record.service_provider,
-    welsh_offender: map_welsh_offender(delius_record.ldu_code),
-    mappa_level: map_mappa_level(delius_record.mappa, delius_record.mappa_levels)
+        com_name: map_com_name(delius_record.offender_manager),
+        crn: delius_record.crn,
+        tier: map_tier(delius_record.tier),
+        team: team,
+        case_allocation: delius_record.service_provider,
+        welsh_offender: map_welsh_offender(delius_record.ldu_code),
+        mappa_level: map_mappa_level(delius_record.mappa, delius_record.mappa_levels)
       )
     end
   end
--- a/app/models/calculated_handover_date.rb
+++ b/app/models/calculated_handover_date.rb
@@ -13,12 +13,27 @@ class CalculatedHandoverDate < ApplicationRecord
   validates :nomis_offender_id, uniqueness: true, presence: true
   validates :reason, presence: true
 
+  after_save :push_to_delius
+
   def self.recalculate_for(offender)
     record = self.find_or_initialize_by(nomis_offender_id: offender.offender_no)
     record.update!(
       start_date: offender.handover_start_date,
       handover_date: offender.responsibility_handover_date,
       reason: offender.handover_reason
+    )
+  end
+
+private
+
+  def push_to_delius
+    # Only push if the dates have changed
+    return unless saved_change_to_start_date? || saved_change_to_handover_date?
+
+    HmppsApi::CommunityApi.set_handover_dates(
+      offender_no: nomis_offender_id,
+      handover_start_date: start_date,
+      responsibility_handover_date: handover_date
     )
   end
 end
--- a/app/models/calculated_handover_date.rb
+++ b/app/models/calculated_handover_date.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+class CalculatedHandoverDate < ApplicationRecord
+  # This is quite a loose relationship. It exists so that CaseInformation
+  # deletes cascade and tidy up associated CalculatedHandoverDate records.
+  # Ideally CalculatedHandoverDate would belong to a higher-level
+  # Offender model rather than nDelius Case Information
+  belongs_to :case_information,
+             primary_key: :nomis_offender_id,
+             foreign_key: :nomis_offender_id,
+             inverse_of: :responsibility
+
+  validates :nomis_offender_id, uniqueness: true, presence: true
+  validates :reason, presence: true
+end
--- a/app/models/calculated_handover_date.rb
+++ b/app/models/calculated_handover_date.rb
@@ -12,4 +12,13 @@ class CalculatedHandoverDate < ApplicationRecord
 
   validates :nomis_offender_id, uniqueness: true, presence: true
   validates :reason, presence: true
+
+  def self.recalculate_for(offender)
+    record = self.find_or_initialize_by(nomis_offender_id: offender.offender_no)
+    record.update!(
+      start_date: offender.handover_start_date,
+      handover_date: offender.responsibility_handover_date,
+      reason: offender.handover_reason
+    )
+  end
 end
--- a/db/migrate/20201023133523_create_calculated_handover_dates.rb
+++ b/db/migrate/20201023133523_create_calculated_handover_dates.rb
@@ -1,0 +1,21 @@
+class CreateCalculatedHandoverDates < ActiveRecord::Migration[6.0]
+  def change
+    create_table :calculated_handover_dates do |t|
+      t.date :start_date
+      t.date :handover_date
+      t.string :reason, null: false
+
+      t.timestamps
+
+      t.references :nomis_offender,
+                   type: :string, null: false,
+                   index: { unique: true },
+                   foreign_key: {
+                     to_table: :case_information,
+                     primary_key: :nomis_offender_id,
+                     on_delete: :cascade,
+                     on_update: :cascade
+                   }
+    end
+  end
+end
--- a/spec/jobs/process_delius_data_job_spec.rb
+++ b/spec/jobs/process_delius_data_job_spec.rb
@@ -2,7 +2,8 @@
 
 require 'rails_helper'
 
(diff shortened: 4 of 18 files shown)


ollietreend:
I've added the **'Do Not Merge'** label to this PR as it is not yet ready to be deployed to production.

It will need testing in pre-production against nDelius pre-prod before it's ready for production.

spec/jobs/process_delius_data_job_spec.rb
```diff
@@ -252,6 +252,49 @@
...
     end
+
+    describe 'pushing handover dates into nDelius' do
+      let!(:d1) {
+        create(:delius_data, tier: 'C', team_code: team.shadow_code, team: team.name,
+               ldu_code: ldu.code, ldu: ldu.name)
+      }
+
+      before do
+        ActiveJob::Base.queue_adapter = :test
```
starswan, line 263: oops - this test has the side-effect of changing the queue adapter...

spec/support/helpers/api_helper.rb
```diff
@@ -135,6 +135,12 @@ def stub_multiple_offenders(offenders, bookings)
       to_return(body: bookings.to_json)
   end
 
+  # Stub an 'empty' response from the Prison API, indicating that the offender does not exist in NOMIS
+  def stub_non_existent_offender(offender_no)
+    stub_request(:get, "#{T3}/prisoners/#{offender_no}").
```
starswan, line 140: is this really what the Prison API does? Does it not 404?

spec/rails_helper.rb
```diff
@@ -41,6 +41,11 @@
     page.driver.browser.manage.window.resize_to(1280,3072)
   end
 
+  config.before(:each) do
+    # Stop various test setups from indirectly attempting to push handover dates to the Community API
```
starswan, line 45: This seems too agressive - tagging the relevant ones (see POM-81) would be much less invasive

spec/jobs/push_handover_dates_to_delius_job_spec.rb
```diff
@@ -0,0 +1,52 @@
+require 'rails_helper'
+
+RSpec.describe PushHandoverDatesToDeliusJob, type: :job do
+  let(:nomis_offender) { build(:nomis_offender) }
+  let(:offender_no) { nomis_offender.fetch(:offenderNo) }
+
+  before do
+    stub_auth_token
+    allow(HmppsApi::CommunityApi).to receive(:set_handover_dates)
```
starswan, line 9: this is a very weak stub. IIMHO it should be a mock - expect().to receive(:set_handover_dates).with(<the correct parameters>)

app/jobs/push_handover_dates_to_delius_job.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+class PushHandoverDatesToDeliusJob < ApplicationJob
+  queue_as :default
+
+  def perform(nomis_offender_id)
+    offender = OffenderService.get_offender(nomis_offender_id)
```
starswan, line 7: It looks like this job does too much. Actually we had the offender object in our hands in the ProcessDeliusDataJob#import_data - seems a shame to go and get it again? Maybe this job doesn't need to exist (but I guess it does because of the rake task) - now I'm not sure...

spec/support/helpers/api_helper.rb
ollietreend: @starswan It feels odd, but yes: this is what the Prison API does.

The `/prisoners/{offenderNo}` endpoint returns an array of offenders matching the specified offender number. So in practice, we either return 0 or 1 elements in the array depending on if the offender was found or not.

This is documented in the [Prison API docs](https://api-dev.prison.service.justice.gov.uk/swagger-ui.html), but also I've verified that this is the actual behaviour of the API in production:

```
GET /api/prisoners/123abc (0.087 s)
=> []
```

lib/tasks/delius_push_handover_dates.rake
```diff
@@ -0,0 +1,25 @@
+namespace :delius do
+  desc 'Push handover dates into nDelius for all offenders across the estate'
+  task push_handover_dates: :environment do |_task|
+    Rails.logger = Logger.new(STDOUT)
+
+    prisons = PrisonService.prison_codes
```
starswan, line 6: you could just do PrisonService::PRISONS.each ? Not a biggie

starswan, line 6: seems a shame that you can't do Prison.all.each_with_index...

spec/jobs/process_delius_data_job_spec.rb
```diff
@@ -252,6 +252,49 @@
         expect(c1.reload.tier).to eq('C')
       end
     end
+
+    describe 'pushing handover dates into nDelius' do
+      let!(:d1) {
```
starswan, line 257: d1 is a bad name - probably not great to continue the bad pattern

app/jobs/process_delius_data_job.rb
```diff
@@ -62,6 +62,10 @@ def process_record(delius_record)
     end
 
     update_com_name(delius_record)
+
+    if case_information.saved_changes?
```
starswan, line 66: could we not just invert the 'unless' into an if - so it becomes if case.information.save <nDelius> else <create error>? Also inlining this process_record method might lead down the path of just calling nDelius inline as we have the offender object in our hands?

ollietreend:
@starswan Can I get you to re-review this please?

In response to your comments, I've made the following changes:

- `ProcessDeliusDataJob` now pushes directly to the Community API, rather than queuing up more jobs
- Replace global stub with spec tag `:disable_push_to_delius`, to stop test attempting to push handover dates to the Community API
- Add `Prison.all` convenience method

spec/support/helpers/api_helper.rb
starswan: I think this needs a clearer comment - and maybe moving into its own helper method as well just to enphasise the point. Despite the documentation, IMHO this is a violation of the 'principle of least surprise'

starswan (approved):

ollietreend:
@starswan I've changed this PR to introduce the new concept of 'Calculated Handover Dates'.

This will allow us to cache handover dates within the service, so we know not to re-push the same dates into nDelius over & over again.

Please could you re-review?

app/models/calculated_handover_date.rb
```diff
@@ -0,0 +1,36 @@
...
+             foreign_key: :nomis_offender_id,
+             inverse_of: :responsibility
+
+  validates :nomis_offender_id, uniqueness: true, presence: true
+  validates :reason, presence: true
+
+  after_save :push_to_delius
+
+  def self.recalculate_for(offender)
+    record = self.find_or_initialize_by(nomis_offender_id: offender.offender_no)
```
starswan, line 15: think find_or_initialize_by takes a block, so this could be a bit more idiomatic

app/models/prison.rb
```diff
@@ -38,6 +38,10 @@ def allocations_for(staff_id)
     }
   end
 
+  def self.all
```
starswan, line 41: as previously discussed, 'all' isn't a good thing to query on' as it included prisons which are not live

db/migrate/20201023133523_create_calculated_handover_dates.rb
```diff
@@ -0,0 +1,13 @@
+class CreateCalculatedHandoverDates < ActiveRecord::Migration[6.0]
+  def change
+    create_table :calculated_handover_dates do |t|
+      t.string :nomis_offender_id, null: false
```
starswan, line 4: it would be nice to maybe use rails conventions here - use the case information id rather than nomis_offender_id everywhere.

spec/models/calculated_handover_date_spec.rb
```diff
@@ -0,0 +1,161 @@
...
+      end
+
+      describe "when the dates haven't changed" do
+        let(:existing_start_date) { offender.handover_start_date }
+        let(:existing_handover_date) { offender.responsibility_handover_date }
+        let(:existing_reason) { offender.handover_reason }
+
+        it "does nothing" do
+          old_updated_at = existing_record.updated_at
+
```
starswan, line 108: tests can set updated_at, so there's no need to sleep - simply check that it hasn't changed

app/models/calculated_handover_date.rb
ollietreend: I did try this at first, but not sure it works for our needs here.

`find_or_initialize_by` only runs the in the case of 'initialize', whereas we want to update the attributes regardless of whether the record was 'found' or 'initialized'.

app/models/prison.rb
ollietreend: I'll remove the `Prison.all` method as it's now unused (as no longer used by the rake task).

spec/models/calculated_handover_date_spec.rb
ollietreend: I was following the approach used in the CaseInformation spec:
https://github.com/ministryofjustice/offender-management-allocation-manager/blob/72f6199ce652d8866bdd0daa06e2d272cfad9845/spec/models/case_information_spec.rb#L6-L11

But it did feel a little weird to sleep in the tests.

I've just changed it to use time travel instead. What do you think of this approach @starswan?

app/models/calculated_handover_date.rb
ollietreend: @starswan To make this method more idiomatic, I've just replaced the use of `assign_attributes` + `save!` with `update!` instead.

db/migrate/20201023133523_create_calculated_handover_dates.rb
ollietreend: @starswan I've changed this to use `references`, which also adds a foreign key.

starswan (approved):

PR #1365, opened 2020-10-14
POM-778 - use undercover gem for circle:ci test coverage

starswan:


ollietreend:
@starswan as per our discussion: it'd be a shame to lose the HTML formatted output.

I think this should allow us to output both formats: https://www.rubydoc.info/gems/simplecov/0.6.3#using-multiple-formatters

Might be worth trying to see if we can keep both?

ollietreend (approved):

ollietreend:
@starswan Just a thought:

The gem `rugged` (a dependency of `undercover`) won't install without `cmake` being available – which doesn't come preinstalled on macOS.

One nice addition to this PR would be to [add a `Brewfile` to the repo](https://thoughtbot.com/blog/brewfile-a-gemfile-but-for-homebrew) to explicitly mark `cmake` as a system dependency.

This would allow us to document homebrew packages which are dev dependencies – much the same as a `Gemfile` – and then install them with command:
```
brew bundle
```

(I'd also like to replace our `geckodriver-helper` gem dependency with the native `geckodriver` homebrew package at some point, too.)

starswan:
> @starswan Just a thought:
> 
> The gem `rugged` (a dependency of `undercover`) won't install without `cmake` being available – which doesn't come preinstalled on macOS.
> 
> One nice addition to this PR would be to [add a `Brewfile` to the repo](https://thoughtbot.com/blog/brewfile-a-gemfile-but-for-homebrew) to explicitly mark `cmake` as a system dependency.
> 
> This would allow us to document homebrew packages which are dev dependencies – much the same as a `Gemfile` – and then install them with command:
> 
> ```
> brew bundle
> ```
> 
> (I'd also like to replace our `geckodriver-helper` gem dependency with the native `geckodriver` homebrew package at some point, too.)

I introduced geckodriver-helper - just because it was a hidden dependency lurking on everyone's machines, and for PVB you had to be using a specific version otherwise it didn't work. Brewfiles look like a better solution to the same problem - and upgrading to MacOS catalina then becomes possible without forking geckodriver-helper

Brewfile.lock.json
```diff
@@ -0,0 +1,60 @@
...
+      "high_sierra": {
+        "HOMEBREW_VERSION": "2.5.5",
+        "HOMEBREW_PREFIX": "/usr/local",
+        "Homebrew/homebrew-core": "0410fbf1388d40189b73e5382573105f6a7022bf",
+        "CLT": "10.1.0.0.1.1539992718",
+        "Xcode": "10.1",
+        "macOS": "10.13.6"
+      }
+    }
+  }
```
ollietreend, line 59: Unlike `Gemfile.lock`, the `Brewfile.lock.json` is system specific so should really be added to `.gitignore`.

It's not very well documented in the Brewfile project, but they did hint towards that when adding the lock file functionality in Homebrew/homebrew-bundle#552

This also seems to be the behaviour followed by others, e.g. in apple/swift#30714 where they [added `Brewfile.lock.json` to the `.gitignore` file](https://github.com/apple/swift/pull/30714/commits/3d5352d3c8ccc4b917dcb5170966e835ec707c53).

### If we don't `.gitignore` it

I think we'll end up causing a lot of unnecessary noise in commits between developers on the project. Every time a developer runs `brew bundle` to install dependencies, the lockfile will be re-written to reflect their own local system.

ollietreend (approved):

PR #1368, opened 2020-10-15
POM-612: Make the 'case owner' column on 'Make Allocations' page sortable

theKHutDeveloper:
This PR is to address a user need that has been identified several times via user feedback.

SPOs prioritise allocations so that they allocate cases where custody are responsible first. But this is not easy to do currently as the user is not able to sort by case owner (the two states for case owner are 'Community' or 'Custody').

The outcome of this PR is that an SPO will be able to sort the 'Make Allocations' tab by case owner in ascending or descending order. 

**Before:**

Screen Shot 2020-10-15 at 12 22 31

**After:**

_Sorted in ascending order_

Screen Shot 2020-10-15 at 12 15 32

_Sorted in descending order_ 

Screen Shot 2020-10-15 at 12 16 00


spec/controllers/summary_controller_spec.rb
```diff
@@ -8,36 +8,8 @@
   before { stub_sso_data(prison) }
 
   context 'with 4 offenders' do
```
starswan, line 10: it would have been nice not to replicate this - I have no idea why these 4 offenders are required for this test. The context says it all (i.e. its hopeless) - it used to say 'with 2 offenders' which was worse. A new clean test was probably called for.

PR #1374, opened 2020-10-19
POM-486 ldu email 45 days from handover

starswan:
For handovers to proceed smoothly, MOIC needs to notify each LDU to tell them the offenders that are coming up for handover within the next 45 days


theKHutDeveloper (approved):

ollietreend (commented):
@starswan This looks great.

I've added some comments – mostly questions which probably need clarifying due to the ticket not being clear.

But also: one possible code readability improvement.

PR #1382, opened 2020-10-28
MO-60 move com name to case info

starswan:
During the Community API work, it was discovered that the Allocation was filling in the com_name field from the DeliusData table - which is an 'under the hood' DeliusData import. The com_name field always belonged in case_information (it is an external field) and putting it in Allocation caused some subtle bugs around the updated_at field.
This PR moves the com_name field from Allocation to CaseInformation, and provides a rake task to back-fill the com_name field. There is a separate PR to tidy up and remove the field from Allocation completely, as the rake task needs both fields to exist. The data migration could have been slow without complex SQL, so this has been chosen as the simple option

### DIFF
--- a/app/jobs/process_delius_data_job.rb
+++ b/app/jobs/process_delius_data_job.rb
@@ -62,35 +62,28 @@ private
                                   error_type: error_type(field)
       end
     end
-
-    update_com_name(delius_record)
-  end
-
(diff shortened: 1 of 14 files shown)


ollietreend (approved):
@starswan we could also do with updating

PR #1384, opened 2020-10-28
POM-873 use community API rather than XLSX file for delius offender data

starswan:
We currently use an XLSX file extract to load offender data from ndelius. This PR converts that into a load via the community API, but leaves the extract for team and LDU data. It removes the (temporary table) DeliusData model, and also the related TierChange model that was recorded but never surfaced to the user


app/views/case_information/show.html.erb
```diff
@@ -16,13 +16,6 @@
 
 <h1 class="govuk-heading-xl govuk-!-margin-top-4 govuk-!-margin-bottom-4">Case information</h1>
 
-<p class="govuk-!-margin-bottom-8">
-  <% if @delius_data.any? %>
-    Case information last updated from nDelius: <%= format_date(@delius_data.last.updated_at) %>
-  <% end %>
-  Next scheduled update: <%= format_date(@next_update_date) %>
-</p>
-
```
ollietreend, line 25: I wonder if this is still a useful piece of information for users to see.

We're still updating our view of nDelius as a polling/batch job – rather than being live. So it still stands that there's a "last updated" and "next scheduled update".

Would be good to know what to do here. Is there a user need to keep it?

app/services/offender_service.rb
```diff
@@ -18,6 +18,23 @@ def self.get_offender(offender_no)
     }
   end
 
+  def self.get_community_data(nomis_offender_id)
+    community_info = HmppsApi::CommunityApi::get_offender(nomis_offender_id).deep_symbolize_keys
+    registrations = HmppsApi::CommunityApi::get_offender_registrations(nomis_offender_id).
+        map(&:symbolize_keys).select { |r| r.fetch(:active) && r.fetch(:registerLevel).starts_with?('M') }
```
ollietreend, line 24: I think you've got the format wrong for the `registerLevel` field.

I have a feeling this implementation (and associated `:community_registration` factory) isn't faithful to the Community API. The value of `registerLevel` doesn't match what I'd expect to see from the Community API.

The factory returns a String:

```json
"registerLevel": "M1"
```

But [the Swagger docs](https://community-api-public.test.delius.probation.hmpps.dsd.io/swagger-ui/index.html#/Registrations/getOffenderRegistrationsByNomsNumberUsingGET) say it should be a Hash:

```json
"registerLevel": {
  "code": "string",
  "description": "string"
}
```

I've hit the real Community API and can confirm that it behaves as documented. Here's a response which is returned for a 'MAPPA level 0' offender:

```json
"registerLevel": {
  "code": "M0",
  "description": "MAPPA Nominal (level to be determined)"
}
```

ollietreend: **Also:** in real-world Community API responses, the `registrations` endpoint also returns Hashes which don't include a `registerLevel` key.

It looks like `registerLevel` is optional, so we should treat it as such here and ignore registrations without it.

spec/factories/community_data.rb
```diff
@@ -0,0 +1,28 @@
...
+    active { true }
+    probationArea { { nps: true } }
+    staff { { unallocated: false, surname: 'Jones', forenames: 'Ruth Mary' } }
+  end
+
+  factory :community_registration, class: Hash do
+    initialize_with { attributes }
+
+    active { true }
+    registerLevel { 'M1' }
```
ollietreend, line 26: See previous comment about the format of `registerLevel` field. I think this factory is wrong.

spec/support/helpers/api_helper.rb
```diff
@@ -147,4 +148,14 @@ def stub_keyworker(prison_code, offender_id, keyworker)
...
   end
+
+  def stub_community_offender(nomis_offender_id, community_data)
+    stub_request(:get, "#{COMMUNITY_HOST}/offenders/nomsNumber/#{nomis_offender_id}/all").
+        to_return(body: community_data.to_json)
+  end
+
+  def stub_community_registrations(nomis_offender_id, registrations)
+    stub_request(:get, "#{COMMUNITY_HOST}/offenders/nomsNumber/#{nomis_offender_id}/registrations").
+        to_return(body: registrations.to_json)
```
ollietreend, line 159: The response `body` should be a Hash with a nested `registrations` Array.

### For example, given an offender with 'no registrations':

The response `body` from this stub is:

```json
[]
```

But the actual response from the Community API is:

```json
{
  "registrations": []
}
```

app/views/case_information/show.html.erb
starswan: We've lost the ability to know when case information was last updated - although I guess the case_information record updated_at gives some clues...?

ollietreend: @starswan Yes – we don't have that as it stands. But I'm thinking it should be possible for us to implement something which touches a _"last seen in nDelius"_ field every night, even if the record is already 'in sync'.

But that's only worth doing if we think there's still a design / user need for it.

starswan: I've restored the feature for now using last touched for case information which is *nearly* right

app/jobs/process_delius_data_job.rb
```diff
@@ -5,18 +5,16 @@ class ProcessDeliusDataJob < ApplicationJob
...
       DeliusImportError.where(nomis_offender_id: nomis_offender_id).destroy_all
-      delius_data = DeliusData.where(noms_no: nomis_offender_id)
-      if delius_data.count > 1
-        DeliusImportError.create! nomis_offender_id: nomis_offender_id,
-                                  error_type: DeliusImportError::DUPLICATE_NOMIS_ID
-      elsif delius_data.count == 1
-        import_data(delius_data.first)
-      end
+
+      import_data(OpenStruct.new(OffenderService.get_community_data(nomis_offender_id)))
```
ollietreend, line 17: Should this `OpenStruct` actually be a `CommunityDeliusData`?

Otherwise I can't find a use of the `CommunityDeliusData` struct.

app/services/offender_service.rb
```diff
@@ -18,6 +18,24 @@ def self.get_offender(offender_no)
     }
   end
 
+  def self.get_community_data(nomis_offender_id)
+    community_info = HmppsApi::CommunityApi::get_offender(nomis_offender_id).deep_symbolize_keys
+    registrations = HmppsApi::CommunityApi::get_offender_registrations(nomis_offender_id).deep_symbolize_keys
+                        .fetch(:registrations)
+                        .select { |r| r.fetch(:active) && r.dig(:registerLevel, :code).starts_with?('M') }
```
ollietreend, line 25: `registerLevel` is an optional field, and won't exist for every registration (since not every registration is a MAPPA level).

Will this fail if `r.dig(:registerLevel, :code)` is `nil`?

IRB says we'll get:

```
NoMethodError (undefined method `starts_with?' for nil:NilClass)
```

spec/controllers/allocations_controller_spec.rb
```diff
@@ -78,16 +78,19 @@
       end
 
       context 'when DeliusDataJob has updated the COM name' do
```
ollietreend, line 80: Is this context now: "when the COM name has changed in nDelius" ?

spec/controllers/allocations_controller_spec.rb
```diff
@@ -100,7 +103,7 @@
           end
 
           it 'doesnt mess up the allocation history updated_at because we surface the value' do
```
ollietreend, line 105: This test seems less relevant now that COM name isn't stored in the Allocation record.

spec/factories/community_data.rb
```diff
@@ -0,0 +1,28 @@
...
+    probationArea { { nps: true } }
+    staff { { unallocated: false, surname: 'Jones', forenames: 'Ruth Mary' } }
+  end
+
+  factory :community_registration, class: Hash do
+    initialize_with { attributes }
+
+    active { true }
+    registerLevel { { code: 'M1' } }
+  end
```
ollietreend, line 27: Might be nice to have a non-MAPPA trait on this factory, to produce more realistic data from the registrations endpoint. – i.e. to show that we correctly filter out non-MAPPA registration records.

spec/features/delius_import_job_feature_spec.rb
```diff
@@ -55,22 +57,32 @@
...
+      stub_community_registrations(offender_no, [])
     end
 
     include_examples "imports the Delius spreadsheet and creates case information"
   end
 
-  context "when the team is not associated with an LDU" do
+  context "when the team is not associated with an LDU", :js do
     before do
       ENV['DELIUS_EMAIL_FOLDER'] = 'delius_import_feature'
```
ollietreend, line 71: This env var – `ENV['DELIUS_EMAIL_FOLDER']` – is only used by `DeliusImportJob`. But it looks like this test file has now shifted focus to test `ProcessDeliusDataJob` instead. So might need some clean-up.

It's tricky because this feature spec used to test the entire import – i.e. it can read the XLSX file, update teams/LDUs, and then import offenders.

theKHutDeveloper (approved):

spec/features/delius_import_job_feature_spec.rb
ollietreend: As discussed: it makes sense to keep this feature spec until we've completely moved away from the XLSX import (and therefore have stopped the 'auto-fixing' of teams/LDUs).

ollietreend (approved):

PR #1386, opened 2020-10-30
MO-98 send deallocation email when offender is released or transferred

starswan:
When an offender is released on transferred, the POM is no longer involved but they do not get any notification.

This PR adds an email notification to the primary POM when a release event occurs.

### DIFF
--- a/app/models/allocation.rb
+++ b/app/models/allocation.rb
@@ -89,32 +89,6 @@ class Allocation < ApplicationRecord
     end
   end
 
-  def deallocate_offender(movement_type)
-    self.primary_pom_nomis_id = nil
-    self.primary_pom_name = nil
-    self.primary_pom_allocated_at = nil
-    self.secondary_pom_nomis_id = nil
-    self.secondary_pom_name = nil
-    self.recommended_pom_type = nil
-    self.event = if movement_type == Allocation::OFFENDER_RELEASED
-                   DEALLOCATE_RELEASED_OFFENDER
-                 else
-                   DEALLOCATE_PRIMARY_POM
-                 end
-    self.event_trigger = movement_type
-
-    # This is triggered when an offender is released, and previously we
-    # were setting their prison to nil to show that the current allocation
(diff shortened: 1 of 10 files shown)


spec/models/allocation_spec.rb
```diff
@@ -152,7 +155,7 @@
         AllocationService.create_or_update(params)
 
         alloc = described_class.find_by(nomis_offender_id: nomis_offender_id)
-        alloc.deallocate_offender(movement_type)
+        AllocationService.offender_released(alloc)
```
ollietreend, line 158: Is this still the right place for us to be testing this functionality?

Looks like we're now testing the behaviour of the `AllocationService.offender_released` method. Would it fit better into `spec/services/allocation_service_spec.rb`?

app/models/allocation.rb
```diff
@@ -77,32 +77,6 @@ def override_reasons
     JSON.parse(self[:override_reasons]) if self[:override_reasons].present?
   end
 
-  def deallocate_offender(movement_type)
```
ollietreend, line 80: Whilst the `deallocate_offender` method has moved to the `AllocationService`, the related methods `deallocate_primary_pom` and `deallocate_secondary_pom` have stayed in the `Allocation` model.

It feels like they belong together, since they all do very similar things. It could be a bit confusing to have them split across two places.

ollietreend (commented):
Looks good, but I have added a few comments/questions.

app/mailers/allocation_mailer.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+class AllocationMailer < GovukNotifyRails::Mailer
```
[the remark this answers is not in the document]
starswan: Thought that the PomMailer class was becoming a bucket - maybe the above should be moved to reflect allocations, thus shrinking PomMailer?

ollietreend (approved):

PR #1387, opened 2020-10-30
Make the check for nDelius case info more truthful

ollietreend:
The `Offender` object comes with a method `.has_case_information?` which will return true is a case information record has been loaded for that offender.

This was reliant upon the tier field being set, which was slightly hacky and would fail if tier was legitimately empty.

This commit changes the method to check whether a CaseInformation record has been loaded at all, rather than depending on the value of that specific field.


starswan (approved):
Looks good - did you consider doing delegation for the case_information fields like tier rather than keeping them as copies on this object?

PR #1395, opened 2020-11-03
Pom 785 Handover - Chase email (MO-82)

theKHutDeveloper:
The purpose of this PR is to send a follow-up email to LDU's if a community offender manager (COM) has not been allocated a week **AFTER** the offenders handover start date has passed.

Background:
We need to let the LDUs know about cases that need to have a COM allocated as they are coming up for ‘handover’ of case responsibility from prison offender manager (POM) to COM, or the offender is coming up for a first parole hearing.

This is a reminder email if no action has been taken after the first email has been sent (POM-486)

Note: LDU's without an email address are not included.


app/jobs/handover_follow_up_job.rb
```diff
@@ -0,0 +1,32 @@
+class HandoverFollowUpJob < ApplicationJob
+  queue_as :default
+
+  def perform(date)
+    ldus = LocalDivisionalUnit.all.where.not(email_address: nil)
```
starswan, line 5: There is a scope for this which does something slightly different - LocalDivisionalUnit.without_email_addresses

app/jobs/handover_follow_up_job.rb
```diff
@@ -0,0 +1,32 @@
...
+          offender_name: offender.full_name,
+          offender_crn: offender.crn,
+          sentence_type: offender_type,
+          ldu_email: offender.ldu.email_address,
+          prison: PrisonService.name_for(offender.prison_id),
+          start_date: offender.handover_start_date,
+          responsibility_handover_date: offender.responsibility_handover_date,
+          pom_name: pom.full_name,
+          pom_email: pom.email_address
+        ).deliver_later
```
starswan, line 28: IMHO it's better to use deliver_now, so that this job retries on failure rather than creating another job

app/jobs/handover_follow_up_job.rb
```diff
@@ -0,0 +1,32 @@
+class HandoverFollowUpJob < ApplicationJob
+  queue_as :default
+
+  def perform(date)
+    ldus = LocalDivisionalUnit.all.where.not(email_address: nil)
+
+    ldus.each do |ldu|
+      offenders = ldu.teams.map { |team| team.case_information.where(com_name: nil).map(&:nomis_offender_id) }.flatten
+                  .map { |off_id| OffenderService.get_offender(off_id) }
+                  .select { |o| o.handover_start_date.present? && o.handover_start_date == date - 1.week }
```
starswan, line 10: It might be better to use >= 1.week here to be defensive.

app/jobs/handover_follow_up_job.rb
```diff
@@ -0,0 +1,32 @@
+class HandoverFollowUpJob < ApplicationJob
+  queue_as :default
+
+  def perform(date)
+    ldus = LocalDivisionalUnit.all.where.not(email_address: nil)
+
+    ldus.each do |ldu|
+      offenders = ldu.teams.map { |team| team.case_information.where(com_name: nil).map(&:nomis_offender_id) }.flatten
+                  .map { |off_id| OffenderService.get_offender(off_id) }
```
starswan, line 9: The fallout from POM-486 showed that this isn't enough. We need to make sure that offender is sentenced, and that get_offender doesn't return nil

app/jobs/handover_follow_up_job.rb
```diff
@@ -0,0 +1,32 @@
...
+                  .select { |o| o.handover_start_date.present? && o.handover_start_date == date - 1.week }
+
+      offenders.each do |offender|
+        allocation = Allocation.where(prison: offender.prison_id).first
```
starswan, line 13: This looks like it picks up the wrong allocation when there are multiple allocations in a prison

spec/jobs/handover_follow_up_job_spec.rb
```diff
@@ -0,0 +1,127 @@
...
+
+  let(:indeterminate_without_ted) do
+    build_offender(sentence_type: :indeterminate, release_date: nil, ted: nil)
+  end
+
+  let(:determinate_offender) do
+    build_offender(sentence_type: :determinate, release_date: Time.zone.today + 8.months, ted: nil)
+  end
+
+  it "does not queue jobs for LDU's without email addresses" do
```
starswan, line 23: IMHO this pattern doesn't really prove enough - it shoulld be showing that the email gets sent with the correct parameters. There is an example in responsibilities_controller_spec.rb

app/jobs/handover_follow_up_job.rb
theKHutDeveloper: I originally did that but it was quite inefficient

app/jobs/handover_follow_up_job.rb
theKHutDeveloper: Ok

app/jobs/handover_follow_up_job.rb
theKHutDeveloper: We only send this once when a week has past, your suggestion means it would potentially happen every day

app/jobs/handover_follow_up_job.rb
theKHutDeveloper: Rookie mistake :/

spec/jobs/handover_follow_up_job_spec.rb
```diff
@@ -0,0 +1,186 @@
...
+
+    before do
+      allow(OffenderService).to receive(:get_offender).and_return(determinate_offender)
+    end
+
+    it 'does not send emails for offenders whose are not yet approaching handover' do
+      create(:allocation, nomis_offender_id: determinate_offender.offender_no)
+      determinate_offender.load_case_information(case_info)
+
+      expect_any_instance_of(CommunityMailer).not_to receive(:urgent_pipeline_to_community).and_call_original
```
starswan, line 126: the 'and_call_original' doesn't make much sense here as we're expecting not to be called

spec/jobs/handover_follow_up_job_spec.rb
```diff
@@ -0,0 +1,186 @@
+require 'rails_helper'
+
+RSpec.describe HandoverFollowUpJob, type: :job do
+  include ActiveJob::TestHelper
```
starswan, line 4: think this is now not needed as there as no queueing tests

lib/tasks/handover_chase_email.rake
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+require 'rake'
+
+namespace :handover_chase_emails do
+  desc 'Send follow-up emails LDUs two weeks before handover if COM still not assigned'
+  task process: :environment do
+    HandoverFollowUpJob.perform_later(Time.zone.today + 14.days)
```
starswan, line 8: I think there is a problem here (or a problem with my understanding). We are supposed to nudge when the handover date is 1 week past, however this date has 1 week substracted in the code (so it is today + 1.week in production) so I think this alerts 1 week *before* the handover is due rather than 1 week after. I think this should just be Time.zone.today?

spec/jobs/handover_follow_up_job_spec.rb
```diff
@@ -0,0 +1,186 @@
...
+
+  let(:pom) do
+    OpenStruct.new  staff_id: 99999,
```
starswan, line 7: There is a factory called :pom to make things like this.

lib/tasks/handover_chase_email.rake
theKHutDeveloper: That should have been changed, thanks for spotting that

spec/jobs/handover_follow_up_job_spec.rb
theKHutDeveloper: Can you suggest what I should be doing then?

starswan (approved):

PR #1400, opened 2020-11-04
MO-101 Sort on POM and COM column in the allocator view

Katy600:
This PR allows the SPO to sort the list of handovers by COM or POM name

### **Before:**

**SPO's are not able to sort by POM or COM**

Screenshot 2020-03-30 at 08 37 31

### **After** 

**SPOs can sort POM alphabetically**

Screenshot 2020-11-04 at 15 21 39

_SPOs can sort COMs alphabetically_
Screenshot 2020-11-04 at 15 21 23

### DIFF
--- a/app/services/summary_service.rb
+++ b/app/services/summary_service.rb
@@ -10,7 +10,7 @@ class SummaryService
     sortable_fields = {
         allocated: [:last_name, :earliest_release_date, :tier],
         new_arrivals: [:last_name, :prison_arrival_date, :earliest_release_date],
-        handovers: [:last_name, :handover_start_date, :responsibility_handover_date, :case_allocation],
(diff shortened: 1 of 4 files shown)


ollietreend (approved):
Looks good @Katy600 – nice work! 🎉

### One small thing to be aware of:

I've just made a change to the file `spec/features/spo_handover_cases_feature_spec.rb` which is now merged into the `main` branch. It removes VCR from the test in favour of stubbing the offender. You can see that change in this commit: 508d319

It looks like this is the same approach you've already taken in this PR – so the two should be compatible. I just wanted to point it out in case there are merge conflicts in the file when you come to rebase against `main`.

I'm happy to help if there are merge conflicts and you're not sure how to resolve them!

PR #1402, opened 2020-11-06
MO-55 move handover code into seperate controllers to help navigation

starswan:
In order to improve the navigation, the handovers code needed to be re-organised into it's own section so that it can be referenced from the new navigation. This PR makes this move, and hence unblocks the navigation story


app/controllers/prison_staff_application_controller.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+class PrisonStaffApplicationController < PrisonsApplicationController
```
ollietreend, line 3: If I understand the intention of this class correctly, I wonder if a name like `PomApplicationController` would be more descriptive?

From what I can tell: this is an 'application controller' for actions related to the currently signed in POM.

The name "prison staff" is a bit misleading because SPOs/HOMDs are also members of prison staff.

ollietreend: Ah – ignore me. Looks like the "prison staff" name comes from the routing structure `/prisons/:prison_id/staff/:staff_id`, so I guess it makes sense within that context.

Still – it seems like we use "POM" and "staff" interchangeably, which is a little confusing. But not something to be resolved in this PR.

ollietreend (approved):

PR #1408, opened 2020-11-13
Refactor the signin_spo_user method in the helper 

Katy600:
The method _signin_spo_user_ is used for test set up when creating tests involving an SPO user. It now includes stub user. This will hopefully make it easier when writing tests as you no longer need to stub the user separately. It also hard codes the users name and staff id, as there doesn't seem to be a case where it needs to be specific. It makes prisons the only parameter that can be assigned as you might want your spo to be signed into a particular prison. 

**Example:** 
_signin_spo_user('LEI')
prison_summary_unallocated_path('LEI')_

It also fixes all the test that failed as a result of the change.

### DIFF
--- a/spec/support/helpers/features_helper.rb
+++ b/spec/support/helpers/features_helper.rb
@@ -1,8 +1,21 @@
 # frozen_string_literal: true
 
 module FeaturesHelper
-  def signin_spo_user(name = 'MOIC_POM', caseloads = %w[LEI RSI])
-    mock_sso_response(name, [SsoIdentity::SPO_ROLE], caseloads)
+  def signin_spo_user(prisons = ['LEI'])
+    mock_sso_response('MOIC_POM', [SsoIdentity::SPO_ROLE], prisons)
+  end
+
+  def stub_signin_spo(pom, prisons = ['LEI'])
+    stub_auth_token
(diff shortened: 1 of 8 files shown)


spec/support/helpers/features_helper.rb
```diff
@@ -1,8 +1,21 @@
...
 
 module FeaturesHelper
-  def signin_spo_user(name = 'MOIC_POM')
-    mock_sso_response(name, [SsoIdentity::SPO_ROLE])
+  def signin_spo_user(prison = ['LEI'])
+    mock_sso_response('MOIC_POM', [SsoIdentity::SPO_ROLE], prison)
+  end
+
+  def stub_signin_spo(pom, _prison = ['LEI'])
+    stub_auth_token
```
starswan, line 9: this looks really weird? Waht are you trying to achieve?

spec/support/helpers/features_helper.rb
```diff
@@ -1,8 +1,21 @@
...
+    signin_spo_user(_prison = ['LEI'])
```
starswan, line 10: I think the issue is that this isn't really valid ruby syntax. It's assigning a temporary unused variable

starswan (approved):

PR #1410, opened 2020-11-13
Mo 56 alloc history partial path

starswan:
More ongoing refactoring of allocation history reflecting the needs of MO-56. This piece changes the massive 'if' statement in the allocation history.html.erb to set of partials determined by the event name.


theKHutDeveloper (approved):

starswan:
This needs to have MO-36 as well before it can be merged into mainline and deployed @Katy600

PR #1412, opened 2020-11-17
Add handover chase email to cron job to run daily

theKHutDeveloper:


deploy/production/cronjob.yaml
```diff
@@ -104,4 +104,57 @@ spec:
...
+                      key: postgres_name
+                - name: POSTGRES_USER
+                  valueFrom:
+                    secretKeyRef:
+                      name: allocation-rds-instance-output
+                      key: postgres_user
+                - name: REDIS_URL
+                  valueFrom:
+                    secretKeyRef:
+                      name: elasticache-offender-management-allocation-manager-token-cache-preprod
```
starswan, line 157: typo - needs to be elasticache-offender-management-allocation-manager-token-cache-production

starswan: wondering if we shouldn't use a seperate file for each of these cron jobs - that'as the approach I have taken

starswan (approved):

PR #1420, opened 2020-11-20
Mo 87 - Notify SPOs and POMs when no LDU information is available shortly before handover start date 

theKHutDeveloper:
This means that we have no way of notifying the LDU that they will be responsible for the offender. We alert the SPO and POM in these cases so that they can take necessary action outside of the application to deal with this circumstance.

There are 2 SPO views that will display this alert, allocations/new and allocations/show pages. Both views are quite similar so I am only displaying the allocations/new page in this PR.

### - **Allocate a Prison Offender Manager - top of the page**

Screen Shot 2020-11-20 at 15 52 44

### - **Allocate a Prison Offender Manager - lower part of the page**

Screen Shot 2020-11-20 at 15 53 01

**-------------------------------------------------------------------------------------------**

There is one POM view that will display this alert on the prisoners/show page

### - **Prisoner Information - top of the page**
Screen Shot 2020-11-20 at 15 54 49

### - **Prisoner Information - bottom of the page**
Screen Shot 2020-11-20 at 15 55 00

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -50,3 +50,13 @@ a:visited {
  .govuk-header__navigation-item a {
      color: white !important;
  }
+
+.govuk-table__cell-error {
+  border-left: 5px solid #d4351c;
+  padding-left: 10px;
+}
+
+.govuk-table__cell-error-value {
+  color: #d4351c;
+  font-weight: bold;
+}
--- a/app/helpers/offender_helper.rb
+++ b/app/helpers/offender_helper.rb
@@ -31,7 +31,7 @@ module OffenderHelper
     end
   end
 
-  def approaching_handover_without_com?(offender)
+  def needs_com_but_ldu_is_uncontactable?(offender)
     return false unless offender.sentenced?
 
     return false if offender.handover_start_date.nil?
--- a/app/helpers/offender_helper.rb
+++ b/app/helpers/offender_helper.rb
@@ -30,4 +30,16 @@ module OffenderHelper
       type + 'allocated'
     end
   end
+
+  def approaching_handover_without_com?(offender)
+    return false unless offender.sentenced?
+
+    return false if offender.handover_start_date.nil?
+
+    if offender.handover_start_date <= Time.zone.today + 45.days && offender.allocated_com_name.nil?
+      true
+    else
+      false
+    end
+  end
 end
--- a/app/helpers/offender_helper.rb
+++ b/app/helpers/offender_helper.rb
@@ -36,10 +36,12 @@ module OffenderHelper
 
     return false if offender.handover_start_date.nil?
 
-    if offender.handover_start_date <= Time.zone.today + 45.days && offender.allocated_com_name.nil?
-      true
-    else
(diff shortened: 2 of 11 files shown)


app/helpers/offender_helper.rb
```diff
@@ -30,4 +30,18 @@ def event_type(event)
       type + 'allocated'
     end
   end
+
+  def approaching_handover_without_com?(offender)
+    return false unless offender.sentenced?
```
starswan, line 35: I think this method might be nicer as a 'traditional if statement'. If I'm reading it correctly, it can return false even when com_name is not present. (if there's no email address to talk to) which seems a little counter-intuitive given the name of the method. 
I also wonder if it would be better if it returned the positive case rather than the negative one? (unsure)

app/views/shared/_no_com_alert.html.erb
```diff
@@ -0,0 +1,17 @@
+<div class="govuk-error-summary" aria-labelledby="error-summary-title" role="alert" tabindex="-1" data-module="govuk-error-summary">
+  <h2 class="govuk-error-summary__title" id="error-summary-title">
+    You must contact the community probation office
+  </h2>
+  <div class="govuk-error-summary__body">
+    <ul class="govuk-list govuk-error-summary__list">
+      <li id="com_warning"><%= link_to I18n.t('views.handover_alert.no_com', date: format_date(@prisoner.handover_start_date)), "#com-error", data: { turbolinks: false } %></li>
+      <li id="ldu_warning">
+        <% if @prisoner.ldu.try(:name).blank? %>
```
starswan, line 9: This looks like this shoudl be encapsulated as @prisoner.has_ldu?

app/helpers/offender_helper.rb
starswan: Maybe this should be split into 2 for readability. approaching_handover_without_com? sounds like a question, as does offender.has_ldu_email? Then the checks can be more explicit i.e. if approaching_handover_without_com? && offender.ldu_email_missing?

app/assets/stylesheets/application.scss
```diff
@@ -50,3 +50,13 @@ a:visited {
  .govuk-header__navigation-item a {
      color: white !important;
  }
+
```
starswan, line 53: Slightly surprised that this isn't available already - maybe a comment would be appropriate?

app/helpers/offender_helper.rb
```diff
@@ -30,4 +30,18 @@ def event_type(event)
       type + 'allocated'
     end
   end
+
+  def approaching_handover_without_com?(offender)
```
ollietreend, line 34: I'm wondering if this might be a more descriptive name for this method:

```suggestion
  def needs_com_but_ldu_is_uncontactable?(offender)
```

It's a bit longer, but hopefully more descriptive. Not sure if it's too wordy though?

No need to use this, though. Naming things is hard 😓

ollietreend (approved):
This looks good 🎉

I've left one comment about the helper method name – but this is just an opinion about naming, so no need to use it!

PR #1422, opened 2020-11-23
Mo46 add history link to new allocation page

starswan:
Add a link to the 'allocation history' page from the 'new allocation' flow. 
This PR also tightens up the allocations table, as the old code was calling 
Allocation.where(nomis_offender_id: <offender_no>).last which implies that there can be more than 1 allocation per offender. This is not true, so this code has been replaced with a find_by, and the allocations table given a unique key to enforce the constraint. There are no duplicates in production, so this should be extremely safe.

### DIFF
--- a/app/controllers/allocations_controller.rb
+++ b/app/controllers/allocations_controller.rb
@@ -12,6 +12,7 @@ class AllocationsController < PrisonsApplicationController
     @recommended_poms, @not_recommended_poms =
       recommended_and_nonrecommended_poms_for(@prisoner)
     @unavailable_pom_count = unavailable_pom_count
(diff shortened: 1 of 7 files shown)


ollietreend (approved):
@starswan It'd be nice to also update the `Allocation` model validation on the `nomis_offender_id` field so that Rails knows it should be unique, too. Otherwise saves could fail and Rails wouldn't know why.

There's currently a validation defined at line 117, but it only checks for presence:

https://github.com/ministryofjustice/offender-management-allocation-manager/blob/5401b166980b224c2a6e47490bde7b38c3abe376/app/models/allocation.rb#L117-L122

Aside from that – approved ✅

PR #1423, opened 2020-11-25
MO-228 - replace (erroneous) custom partial 

starswan:
We had a custom-built partial to render errors on govuk formbuilder forms. However this partial was (a) incorrect and (b) unnecessary as there is a one-liner which can be added to forms to do all the heavy lifting. 

This PR improves the early allocation feature test to be more robust around that failure, and also fixes it.


spec/features/early_allocation_feature_spec.rb
```diff
@@ -119,7 +128,16 @@
           within '.govuk-error-summary' do
             expect(page).to have_text 'You must say if this is a MAPPA level 2 case'
 
-            expect(all('li').count).to eq(5)
+            expect(all('li').map(&:text)).
+                to match_array([
+                          "You must say if this prisoner has been in an extremism separation centre",
+                          "You must say if there is another reason for early allocation",
+                          "You must say whether this prisoner presents a risk of serious harm",
+                          #"#early-allocation-due-for-release-in-less-than-24months-field-error",
```
theKHutDeveloper, line 136: Does this commented line of code need to be here?

theKHutDeveloper (approved):

PR #1428, opened 2020-11-30
MO-256 refactor responsibility service

starswan:
The Handover Date Service and Responsibility Service have grown apart for 2 years - however they are opposite sides of the same coin. When the handover date for an offender reaches today, that is the same fact as the responsibility switching from 'Responsible' (custody) to 'Suppprting' (community).

This PR brings those 2 pieces of code together (they were calling each other anyway) so that sensible decisions can be made on their calculations confidently (e.g. knowing the difference between an 'N/A' date (supprting responsibility), an Unknown date (e.g. indefinite custody responsibility) and a potentially 'broken' case (unknown responsibility)


app/controllers/coworking_controller.rb
```diff
@@ -51,7 +51,7 @@ def create
   def confirm_removal
     @prisoner = offender(coworking_nomis_offender_id_from_url)
 
-    @responsibility_type = ResponsibilityService.calculate_pom_responsibility(@prisoner)
+    @responsibility_type = HandoverDateService.calculate_pom_responsibility(@prisoner)
```
ollietreend, line 54: Unrelated to this refactor, but a potential bug 🐛:

I'm pretty sure we should be using `@prisoner.pom_responsibility` rather than calculating it manually with `HandoverDateService.calculate_pom_responsibility` here. Otherwise overrides will not be taken into account.

app/services/email_service.rb
```diff
@@ -74,7 +74,7 @@ def url
   end
 
   def current_responsibility
-    ResponsibilityService.
+    HandoverDateService.
       calculate_pom_responsibility(@offender).to_s.downcase
```
ollietreend, line 78: Same thing here. I think this should be:

```suggestion
    @offender.pom_responsibility.to_s.downcase
```

Otherwise overrides aren't taken into account.

app/services/handover_date_service.rb
```diff
@@ -1,24 +1,58 @@
...
-      HandoverData.new nil, nil, 'Recall case - no handover date calculation'
-    elsif offender.nps_case? && offender.indeterminate_sentence? && offender.tariff_date.nil?
-      HandoverData.new nil, nil, 'No earliest release date'
-    elsif ResponsibilityService.recent_prescoed_nps_case?(offender) && offender.indeterminate_sentence?
-      HandoverData.new offender.prison_arrival_date, prescoed_handover_date(offender), 'Prescoed'
+      HandoverData.new SUPPORTING, nil, nil, 'Recall case'
+    elsif offender.immigration_case?
+      HandoverData.new SUPPORTING, nil, nil, 'Immigration Case'
+    elsif offender.nps_case? && offender.indeterminate_sentence? && (offender.tariff_date.nil? || offender.tariff_date < Time.zone.today)
+      HandoverData.new RESPONSIBLE, nil, nil, 'No earliest release date'
```
ollietreend, line 34: @starswan As discussed on Slack earlier today – there is a discrepancy between the Handover Date Service and the Responsibility Service when calculating on **an indeterminate offender with a Tariff Date in the past**.

**Handover Date Service** would still give handover dates, even though they're in the past:

https://github.com/ministryofjustice/offender-management-allocation-manager/blob/0111140c8e55c79818a79a0b6e6d81b30f143ee2/app/services/handover_date_service.rb#L16-L18

And the **Responsibility Service** would say the POM is responsible:

https://github.com/ministryofjustice/offender-management-allocation-manager/blob/0111140c8e55c79818a79a0b6e6d81b30f143ee2/app/services/responsibility_service.rb#L30-L33

The change made in this PR would change the behaviour:
- Handover dates become `nil` (where they were previously dates in the past)
- Responsibility remains unchanged as `'Responsible'`

I'm making this comment to flag that _I think_ this refactor changes the behaviour for these offenders. Is that the expected behaviour?

ollietreend (commented):
@starswan this refactor is much needed – it should make things a lot easier to reason with. I've added a couple of comments/questions – happy to discuss.

app/services/handover_date_service.rb
starswan: One for Laura I think. If the tariff_date is in the past but the offender is still in prison, IMHO it's the same as not having a tariff date - they have failed their parole hearing, have a PED or a PRD (obviously we can't get PRD). I don't think I changed any tests substantially which would indicate that this behaviour has changed? [Not that that proves anything, just that we might have an uncovered case - bit annoying if so]

spec/services/handover_date_service_spec.rb
```diff
@@ -371,10 +376,10 @@
           let(:indeterminate_sentence) { true }
 
           context 'with tariff date in the future' do
```
starswan, line 378: This is a bit scary - this test *should* have failed on 2/11/20

app/controllers/coworking_controller.rb
starswan: This code is a bit strange. It is trying to display the fact that the Primary POM is in a Responsible/Supporting role as we remove the co-working POM. I guess you're right.

app/services/email_service.rb
starswan: same as above - it's mostly harmless as it's display code, but it's probably wrong

app/controllers/coworking_controller.rb
ollietreend: Fixed by #1435

app/services/email_service.rb
ollietreend: Fixed by #1435

app/services/handover_date_service.rb
ollietreend: Looks like it used to be covered by a test:

https://github.com/ministryofjustice/offender-management-allocation-manager/blob/bac5eba830917919ca61a81ec7db0ccf22abad32/spec/services/responsibility_service_spec.rb#L19-L32

But this has since been refactored / removed from the codebase, and I don't think we explicitly cover this scenario any more.

ollietreend (approved):

PR #1432, opened 2020-11-30
Mo 269 early allocation fields

starswan:
Early allocation needs to know whether an allocation assessment was done < 18 or > 18 months - so store it in the model so it can be queried later (it's quite enough to do 'sent', as 'ineligible' assessments are not sent but can be > 18 months)

This PR added those fields, and also adds a rudimentary EmailHistory model to track emails sent to the LDU as a result

### DIFF
--- a/app/models/email_history.rb
+++ b/app/models/email_history.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+class EmailHistory < ApplicationRecord
+  AUTO_EARLY_ALLOCATION = 'auto_early_allocation'
+  DISCRETIONARY_EARLY_ALLOCATION = 'discretionary_early_allocation'
+
+  validates_presence_of :name, :nomis_offender_id
+
+  validates :event, inclusion: { in: [AUTO_EARLY_ALLOCATION, DISCRETIONARY_EARLY_ALLOCATION], allow_nil: false }
+  validates :prison, inclusion: { in: PrisonService.prison_codes, allow_nil: false }
+  validates :email, presence: true, 'valid_email_2/email': true
+end
--- a/spec/controllers/early_allocations_controller_spec.rb
+++ b/spec/controllers/early_allocations_controller_spec.rb
@@ -19,7 +19,8 @@ RSpec.describe EarlyAllocationsController, :allocation, type: :controller do
     ]
   }
 
-  let(:offender) { build(:nomis_offender) }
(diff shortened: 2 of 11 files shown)


app/models/email_history.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class EmailHistory < ApplicationRecord
+  EARLY_ALLOCATION_CREATE = 'create_early_allocation'
```
ollietreend, line 4: I think we're going to need two different events: one for each possible EA outcome which triggers an email.

The prototype [case history page](https://hmpps-moic-staging.herokuapp.com/ea-case-activity) differentiates between 'eligible' and 'discretionary' emails sent to the LDU:

Screenshot 2020-11-30 at 16 34 33

Note the difference in title between:

- **Early allocation decision requested** (for a discretionary outcome)
- **Early allocation assessment form sent** (for an eligible outcome)

Or were you thinking that the type of email sent would be inferred from the outcome on the Early Allocation record?

---

### Naming

In terms of naming, I think it might be better to use `_REFERRAL` rather than `_CREATE` in the name – after all, the reason for this email is _referring it to the LDU_, not just _creating_ the early allocation record. But that's just my opinion – naming is hard.

spec/controllers/early_allocations_controller_spec.rb
```diff
@@ -93,14 +93,17 @@
...
+            post :create, params: { prison_id: prison,
+                                    prisoner_id: nomis_offender_id,
+                                    early_allocation: date_params.merge(s1_boolean_params) }
+          }.to change(EmailHistory, :count).by(1)
           assert_template('eligible')
 
           expect(early_allocation.prison).to eq(prison)
           expect(early_allocation.created_by_firstname).to eq(first_pom.firstName)
           expect(early_allocation.created_by_lastname).to eq(first_pom.lastName)
+          expect(early_allocation.created_within_referral_window).to eq(true)
```
ollietreend, line 106: Is there any value in having a context to assert that `created_within_referral_window` is `false` and an email is not sent when `> 18 months`?

I know we have a split between `<= 18 months` and `> 18 months` in the feature spec – would the same thing would make sense here?

Just a question – not a request.

ollietreend (commented):
@starswan a couple of comments/questions, but otherwise this looks good

ollietreend (approved):

PR #1444, opened 2020-12-01
MO-95 remove 'change' links for nDelius case info records

starswan:
When case information data has been sourced from nDelius, it makes little sense to allow the user to edit it. It can be overwritten at any time (either from the nDelius API, or from DeliusData from the spreadsheet import) so this is a confusing mis-feature

### DIFF
--- a/app/jobs/process_delius_data_job.rb
+++ b/app/jobs/process_delius_data_job.rb
@@ -68,13 +68,14 @@ private
     find_case_info(delius_record).tap do |case_info|
       team = map_team(delius_record.team_code)
       case_info.assign_attributes(
-        com_name: map_com_name(delius_record.offender_manager),
-        crn: delius_record.crn,
-        tier: map_tier(delius_record.tier),
-      team: team,
-      case_allocation: delius_record.service_provider,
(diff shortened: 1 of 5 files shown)


app/jobs/process_delius_data_job.rb
```diff
@@ -68,13 +68,14 @@ def map_delius_to_case_info(delius_record)
...
-      welsh_offender: map_welsh_offender(delius_record.ldu_code),
-      mappa_level: map_mappa_level(delius_record.mappa, delius_record.mappa_levels)
+        manual_entry: false,
+      com_name: map_com_name(delius_record.offender_manager),
+      crn: delius_record.crn,
+      tier: map_tier(delius_record.tier),
+    team: team,
+    case_allocation: delius_record.service_provider,
+    welsh_offender: map_welsh_offender(delius_record.ldu_code),
+    mappa_level: map_mappa_level(delius_record.mappa, delius_record.mappa_levels)
```
ollietreend, line 78: Not sure what's happened here – looks like line indentation has shifted around, but I don't think it was wrong in the first place? Was this accidental?

ollietreend (approved):

ollietreend (approved):

PR #1448, opened 2020-12-02
Standardise cron job configuration

ollietreend:
This PR standardises the configuration for Kubernetes cron jobs.

### Summary of changes

- Improve discoverability of cron jobs by giving each cron job its own config file, and introducing a filename prefix convention to make it easy to find them in the Kubernetes config file.
- Configure Kubernetes to clean up old finished cron jobs (closes MO-227)
- Add missing `REDIS_URL` environment variable to a staging cron job

For more details, read the commit messages.


deploy/production/cron-process-movements.yaml
```diff
@@ -0,0 +1,54 @@
+apiVersion: batch/v1beta1
+kind: CronJob
+metadata:
+  name: offender-manager-process-movements
+spec:
+  schedule: "0 2 * * *"
+  startingDeadlineSeconds: 180
```
ollietreend, line 7: This is an outlier in our config file. No other jobs use it. It's only specified for the 'process movements' job in production.

It looks like it was introduced for good reason in commit b30969f083109243eab82729cad165d0d3657140. From the commit message:
> we are willing to wait past the scheduled start time if the cron job isn't scheduled at the exact time.

Is this important? Should we remove it? Or should it be set on all cron jobs?

@starswan It would be good to get your opinion on this. You may have more context than me.

starswan: Sorry no clue. You would have hoped that a sensible default would have applied here. Can't imaging why k8s would refuse to run a cronjob just because it started a bit late.?

starswan (approved):

deploy/production/cron-process-movements.yaml
ollietreend: I've [asked the Cloud Platform team](https://mojdt.slack.com/archives/C57UPMZLY/p1606925074285800) about it, and the consensus seems to be that `startingDeadlineSeconds` isn't needed.

The [Kubernetes CronJob documentation](https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/) is very unclear, but I think it's useful for jobs which are scheduled to run _very frequently_ – e.g. once per minute – to avoid them becoming permanently disabled after 100 failed runs. Since we run this once per night, 100 failures would take months to reach, so is not relevant to us.

I don't think is anything to do with 'missed schedules' as commit b30969f implied.

I'll remove this line from the config to make it consistent with the other jobs.

PR #1454, opened 2020-12-07
M0-20 early allocation case history

starswan:
This PR adds all the early allocation scenarios to the 'case history' section fot an offender


ollietreend (approved):
Nice work on the presenters 👍

One thing which irks me a little is our built-in assumption that the 'updated at' timestamp will always truthfully reflect the timestamp of a 'community decision recorded' event. In theory that's true, but what if we ever need to touch those records for some other reason? But that's beyond the scope for this PR...

PR #1457, opened 2020-12-07
MO-20 pre-refactor

starswan:
Refactoring to support MO-20 work without too much noise cluttering the actual story

### DIFF
--- a/app/controllers/allocations_controller.rb
+++ b/app/controllers/allocations_controller.rb
@@ -96,12 +96,12 @@ class AllocationsController < PrisonsApplicationController
   def history
     @prisoner = offender(nomis_offender_id_from_url)
 
-    history = Allocation.find_by(nomis_offender_id: nomis_offender_id_from_url).history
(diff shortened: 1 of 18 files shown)


app/models/early_allocation.rb
```diff
@@ -95,7 +95,7 @@ def validate_stage2
             inclusion: { in: [true, false], allow_nil: false },
             if: -> { recording_community_decision }
 
-  def eligible?
+  def auto_eligible?
```
ollietreend, line 98: What's the correct name for this? I'm not convinced it should be renamed.

I'm working on MO-11 which exclusively uses the term "eligible" – even in the UI to users. It **does not** call them "auto eligible".

starswan: This means auto_eligible - hence the rename. I found it confusing when working on the allocation history story (MO-20) as a case can be eligible but say eligible? false...!

ollietreend (approved):

PR #1458, opened 2020-12-07
MO-220: Part 1 - Add outcome field to Early Allocations table. 

theKHutDeveloper:
Add before validation to make sure that outcome is stored when form is saved.

### DIFF
--- a/app/models/early_allocation.rb
+++ b/app/models/early_allocation.rb
@@ -1,12 +1,16 @@
 # frozen_string_literal: true
 
 class EarlyAllocation < ApplicationRecord
+  before_validation :record_outcome
+
   belongs_to :case_information,
              primary_key: :nomis_offender_id,
              foreign_key: :nomis_offender_id,
              inverse_of: :early_allocations
 
   validates_presence_of :prison, :created_by_firstname, :created_by_lastname
+
+  validates_inclusion_of :outcome, in: %w(eligible ineligible discretionary)
 
   validates :oasys_risk_assessment_date,
             presence: true,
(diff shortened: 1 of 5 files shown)


app/models/early_allocation.rb
```diff
@@ -1,13 +1,17 @@
 # frozen_string_literal: true
 
 class EarlyAllocation < ApplicationRecord
+  before_validation :record_outcome
```
starswan, line 4: think this should be before_save ?

app/models/early_allocation.rb
```diff
@@ -117,4 +121,11 @@ def stage1_eligible?
   def all_stage2_false?
     STAGE2_PLAIN_BOOLEAN_FIELDS.map(&method(:public_send)).none?
   end
+
+  def record_outcome
+    return self.outcome = 'eligible' if eligible?
```
starswan, line 126: Think this would be nicer as a simple 'if-else' statement rather than early returns. Also I think a comment would be handy to explain the reason for this field. As we are setting it programattically, the validation should probably be strict? if we overwrite before_save with guaranteed one of 3 values, maybe the validation can just go away? (especially if the last 'if' is just an else in record_outcome)

spec/factories/early_allocations.rb
```diff
@@ -14,6 +14,7 @@
     mappa_level_3 { false }
     cppc_case { true }
     stage2_validation { false }
+    outcome { 'eligible' }
```
starswan, line 17: outcome should be private to the model - not set here.

app/models/early_allocation.rb
theKHutDeveloper: @starswan If I change it to before_save I can't have the validation as it will fail automatically due to the validation. Other option is to remove the validation, but then you will have an outcome set to nil. Unless you have another idea?

starswan: outcome should never be nil on save - it should be 1 of the 3 valid values. Its a little confusing, but the EA can only be in 1 or 3 (5?) states - auto_eligible, discretionary or ineligible. (extra 2 for discretionary true/false).

starswan (approved):

PR #1462, opened 2020-12-08
MO-222 add EA notification to prisoner info screen

starswan:
Add a notification to the prisoner info screen when there is an early allocation that needs to be followed up as we are now within the handover window

### DIFF
--- a/app/views/prisoners/show.html.erb
+++ b/app/views/prisoners/show.html.erb
@@ -8,6 +8,14 @@
   %>
 <% end %>
 
+<% if @prisoner.needs_early_allocation_notify? %>
+  <% content_for :info_banner do %>
+    <%= @prisoner.full_name %> might be eligible for early allocation to the community probation team.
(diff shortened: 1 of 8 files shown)


app/views/prisoners/show.html.erb
```diff
@@ -8,6 +8,14 @@
   %>
 <% end %>
 
+<% if @prisoner.needs_early_allocation_notify? %>
```
ollietreend, line 11: @starswan I think this criteria is subtly wrong.

As I understand it, the requirement is to show the banner if:
- the offender is in the referral window
- there has been AT LEAST ONE early allocation assessment completed which was eligible or discretionary

i.e. it should take all past Early Allocations into account, **not just the most recent one**

---

This same requirement exists in my ticket MO-11 (which is why I spotted it) as well as MO-220.

It seems like something that would be useful to have in the offender object, given that it uses the above criteria. Would that be called something like `is_potentially_eligible_for_early_allocation?` (a bad name, but it's hard to succinctly explain what this criteria represents)

ollietreend (approved):

PR #1464, opened 2020-12-10
MO-107 - refactor early allocation journey to use form objects

starswan:
The Early Allocation code is full of flags, trying to decide when to validate which fields on the form. This PR breaks that functionality out into some 'form object' proxies that mixin ActiveModel::Model to make the code easier to read, reason about and change


app/controllers/early_allocations_controller.rb
```diff
@@ -5,44 +5,48 @@ class EarlyAllocationsController < PrisonsApplicationController
...
-  def edit
-    @early_assignment = EarlyAllocation.where(offender_id_from_url).last
-  end
-
-  def update
-    @early_assignment = EarlyAllocation.where(offender_id_from_url).last
-
-    if @early_assignment.update(community_decision_params)
-      redirect_to prison_prisoner_path(@prison.code, @early_assignment.nomis_offender_id)
+  def create
```
ollietreend, line 37: It would be nice if this method was called `stage2` – in the same way that there's a `stage1` method.

The same is true for the discretionary route – would be nice to have a `stage3` method and form object. Right now we have the `discretionary` method, but no 'stage 3' form object.

This approach would give us a really clear & obvious relationship between the view => form object => controller method for each stage of the form, and sets a strong convention which can be followed elsewhere in the future.

It would mean losing the `create` method altogether – which I realise breaks the traditional REST resource methods. But since this is a multi-stage form/wizard, is there really such thing as one `create` endpoint? The creation could happen in any one of the stages (e.g. in stage 1 as 'eligible'), so it could be argued that the convention of having one single `create` method isn't helpful here.

app/controllers/early_allocations_controller.rb
```diff
@@ -5,44 +5,48 @@ class EarlyAllocationsController < PrisonsApplicationController
 
   def new
     case_info = CaseInformation.find_by offender_id_from_url
-    @early_assignment = case_info.early_allocations.new
+    @early_assignment = EarlyAllocationStage1.new offender_id_from_url
```
ollietreend, line 8: I'm not sure why we've used the name `@early_assignment` throughout the EA controller/views, but it's a bit confusing because nowhere else do we call it an 'assignment'. Perhaps this is an old name which is no longer used?

It would be nice to clean this up and use the name `@early_allocation` – or even `@assessment` as it seems to be colloquially known.

ollietreend (commented):
This is a really nice improvement 👍

I've added a few comments of things to think about. Very happy to talk any of it through. Mostly things around consistency / convention.

ollietreend (approved):

PR #1469, opened 2020-12-10
Mo 220 Early Allocation Suitability email cron job

theKHutDeveloper:
This PR identifies the POM's that have offenders with: 

- Early Allocation assessment forms with the outcome of 'discretionary' or 'eligible'

- that were previously completed before the offenders had 18 months left to serve of their sentence 

- and now have exactly 18 months left until release. 

At this point we email the POM to remind them of their offenders that meets this criteria and ask that they review the Early Allocation assessment form/s.

### DIFF
--- a/app/jobs/suitable_for_early_allocation_email_job.rb
+++ b/app/jobs/suitable_for_early_allocation_email_job.rb
@@ -6,7 +6,7 @@ class SuitableForEarlyAllocationEmailJob < ApplicationJob
   EQUIP_URL = 'https://equip-portal.rocstac.com/CtrlWebIsapi.dll/?__id=webDiagram.show&map=0%3A9A63E167DE4B400EA07F81A9271E1944&dgm=4F984B45CBC447B1A304B2FFECABB777'
 
   def perform
-    offenders = EarlyAllocation.where(created_within_referral_window: false).where.not(outcome: 'ineligible').pluck(:nomis_offender_id).uniq
+    offenders = EarlyAllocation.suitable_offenders_pre_referral_window
 
     offenders.each do |offender_no|
       allocation = Allocation.where(nomis_offender_id: offender_no).first
@@ -18,10 +18,16 @@ class SuitableForEarlyAllocationEmailJob < ApplicationJob
         EarlyAllocationMailer.review_early_allocation(
           email: pom.email_address,
           prisoner_name: prisoner.full_name,
-          start_page_link: Rails.application.routes.url_helpers.prison_prisoner_early_allocations_path(
+          start_page_link: Rails.application.routes.url_helpers.prison_prisoner_early_allocations_url(
             prison_id: prisoner.prison_id,
             prisoner_id: prisoner.offender_no),
           equip_guidance_link: EQUIP_URL).deliver_now
+
+        EmailHistory.create! nomis_offender_id: prisoner.offender_no,
+                             name: pom.full_name,
+                             email: pom.email_address,
+                             event: EmailHistory::SUITABLE_FOR_EARLY_ALLOCATION,
+                             prison: prisoner.prison_id
       end
     end
   end
--- a/app/jobs/suitable_for_early_allocation_email_job.rb
+++ b/app/jobs/suitable_for_early_allocation_email_job.rb
@@ -5,34 +5,34 @@ class SuitableForEarlyAllocationEmailJob < ApplicationJob
 
   EQUIP_URL = 'https://equip-portal.rocstac.com/CtrlWebIsapi.dll/?__id=webDiagram.show&map=0%3A9A63E167DE4B400EA07F81A9271E1944&dgm=4F984B45CBC447B1A304B2FFECABB777'
 
-  def perform
-    offenders = EarlyAllocation.suitable_offenders_pre_referral_window
+  def perform(offender_no)
+    allocation = Allocation.where(nomis_offender_id: offender_no).first
+    return if allocation.nil?
 
-    offenders.each do |offender_no|
-      allocation = Allocation.where(nomis_offender_id: offender_no).first
-      next if allocation.nil?
+    prisoner = OffenderService.get_offender(offender_no)
 
-      prisoner = OffenderService.get_offender(offender_no)
-      if !prisoner.nil? && prisoner.within_early_allocation_window?
+    if !prisoner.nil? && prisoner.within_early_allocation_window?
 
-        already_emailed = EmailHistory.where(nomis_offender_id: offender_no, event: EmailHistory::SUITABLE_FOR_EARLY_ALLOCATION)
+      already_emailed = EmailHistory.where(
+        nomis_offender_id: offender_no,
+        event: EmailHistory::SUITABLE_FOR_EARLY_ALLOCATION).where('created_at > ?', prisoner.sentence_start_date)
 
-        if already_emailed.empty?
-          pom = PrisonOffenderManagerService.get_pom_at(prisoner.prison_id, allocation.primary_pom_nomis_id)
-          EarlyAllocationMailer.review_early_allocation(
-            email: pom.email_address,
-            prisoner_name: prisoner.full_name,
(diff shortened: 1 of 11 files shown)


spec/jobs/suitable_for_early_allocation_email_job_spec.rb
```diff
@@ -0,0 +1,114 @@
...
+      end
+
+      it 'sends email' do
+        create(:early_allocation, created_at: Time.zone.today - 9.months, updated_at: Time.zone.today - 9.months,
+               created_within_referral_window: false, nomis_offender_id:  offender.offender_no)
+
+        expect_any_instance_of(EarlyAllocationMailer).to receive(:review_early_allocation).with(
+          email: pom.email_address,
+          prisoner_name: offender.full_name,
+          start_page_link: "/prisons/#{offender.prison_id}/prisoners/#{offender.offender_no}/early_allocations",
```
ollietreend, line 106: It looks like the start page link only holds the URL path right now (e.g. `/prisons/AAA/prisoners/BBB/early_allocations`) – but is missing the hostname from the beginning (e.g. `https://example.com`) so will not be a clickable link for users.

It looks like we should be able to use `prison_prisoner_early_allocations_url` rather than `prison_prisoner_early_allocations_path` and it give us a full absolute URL.

ollietreend (commented):
Only one thing from me – otherwise this looks good 👍

app/jobs/suitable_for_early_allocation_email_job.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+class SuitableForEarlyAllocationEmailJob < ApplicationJob
+  queue_as :mailer
+
+  EQUIP_URL = 'https://equip-portal.rocstac.com/CtrlWebIsapi.dll/?__id=webDiagram.show&map=0%3A9A63E167DE4B400EA07F81A9271E1944&dgm=4F984B45CBC447B1A304B2FFECABB777'
+
+  def perform
+    offenders = EarlyAllocation.where(created_within_referral_window: false).where.not(outcome: 'ineligible').pluck(:nomis_offender_id).uniq
+
```
starswan, line 10: It would be nice to make this a scope, so that EarlyAllocation encapsulates this behaviour rather than the client knowing about it - it's nice if 'outcome' is considered private to EarlyAllocation even if it's not enforceable by rails

app/jobs/suitable_for_early_allocation_email_job.rb
```diff
@@ -0,0 +1,40 @@
...
+    offenders.each do |offender_no|
+      allocation = Allocation.where(nomis_offender_id: offender_no).first
+      next if allocation.nil?
+
+      prisoner = OffenderService.get_offender(offender_no)
+      if !prisoner.nil? && eligible_today?(prisoner)
```
starswan, line 16: I worry about these edge-valued tests - if we miss a day for any reason, then the test will be false and the mail won't get sent. We have an EmailHistory model now which maybe we should be using here? - wonder if we could use it to make the test threshold-based instead?

deploy/staging/cron-early-allocation-suitability-email-job.yaml
```diff
@@ -0,0 +1,52 @@
...
+  concurrencyPolicy: Forbid
+  successfulJobsHistoryLimit: 3
+  failedJobsHistoryLimit: 3
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          containers:
+            - name: early-allocation-suitability-email-job
+              image: 754256621582.dkr.ecr.eu-west-2.amazonaws.com/offender-management/offender-management-allocation-manager:testing
```
starswan, line 16: needs to be 'latest' not 'testing'

spec/jobs/suitable_for_early_allocation_email_job_spec.rb
```diff
@@ -0,0 +1,114 @@
...
+
+        expect_any_instance_of(EarlyAllocationMailer).not_to receive(:review_early_allocation)
+        described_class.perform_now
+      end
+    end
+
+    context 'when offender is not allocated to a POM' do
+      let(:release_date) { Time.zone.today + 17.months }
+
+      before do
```
starswan, line 49: It's more idiomatic to not create data rather than destroy things like this - the context above probably should say 'with an allocation'

spec/jobs/suitable_for_early_allocation_email_job_spec.rb
theKHutDeveloper: Done

deploy/staging/cron-early-allocation-suitability-email-job.yaml
theKHutDeveloper: Done

spec/jobs/suitable_for_early_allocation_email_job_spec.rb
theKHutDeveloper: Done

app/jobs/suitable_for_early_allocation_email_job.rb
theKHutDeveloper: Hopefully the changes I've made fixes this

starswan (approved):

app/jobs/suitable_for_early_allocation_email_job.rb
```diff
@@ -0,0 +1,39 @@
...
+    offenders = EarlyAllocation.suitable_offenders_pre_referral_window
+
+    offenders.each do |offender_no|
+      allocation = Allocation.where(nomis_offender_id: offender_no).first
+      next if allocation.nil?
+
+      prisoner = OffenderService.get_offender(offender_no)
+      if !prisoner.nil? && prisoner.within_early_allocation_window?
+
+        already_emailed = EmailHistory.where(nomis_offender_id: offender_no, event: EmailHistory::SUITABLE_FOR_EARLY_ALLOCATION)
```
ollietreend, line 18: Not sure what the answer is: but how do we make sure that we don't skip sending the email again when the prisoner re-enters prison under a new sentence?

Right now, there will be an old email in the `EmailHistory`, so we'd skip sending to the POM for newer sentences.

app/jobs/suitable_for_early_allocation_email_job.rb
```diff
@@ -0,0 +1,39 @@
...
+
+class SuitableForEarlyAllocationEmailJob < ApplicationJob
+  queue_as :mailer
+
+  EQUIP_URL = 'https://equip-portal.rocstac.com/CtrlWebIsapi.dll/?__id=webDiagram.show&map=0%3A9A63E167DE4B400EA07F81A9271E1944&dgm=4F984B45CBC447B1A304B2FFECABB777'
+
+  def perform
+    offenders = EarlyAllocation.suitable_offenders_pre_referral_window
+
+    offenders.each do |offender_no|
```
ollietreend, line 11: I'm wondering whether a more resilient approach would be to schedule one job per offender.

I'm thinking the job itself could effectively be whatever's inside this `offenders.each` block, with the job accepting `offender_no` as an argument. And then we could move the call to `EarlyAllocation.suitable_offenders_pre_referral_window` into the rake task, and have it schedule one job for each offender.

My reason for preferring one job per offender, rather than one job for all offenders, is that we can avoid one bad offender stopping the other emails from going out. For example, if one offender mid-way through the list is broken – e.g. either the Prison API or GOV.UK Notify raises an error, and it means we can't send the email – the job will fail only for that one offender. All other offenders are unaffected because they're done in independent jobs. And any failed offenders will be retried independently, rather than needing to loop through them all again.

app/jobs/suitable_for_early_allocation_email_job.rb
theKHutDeveloper: Completely agree, I didn't think of this

app/jobs/suitable_for_early_allocation_email_job.rb
theKHutDeveloper: I don't disagree with this

ollietreend (approved):

PR #1471, opened 2020-12-11
M0-196 performance improvements

starswan:
...maybe


app/services/hmpps_api/api_deserialiser.rb
```diff
@@ -9,13 +9,5 @@ def deserialise_many(memory_model_class, payload_list)
         }
```
ollietreend, line 9: Is there any value in also losing the `ApiDeserialiser#deserialise_many` method? And then we can get rid of the whole class.

app/services/hmpps_api/prison_api/offender_api.rb
```diff
@@ -89,16 +88,12 @@ def self.get_bulk_sentence_details(booking_ids)
           client.post(route, booking_ids)
         }
 
-        data.each_with_object({}) { |record, hash|
-          next unless record.key?('bookingId')
```
ollietreend, line 93: It looks like we've lost the guard clause which filtered out records that were missing a `bookingId`.

Not sure if this is important / needed, or if it was unnecessary? Was it purposefully removed?

ollietreend (commented):
This looks like it could be helpful for performance.

From what I can tell – aside from some code refactoring – the main changes in this PR are:

- Increase Prison API `FETCH_SIZE` from 200 to 300 offenders per result – resulting in less calls to the API when listing offenders in a prison
- Switch off asset debugging – to speed up page loads in development mode

Is that right?

app/services/hmpps_api/prison_api/offender_api.rb
starswan: yes - the bookingId field is mandatory from the API.

starswan:
> This looks like it could be helpful for performance.
> 
> From what I can tell – aside from some code refactoring – the main changes in this PR are:
> 
>     * Increase Prison API `FETCH_SIZE` from 200 to 300 offenders per result – resulting in less calls to the API when listing offenders in a prison
> 
>     * Switch off asset debugging – to speed up page loads in development mode
> 
> 
> Is that right?

I think maybe the FETCH_SIZE thing is a red-herring and can probably go back - it's useful that our tests don't depend on it being set to 200 though. The main thing was removing a respond_to? in a loop, which had some impact with 1000+ offenders (but not as much as I hoped it would)

app/models/prison.rb
```diff
@@ -50,7 +50,11 @@ def each
 
   class OffenderEnumerator
     include Enumerable
-    FETCH_SIZE = 200 # How many records to fetch from nomis at a time
+    # How many records to fetch from nomis at a time
+    # optimum value for this is unknown - bigger sounds good, but doesn't seem
+    # to make much of an impact. Smaller is clearly bad as it results in multiple
+    # round-trips to eveything - initial API, plus CaseInformation plus temp_movements + arrival dates
```
ollietreend, line 56: Typo:
```suggestion
    # round-trips to everything - initial API, plus CaseInformation plus temp_movements + arrival dates
```

PR #1472, opened 2020-12-11
MO-11: Early Allocations start page

ollietreend:
Jira ticket: MO-11

This Pull Request adds the Early Allocations start page, and updates the prisoner profile page to reflect the new workflow in the status message and links.

### Change to prisoner profile page

The 'Early allocation referral' column now has a new name (used to be called 'Early allocation status'), and new status messages and links:

Screenshot 2020-12-11 at 17 45 12

### Early Allocation start page

Screenshot 2020-12-11 at 17 48 38

### DIFF
--- a/app/controllers/early_allocations_controller.rb
+++ b/app/controllers/early_allocations_controller.rb
@@ -3,6 +3,10 @@
 class EarlyAllocationsController < PrisonsApplicationController
   before_action :load_prisoner
 
+  def index
+    @early_allocations = EarlyAllocation.where(offender_id_from_url).order(created_at: :desc)
+  end
+
   def new
     case_info = CaseInformation.find_by offender_id_from_url
     @early_assignment = case_info.early_allocations.new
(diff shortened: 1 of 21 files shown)


app/controllers/early_allocations_controller.rb
```diff
@@ -61,7 +65,7 @@ def discretionary
   end
 
   def show
-    @early_assignment = EarlyAllocation.where(offender_id_from_url).last
+    @early_assignment = EarlyAllocation.where(offender_id_from_url).order(created_at: :desc).limit(1).first
```
starswan, line 68: limit(1) is not required here, as first already does it

app/helpers/early_allocation_helper.rb
```diff
@@ -24,4 +72,11 @@ def active_status(early_allocation)
       'Not eligible'
     end
   end
+
+private
+
+  def get_most_recent(early_allocations)
```
starswan, line 78: it is considered poor practice to perform queries in view helpers - surely we already have the data required, and in the correct order? - so early_allocations.last would work here?

starswan (approved):

PR #1473, opened 2020-12-14
MO-308 local delivery unit model

starswan:


app/models/local_delivery_unit.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+# This is the new (2021) LDU - Local Delivery Unit.
+# The old (incorrectly named) one will be removed in February 2021 after the PDU/LDU go live
+class LocalDeliveryUnit < ApplicationRecord
+  VALID_COUNTRIES = ['England', 'Wales'].freeze
+
+  validates :country, inclusion: { in: VALID_COUNTRIES }
+  validates :code, presence: true, uniqueness: true, format: { with: Team::TEAM_LDU_CODE_REGEX }
```
ollietreend, line 9: It might be nice to duplicate the validation regex for LDU code format into this model. That way, this new `LocalDeliveryUnit` model will be completely independent of the old LDU/Team models, rather than building in a dependency. Ideally they wouldn't know about each other.

app/models/local_delivery_unit.rb
```diff
@@ -0,0 +1,12 @@
...
+  validates :name, presence: true
+  validates :email, presence: true, 'valid_email_2/email': true
```
ollietreend, line 11: It could be nice to implement auto-trimming of email addresses here. In fact, a simple gem such as [`auto_strip_attributes`](https://github.com/holli/auto_strip_attributes) might be a nice way to abstract it – to auto-trim the name and code, too.

This could make a nice re-usable pattern for MO-266 to follow.

ollietreend (commented):
A couple of things to consider – but both entirely optional, so no action necessarily needed.

ollietreend (approved):

spec/features/admin_feature_spec.rb
```diff
@@ -108,6 +108,26 @@
...
+          click_link 'New Localdeliveryunit'
+          sleep 1
+          fill_in 'Code', with: 'TN01'
+          fill_in 'Name', with: 'Thing'
+          fill_in 'Email address', with: 'test@example.com'
+          select 'England'
+          click_button 'Create Local delivery unit'
+          sleep 1
+        }.to change(LocalDeliveryUnit, :count).by(1)
+        sleep 20
```
ollietreend, line 127: Are any of these `sleep`s needed?

The `sleep 20` in particular seems like it might be a left-over from debug/dev?

PR #1478, opened 2020-12-16
MO-297 use new ldu data

starswan:
As part of the LDU/PDU changes, Teams will be moved under the correct LDUs - so there will be no more need for the 'shadow team' concept with all the pian that brought this project.

There will be a complete 'new' set of LDUs (which are actually LocalDeliveryUnits) so in a clean break, a new table of LocalDeliveryUnits has been created with an 'enabled' flag so that they can be created in advance of the rollout. 
The first 'new' LDU is Kent going live on 9th January, with the rest following on 29th. After that we should be able to clean up the old team mappings, and delete the 'Team' and 'LocalDivisionalUnit' models along with the import job from the NART report (and decommision the email address associated with it also)

### DIFF
--- a/app/jobs/process_delius_data_job.rb
+++ b/app/jobs/process_delius_data_job.rb
@@ -54,12 +54,19 @@ private
         com_name: delius_record.offender_manager,
         crn: delius_record.crn,
         tier: map_tier(delius_record.tier),
+        local_delivery_unit: map_ldu(delius_record.ldu_code),
         team: team,
+        team_name: team&.name,
         case_allocation: delius_record.service_provider,
         welsh_offender: map_welsh_offender(delius_record.ldu_code),
         mappa_level: map_mappa_level(delius_record.mappa_levels)
       )
     end
+  end
+
+  # map the LDU regardless of enabled switch, but only expose it from offender when enabled
+  def map_ldu(ldu_code)
+    LocalDeliveryUnit.find_by(code: ldu_code)
   end
 
   def map_team(team_code)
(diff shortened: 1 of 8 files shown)


app/jobs/process_delius_data_job.rb
```diff
@@ -54,14 +54,20 @@ def map_delius_to_case_info(delius_record)
...
+        team_name: team&.name,
         case_allocation: delius_record.service_provider,
         welsh_offender: map_welsh_offender(delius_record.ldu_code),
         mappa_level: map_mappa_level(delius_record.mappa_levels)
       )
     end
   end
 
+  def map_ldu(ldu_code)
+    LocalDeliveryUnit.find_by(enabled: true, code: ldu_code)
```
ollietreend, line 68: It might be nicer to drop the `enabled: true` condition here, and instead implement it in the `OffenderBase.ldu` private method.

That way, the new LDU mappings will still be captured, and can be 'switched on' immediately by flipping the `enabled` switch.

Otherwise, with the current implementation, we'd need to enable the LDU and then wait overnight for this process job to run and add the new LDU mapping.

app/models/hmpps_api/offender_base.rb
```diff
@@ -195,6 +195,10 @@ def within_early_allocation_window?
 
   private
 
+    def ldu
+      @case_information&.local_delivery_unit || @case_information&.local_divisional_unit
```
ollietreend, line 199: See previous comment ^^^ we could perform a check here and only use the new Local Delivery Unit if it's `.enabled?`

---

Also: it'd be nice to put a comment on this method highlighting that this is where the 'fallback' happens.

Without a comment, it could look confusing since the two model names are deceptively similar – at a skim, it might look like we're returning the same thing twice. Bit of a gotcha hiding in there.

ollietreend (approved):

PR #1479, opened 2020-12-16
MO-219/221 cope with nil values for POM and COM during sort

starswan:
MO-219 and MO-221 are about sorting of the handover section not working when either POM or COM is nil.

This PR fixes this defect, and also tidies up a little in the same area - specifically removing the 'special' earliest_release_date sorting - which was a little curious as there are lots of dates (other than earliest release date) which might have had the potential for nil values.

### DIFF
--- a/app/services/bucket.rb
+++ b/app/services/bucket.rb
@@ -16,15 +16,25 @@ class Bucket
     @items << item
   end
 
-  def sort_bucket!(field, direction = :asc)
+  def sort_bucket!(field, direction)
     return unless @valid_sort_fields.include?(field)
 
-    if field == :earliest_release_date
-      @items.sort_by! { |e| e.public_send(field) || Date.new(1) }
-    else
-      @items.sort_by!(&field)
+    @items.sort! do |x, y|
+      if direction == :asc
+        a = x.public_send(field)
+        b = y.public_send(field)
+      else
+        b = x.public_send(field)
+        a = y.public_send(field)
+      end
+      # ensure that nil values sort low, but other values sort by comparison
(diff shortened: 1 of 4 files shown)


app/services/bucket.rb
```diff
@@ -16,14 +16,11 @@ def <<(item)
...
+  def sort_bucket!(field, direction)
     return unless @valid_sort_fields.include?(field)
 
-    if field == :earliest_release_date
-      @items.sort_by! { |e| e.public_send(field) || Date.new(1) }
-    else
-      @items.sort_by!(&field)
-    end
+    # convert nil into '' - date and string sorting should now be handled well
+    @items.sort_by! { |e| e.public_send(field).to_s }
```
ollietreend, line 23: This change will result in all values being sorted as strings.

### Sorting numbers

This means that numeric values will end up being sorted 'strangely':

```ruby
[2, 100, 10].map(&:to_s).sort
=> ["10", "100", "2"]
```

rather than what a human would typically expect:

```ruby
[2, 100, 10].sort
=> [2, 10, 100]
```

I don't think we currently have any columns with numeric values. It does feel like a bit of a 'UX smell' though – we're replacing one sorting 'quirk' for another – but maybe this is good enough for now? Otherwise I'd suggest we implement [a natural sorting algorithm](https://github.com/rwz/natural_sort).

ollietreend (approved):

app/services/bucket.rb
ollietreend: Resolved with the new sorting algorithm in 1bcbb82 – numeric values will now sort.

PR #1482, opened 2020-12-17
Mo 234 view previous early allocation forms

Katy600:
This PR creates an uneditable, html version of the form that was entered by a previous POM.

It is a stand alone page that will be used by [MO-11 View Early Assessment start page](https://dsdmoj.atlassian.net/browse/MO-11) and is displayed when a user(POM) clicks on a link. 
It allows the current pom to view all the previous answers that had been entered.

[View early allocations form design here ](https://hmpps-moic-staging.herokuapp.com/view-past-ea-form)

### DIFF
--- a/app/views/early_allocations/show.html.erb
+++ b/app/views/early_allocations/show.html.erb
@@ -1,0 +1,82 @@
+<div class="govuk-width-container">
+
+  <%= render :partial => "/shared/backlink", :locals => { :page => @referrer || root_path } %>
+
+  <main class="govuk-main-wrapper " id="main-content" role="main">
+    <dl class="govuk-grid-row">
+
+
+      <div class="govuk-grid-column-two-thirds">
+        <h1 class="govuk-heading-l">View previous early allocation assessment</h1>
+
+        <h2 class="govuk-heading-m">Summary</h2>
+
+        <table class="govuk-table">
+          <thead class="govuk-table__head">
+          <tr class="govuk-table__row">
+            <th id="assessment-date-label" scope="col" class="govuk-table__header app-custom-class">Assessment date</th>
+            <th id="outcome-label" scope="col" class="govuk-table__header app-custom-class">Outcome</th>
+            <th id="pom-name-label" scope="col" class="govuk-table__header app-custom-class">POM name</th>
+          </tr>
+          </thead>
+          <tbody class="govuk-table__body">
+          <tr class="govuk-table__row">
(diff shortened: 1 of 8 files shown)


spec/factories/early_allocations.rb
```diff
@@ -18,6 +18,10 @@
 
     association :case_information
 
+    trait :eligible do
+
+    end
```
ollietreend, line 23: Adding this trait a nice touch – I like it. But it might confuse people if they don't realise that the 'default' state of this factory is to be eligible, and therefore this trait needs to do 'nothing'.

It might be nice to add a simple one-liner comment in here to explain why this trait changes nothing.

Something along the lines of "Do nothing – eligible is the default outcome" should do the trick.

spec/views/early_allocations/show.html.erb_spec.rb
```diff
@@ -0,0 +1,116 @@
...
+  context 'when reviewing an early assessment' do
+    before do
+      assign(:case_information, build(:case_information))
+      assign(:offender, build(:offender, sentence: attributes_for(:sentence_detail)))
+      assign(:early_assignment, create(:early_allocation, :eligible, created_at: '05/11/2021', created_by_firstname: 'Olivia', created_by_lastname: 'Mann'))
+
+      render
+    end
+
+    let(:page) { Nokogiri::HTML(rendered) }
```
ollietreend, line 15: It looks like `let(:page)` this line could be DRYed out by pulling it to the top of this spec.

I think it's used by most of the `context` blocks – so it should only need to be declared once in the top-level `RSpec.describe` block.

app/views/early_allocations/show.html.erb
```diff
@@ -0,0 +1,83 @@
...
+    <dl class="govuk-grid-row">
+
+
+      <div class="govuk-grid-column-two-thirds">
+        <h1 class="govuk-heading-l">View previous early allocation assessment</h1>
+
+        <h2 class="govuk-heading-m">Summary</h2>
+
+        <table class="govuk-table">
+          <!-- <caption class="govuk-table__caption">Previously saved assessments</caption> -->
```
ollietreend, line 15: It looks like this HTML comment isn't needed – let's remove it 🙂

ollietreend (commented):
Great work @Katy600 – this is looking really good.

I've added a few comments – but nothing major!

Happy to talk it through if you'd like to.

ollietreend (approved):
Looks great @Katy600 🎉

PR #1495, opened 2021-01-04
Feature/early allocations

starswan:
Now the code freeze has ended, we should now merge this into main


ollietreend (approved):
All commits were already approved as part of their own PRs to merge into this branch.

This branch should therefore be safe to merge into `main`.

PR #1501, opened 2021-01-07
MO-281 upgrade faraday to 1.x

starswan:
Our copy of the faraday need updating, but it is a major version change so it is not backwards-compatible.

This will enable the neew sentry-ruby gem (amongst others)

### DIFF
--- a/spec/jobs/recalculate_handover_date_job_spec.rb
+++ b/spec/jobs/recalculate_handover_date_job_spec.rb
@@ -77,7 +77,7 @@ RSpec.describe RecalculateHandoverDateJob, type: :job do
     it 'raises an exception so the job will go into the retry queue' do
       expect {
         described_class.perform_now(offender_no)
-      }.to raise_error(Faraday::ClientError)
+      }.to raise_error(Faraday::Error)
     end
   end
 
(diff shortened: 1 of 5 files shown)


spec/services/hmpps_api/client_spec.rb
```diff
@@ -64,7 +64,7 @@
 
       it 'raises a Faraday::ClientError error' do
         expect { client.get(route) }.
-          to raise_error(Faraday::ClientError, "the server responded with status #{status}")
+          to raise_error(Faraday::ServerError, "the server responded with status #{status}")
       end
```
ollietreend, line 68: Can we change the test description to match?

```
it 'raises a Faraday::ServerError error' do
```

spec/jobs/recalculate_handover_date_job_spec.rb
```diff
@@ -77,7 +77,7 @@
     it 'raises an exception so the job will go into the retry queue' do
       expect {
         described_class.perform_now(offender_no)
-      }.to raise_error(Faraday::ClientError)
+      }.to raise_error(Faraday::ServerError)
```
ollietreend, line 80: Technically this could be `Faraday::Error` to catch _any_ HTTP error.

Not necessarily needed, though – I think this is still clear as it is.

ollietreend (approved):

PR #1505, opened 2021-01-12
MO-200: Bug fix - debugging page breaks for offender without a sentence start date

theKHutDeveloper:
This bug also occurs in the same scenario on the prisoner show page.

The sentence start date is required when calculating the POM is RESPONSIBLE or
SUPPORTING. When there is an offender without this date the page will crash as
the app is trying to do calculations on a nil object.

The fix I have put in place is to return UNKNOWN if there is no sentence_start_date
before we do the calculation.


ollietreend (commented):
This looks good to me – it solves the immediate problem.

I'd be interested to hear @starswan's input though, because I know there was an idea about fixing it closer to the view in a Debugging Presenter. But given the fact that this has also appeared in the PrisonersController#show method, so isn't limited just to the Debugging view, it seems like fixing in the data model (as you have done in this PR) is more appropriate.

starswan (approved):

PR #1508, opened 2021-01-13
raise HTTP errors in OAuth client

starswan:
https://sentry.io/organizations/ministryofjustice/issues/2148575505/?project=5584139&query=is%3Aunresolved

OAuth client doesn't raise when HTTP errors happen, so it fails by accident

### DIFF
--- a/app/services/hmpps_api/oauth/client.rb
+++ b/app/services/hmpps_api/oauth/client.rb
@@ -7,7 +7,9 @@ module HmppsApi
 
       def initialize(host)
         @host = host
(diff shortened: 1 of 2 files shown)


ollietreend (approved):

ollietreend (commented):
@starswan it'd be nice to test this new behaviour – we have a pattern for it in the API client spec here:

https://github.com/ministryofjustice/offender-management-allocation-manager/blob/main/spec/services/hmpps_api/client_spec.rb#L35-L70

PR #1509, opened 2021-01-13
MO-114 Adds a secondary button so an SPO can allocate after updating

Katy600:
This PR does 4 things:

1. Adds an additional secondary button to the missing information page called 'Save and allocate'.
[Additional button 'Save and allocate' design](https://mojuk.invisionapp.com/console/share/YW13A7LX98/478478131)
Previously the allocation workflow on MOiC expects that users to add missing information to all cases in that tab and then proceed to make allocations as a separate task. 
This new button allows them to allocate that case to a POM straight away once they’ve added missing information they want. 
2. It removes this button from the edit journey when a user just wants to 'Update' the missing information. This is to prevent a user becoming confusing and accidentally ending up on an allocation journey instead of returning to the prisoner page they had been on previously.
3. It makes changes to some the vcr tests so that they are clicking on the label as opposed to selecting the radio button. This is because the gov design hides the radio button and therefore expects the label to be clicked. 
4. It also fixed some accessibility issues with the labels which were discovered


ollietreend (approved):
Nice work @Katy600 🎉 

And it looks like you've left the feature spec in a better state than it was when you picked up this ticket 🙂

PR #1511, opened 2021-01-13
MO-266: Remove whitespace on VLO emails

theKHutDeveloper:
If there are leading or trailing spaces around an email address entered on to the VLO form, the form will generate an error message. This commit removes the whitespaces before validation so errors won't occur for this issue.

### DIFF
--- a/app/models/victim_liaison_officer.rb
+++ b/app/models/victim_liaison_officer.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 class VictimLiaisonOfficer < ApplicationRecord
+  auto_strip_attributes :email
   # Other version fields (user_first_name and user_last_name) filled in by controller
   has_paper_trail meta: { nomis_offender_id: :nomis_id_for_paper_trail }
 
--- a/spec/features/prisoner_info_spec.rb
+++ b/spec/features/prisoner_info_spec.rb
@@ -68,7 +68,8 @@ feature 'View a prisoner profile page', :allocation do
 
         # fill in missing fields and submit
         fill_in 'Last name', with: 'Smith'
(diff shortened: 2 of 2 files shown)


app/models/victim_liaison_officer.rb
```diff
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 class VictimLiaisonOfficer < ApplicationRecord
+  before_validation :omit_whitespace_on_email
```
ollietreend, line 4: There is a generic/reusable pattern provided by a gem in the app which can do this without needing a custom method in the model. It was added with MO-308, so is only relatively new. I should probably have left a note in this ticket to point that out – sorry!

Here's an example how it's used in the `LocalDeliveryUnit` model:

https://github.com/ministryofjustice/offender-management-allocation-manager/blob/c763d5d961834961b9adbfc0c9148b54eac12c6d/app/models/local_delivery_unit.rb#L9

In this PR, it should be possible to replace the `before_validation` with `auto_strip_attributes`, and then we won't need the custom method `omit_whitespace_on_email` in this model. One less line of code for us to maintain 😄

spec/features/prisoner_info_spec.rb
```diff
@@ -68,7 +68,8 @@
 
         # fill in missing fields and submit
         fill_in 'Last name', with: 'Smith'
-        fill_in 'Email address', with: 'jim.smith@hotmail.com'
+        # email has leading and trailing whitespaces, that are removed before validation
+        fill_in 'Email address', with: ' jim.smith@hotmail.com '
```
ollietreend, line 72: Nice test 👍

ollietreend (approved):

PR #1515, opened 2021-01-15
Mo217 pom handover like spo

starswan:
This change makes the POM handover view almost identical to the SPO one - apart from the POM Name which is replaced by the rensponsibility field

### DIFF
--- a/app/controllers/caseload_handovers_controller.rb
+++ b/app/controllers/caseload_handovers_controller.rb
@@ -2,12 +2,18 @@
 
 class CaseloadHandoversController < PrisonStaffApplicationController
   def index
-    offenders = @pom.pending_handover_offenders
+    bucket = Bucket.new([:last_name, :handover_start_date, :responsibility_handover_date,
+                         :pom_responsibility, :allocated_com_name, :case_allocation])
+    @pom.pending_handover_offenders.each { |o| bucket << o }
+    if params['sort']
+      sort_params = params['sort'].split.map { |s| s.downcase.to_sym }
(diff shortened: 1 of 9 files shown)


spec/features/spo_handover_cases_feature_spec.rb
```diff
@@ -67,71 +68,106 @@
...
-        expect(page).to have_css(".offender_row_#{index}", text: name)
+      it 'stops staff without the SPO role from viewing the SPO page' do
+        visit prison_handovers_path(prison)
+        expect(page).to have_current_path('/401')
       end
 
-      click_link('POM')
-      check_poms.reverse.each_with_index do |name, index|
-        expect(page).to have_css(".offender_row_#{index}", text: name)
+      it 'can load the POM handovers page', :js do
```
ollietreend, line 115: Looks like this feature test now covers both the SPO handover page and the POM handover page.

It might be worth making this explicit/official by renaming the spec file, and changing the name on the root `feature` block. Otherwise it's not obvious that this file also tests a POM feature without looking for it.

ollietreend: It looks like there's also some (very minimal) coverage of the POM handovers feature in `pom_caseload_feature_spec.rb` – would it be more logical to have it together in one place?

I think our expected user journeys have changed slightly following recent design changes – so could be a case of these feature specs needing to 'catch up' with that.

ollietreend (approved):
This looks good – nice use of existing buckets to remove custom functionality 👍

I've left a comment with something to think about. I have a feeling we could restructure the feature specs a little to make them more 'obvious'.

PR #1520, opened 2021-01-18
Fix 'last name' title case discrepancy in test factories

ollietreend:
We had some test factories which were causing flaky test failures, due to a discrepancy between lower/uppercase letters in 'last name' fields.

This is an annoying side-effect of the fact that POM names in NOMIS are always UPPERCASE. Whilst reading them into the application, we convert them to Title Case using Rails' `.titleize` method.

However for surnames such as 'McDonald', the Title Case conversion transforms this to 'Mcdonald' (note the lowercase 'd').

To counteract this, our test factories need to fake names in the same format – i.e. they _also_ need to titleize names.

Without this, some tests sometimes fail due to 'McDonald' not matching 'Mcdonald'. In reality this was just a bug in the factories.

### DIFF
--- a/spec/factories/allocations.rb
+++ b/spec/factories/allocations.rb
@@ -7,7 +7,9 @@ FactoryBot.define do
     end
 
     created_by_name do
-      "#{Faker::Name.first_name} #{Faker::Name.last_name}"
+      # The last name is titleized after it's received from the API, e.g. "McDonald" becomes "Mcdonald"
+      # So we also .titleize the last name here to avoid breaking tests
+      "#{Faker::Name.first_name} #{Faker::Name.last_name.titleize}"
     end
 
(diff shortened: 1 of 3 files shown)


spec/factories/allocations.rb
```diff
@@ -7,7 +7,7 @@
     end
 
     created_by_name do
-      "#{Faker::Name.first_name} #{Faker::Name.last_name}"
+      "#{Faker::Name.first_name} #{Faker::Name.last_name.titleize}"
```
starswan, line 10: It would be good to put a comment somewhere to explain all these - otherwise looks good

starswan (approved):
It would be good to put a comment somewhere to explain all these - otherwise looks good

PR #1521, opened 2021-01-19
MO-75 check for COM supporting when displaying COM error on profile page

starswan:


app/services/handover_date_service.rb
```diff
@@ -1,15 +1,16 @@
...
     end
   end
 
-  RESPONSIBLE = Responsibility.new 'Responsible', true, 'Custody'
-  SUPPORTING = Responsibility.new 'Supporting', false, 'Community'
-  UNKNOWN = Responsibility.new 'Unknown', false, 'Unknown'
+  RESPONSIBLE = Responsibility.new 'Responsible', true, 'Custody', false
+  SUPPORTING = Responsibility.new 'Supporting', false, 'Community', true
+  # 'Null Object' pattern
+  UNKNOWN = Responsibility.new 'Unknown', false, 'Unknown', false
```
ollietreend, line 13: It would be nice & a bit cleaner if we could refactor this service to also include COM responsibility as part of the response. That way we wouldn't have to call the `.com_supporting?` method directly in `OffenderBase#needs_com?`.

It looks like we're half way there – the `Responsibility` object now has a `.community?` method. But it looks like the values are hard-coded based on whether the POM is `RESPONSIBLE` or `SUPPORTING` rather than being the result of the calculation in `.com_supporting?`.

It would be great if `.com_supporting?` could be a private method, so our only public interface continues to be `.handover` and `. calculate_pom_responsibility`.

starswan: That's funny - half of me was thinking that very thing

ollietreend: We've just talked through some ideas and came up with:

- HandoverData should be the single returned object which combines handover dates and responsibility (i.e. remove Responsibility sub-object and promote it to be top-level)
- Replace: HandoverData.responsibility.custody? / .community?
- With: HandoverData.community.responsible? / .supporting? AND .custody.responsible? / .supporting?
- This new model supports tri-state community involvement (not involved, supporting, responsible)
- Avoid passing back strings / descriptions whatsoever – move these to a view helper (where they belong)
- Don't worry about Law of Demeter :-) this is two booleans, so simple enough to not be a problem
- 'Unknown' responsibility == neither custody nor community are responsible (all answers are false) – this is a 'null object' for working around bad data which has made its way into the service

app/models/hmpps_api/offender_base.rb
```diff
@@ -217,8 +217,16 @@ def within_early_allocation_window?
...
 
+    def needs_com_and_ldu_is_uncontactable?
+      needs_com? && sentenced? && allocated_com_name.nil? && ldu_email_address.nil?
+    end
+
   private
 
+    def needs_com?
+      handover.community.responsible? || handover.community.supporting?
+    end
```
ollietreend, line 228: The method names here are slightly misleading:

- `needs_com?` = should a COM be involved in the case?
- `needs_com_and_ldu_is_uncontactable?` = same as `needs_com?` **but also checks if a COM is already allocated** – as well as if the LDU is contactable

Given that they share the beginning of their name, it seems a bit weird that the addition of `_and_ldu_is_uncontactable` to the method name **also adds a check for COM allocation**.

It's the sort of subtle difference that might catch someone out.

However: I recognise the need for both methods, and haven't yet come up with a better name. So just something to think about.

app/models/hmpps_api/offender_base.rb
```diff
@@ -231,7 +239,7 @@ def ldu
     end
 
     def handover
-      @handover ||= if pom_responsibility.custody?
+      @handover ||= if pom_responsibility.responsible?
                       HandoverDateService.handover(self)
                     else
                       HandoverDateService::NO_HANDOVER_DATE
```
ollietreend, line 245: It seems a shame to still be blanking out handover dates when the Community is responsible.

Not sure it's something to tackle in this PR, but just pointing it out. Maybe one to raise and clean up later if this behaviour is no longer desirable.

ollietreend (commented):
@starswan This is looking much nicer with the responsibility refactor 👍

I've added comments to a few things which stood out to me. Not sure any changes are necessary – just thought I'd point them out as things to think about.

ollietreend (approved):

app/models/hmpps_api/offender_base.rb
```diff
@@ -217,8 +217,16 @@ def within_early_allocation_window?
       earliest_date.present? && earliest_date <= Time.zone.today + 18.months
     end
 
+    def needs_com_and_ldu_is_uncontactable?
+      needs_com? && sentenced? && allocated_com_name.nil? && ldu_email_address.nil?
```
ollietreend, line 221: ## ⚠️ Potential bug alert 🐛 

We might want to use `.empty?` instead of `.nil?` here to avoid accidentally letting empty strings through – for COM name and LDU email address.

(Although I realise that the new LDU model does require a valid email address... so maybe this is a moot point.)

PR #1523, opened 2021-01-20
MO-49: Bug Fix - Responsibility handover should not take place before the start of handover

theKHutDeveloper:
This applies to the scenario that the home detention curfew eligibility date is entered incorrectly too early.

This resulted in the responsibility handover is taking place before start of handover.

This commit fixes this by setting the responsibility handover to the same date as the start of handover in this scenario

### DIFF
--- a/app/services/handover_date_service.rb
+++ b/app/services/handover_date_service.rb
@@ -39,7 +39,9 @@ class HandoverDateService
       responsibility = responsibility_override(offender)
       if responsibility.nil?
         handover_date, reason = nps_handover_date(offender)
-        HandoverData.new nps_responsibility(offender, handover_date), nps_start_date(offender), handover_date, reason
+        start_date = nps_start_date(offender)
+        handover_date = start_date if start_date > handover_date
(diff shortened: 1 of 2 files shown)


app/services/handover_date_service.rb
```diff
@@ -39,6 +39,8 @@ def self.handover(raw_offender)
       responsibility = responsibility_override(offender)
       if responsibility.nil?
         handover_date, reason = nps_handover_date(offender)
+        start_date = nps_start_date(offender)
+        handover_date = start_date if start_date > handover_date
         HandoverData.new nps_responsibility(offender, handover_date), nps_start_date(offender), handover_date, reason
```
ollietreend, line 44: Optimisation: you should be able to avoid the second call to `nps_start_date(offender)` in the final line (`HandoverData.new...`) because you already hold the calculated start date in the variable `start_date`

ollietreend (approved):
@theKHutDeveloper I've added one comment with a small optimisation – but other than that, this looks good 👍

PR #1526, opened 2021-01-21
MO-193: Prevents pom name from displaying in DPS quick look screen after an offenders sentence has ended

Katy600:
POM names have remained on DPS quick look screen after an offender sentence has ended which has caused some confusion. This is because currently an offender can not be unallocated from a POM. This is PR is just a quick fix.

### DIFF
--- a/app/controllers/api/allocation_api_controller.rb
+++ b/app/controllers/api/allocation_api_controller.rb
@@ -4,7 +4,9 @@ module Api
   class AllocationApiController < Api::ApiController
     def show
       render_404('Not ready for allocation') && return if allocation.nil?
-      render_404('Not allocated') && return if allocation.primary_pom_nomis_id.nil?
+
+      offender = OffenderService.get_offender(offender_number)
+      render_404('Not allocated') && return unless allocation.active? && offender.inside_omic_policy?
 
       render json: allocation_as_json
     end
--- a/spec/controllers/api/allocation_api_controller_spec.rb
+++ b/spec/controllers/api/allocation_api_controller_spec.rb
@@ -1,0 +1,75 @@
+require 'rails_helper'
+
+RSpec.describe Api::AllocationApiController, :allocation, type: :controller do
+  describe '#show' do
+    let(:rsa_private) { OpenSSL::PKey::RSA.generate 2048 }
+    let(:stub_url) { "#{api_host}/secure/offenders/nomsNumber/#{offender.fetch(:offenderNo)}/prisonOffenderManager" }
+    let(:api_host) { Rails.configuration.community_api_host }
+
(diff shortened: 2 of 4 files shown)


spec/controllers/api/allocation_api_controller_spec.rb
```diff
@@ -0,0 +1,68 @@
+require 'rails_helper'
+
+RSpec.describe Api::AllocationApiController, type: :controller do
```
ollietreend, line 3: There is also a spec file testing existing API functionality located here:

```
spec/api/allocation_api_spec.rb
```

The format of the spec is slightly different from what I'm used to – I think it's also used to generate swagger documentation. It might be worth checking with @starswan to see if this test should actually be consolidated into that one...

app/controllers/api/allocation_api_controller.rb
```diff
@@ -6,6 +6,9 @@ def show
       render_404('Not ready for allocation') && return if allocation.nil?
       render_404('Not allocated') && return if allocation.primary_pom_nomis_id.nil?
 
+      offender = OffenderService.get_offender(offender_number)
+      render_404('Not sentenced') && return if not offender.sentenced?
```
ollietreend, line 10: This introduces a new error response body – `"Not sentenced"`.

Since this API is consumed by DPS, I wonder how risky it is to do this.

`"Not allocated"` could also be valid to use here, which means we don't introduce a new response body.

But: again, it's worth checking with @starswan because he'll understand this more than I do!

ollietreend (approved):

app/controllers/api/allocation_api_controller.rb
```diff
@@ -4,7 +4,9 @@ module Api
   class AllocationApiController < Api::ApiController
     def show
       render_404('Not ready for allocation') && return if allocation.nil?
-      render_404('Not allocated') && return if allocation.primary_pom_nomis_id.nil?
+
+      offender = OffenderService.get_offender(offender_number)
+      render_404('Not allocated') && return if allocation.primary_pom_nomis_id.nil? || !offender.inside_omic_policy?
```
ollietreend, line 9: This line might be easier to read if rewritten as:

```suggestion
      render_404('Not allocated') && return unless allocation.active? && offender.inside_omic_policy?
```

I think this should be functionally equivalent (the tests should fail if it isn't!), however it does a couple of things:
- Uses the method `allocation.active?` which does exactly what you need, but is easier to read than `allocation.primary_pom_nomis_id.nil?`
- Reorders the `if` statement to become an `unless` – which in this case seems to read more clearly

(In my opinion) it makes the intention of this line clearer to understand:
> Return 404, unless the offender has an active allocation and is inside OMIC policy

PR #1527, opened 2021-01-25
MO-293 give the early allocation OaSys date question its own page

starswan:
This changes the Early Allocation journey so that there is a seperate page for the OASys date question - this helps the user bale from the from if the OaSys hasn't been completed rather than asking them 6 pointless questions

### DIFF
--- a/spec/models/early_allocation_spec.rb
+++ b/spec/models/early_allocation_spec.rb
@@ -7,8 +7,16 @@ RSpec.describe EarlyAllocation, type: :model do
     expect(build(:early_allocation)).to be_valid
   end
 
-  context 'when oasys date more than 2 years adrift' do
-    let(:ea) { build(:early_allocation_eligible_form, oasys_risk_assessment_date: Time.zone.today - 3.years) }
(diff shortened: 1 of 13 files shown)


theKHutDeveloper (approved):

spec/models/early_allocation_spec.rb
```diff
@@ -7,8 +7,8 @@
     expect(build(:early_allocation)).to be_valid
   end
 
-  context 'when oasys date more than 2 years adrift' do
-    let(:ea) { build(:early_allocation_eligible_form, oasys_risk_assessment_date: Time.zone.today - 3.years) }
+  context 'when oasys date more than 2 months adrift' do
```
ollietreend, line 10: "more than 2 months adrift" seems like odd wording – I thought the rule is "within the last 3 months"?

Also – it looks like the OASys date must be in the last 3 month _exclusive_ of exactly 3 months ago. (Hence `today - 3.months` is invalid). Is that right, or should it be inclusive of `today - 3.months`?

ollietreend (approved):

PR #1528, opened 2021-01-27
MO-359: Use new LocalDeliveryUnit model

ollietreend:
Jira ticket: MO-359

---

This PR brings MPC one step closer to fully utilising the new Local Delivery Unit model.

It adapts the two handover email jobs sent to LDUs (the 45 day upcoming email, and the chaser email) to work with both 'old' LDUs (Team/LocalDivisionalUnit, deprecated after 31/01/2021) and 'new' LDUs (LocalDeliveryUnit).

For the month of January 2021, 'old' and 'new' LDUs have been running in parallel, which is why this work is needed to allow the handover email jobs to work with either model.

From 31/01/2021, the 'old' LDUs will be deprecated and the Team/LocalDivisionalUnit functionality will need to be removed from the codebase. So the switching functionality implemented in this PR will only be needed until that code cleanup is done.

Also included in this PR:
- Allow `CaseInformation` to be valid if there's a (new) `LocalDeliveryUnit` but no (old) `Team`
- Update the ActiveAdmin 'offenders without an LDU email address' dashboard to exclude offenders who have a `LocalDeliveryUnit` but no `Team`
- Make the `LocalDeliveryUnit` test factory generate realistic LDU names (using UK counties from the `Faker` library)
- Create `LocalDeliveryUnit` records in the database seed script (`db/seeds.rb`)

### Commit history

I've tried to keep the commit history clear to show the steps taken – so feel free to read through them.

Before merging this PR, I will squash the 'refactor' commits into their respective 'main' commit. I've left them in for now as they show my working, so might be helpful for code review.

### DIFF
--- a/lib/tasks/handover_chase_email.rake
+++ b/lib/tasks/handover_chase_email.rake
@@ -5,6 +5,8 @@ require 'rake'
 namespace :handover_chase_emails do
   desc 'Send follow-up emails LDUs two weeks before handover if COM still not assigned'
   task process: :environment do
-    HandoverFollowUpJob.perform_later(Time.zone.today)
+    LocalDivisionalUnit.with_email_address.each do |ldu|
+      HandoverFollowUpJob.perform_later(ldu)
+    end
   end
 end
--- a/lib/tasks/handover_chase_email.rake
+++ b/lib/tasks/handover_chase_email.rake
@@ -5,7 +5,14 @@ require 'rake'
 namespace :handover_chase_emails do
   desc 'Send follow-up emails LDUs two weeks before handover if COM still not assigned'
   task process: :environment do
+    # Send to 'old' LDUs
+    # These will be removed in February 2021 after the PDU/LDU go live
     LocalDivisionalUnit.with_email_address.each do |ldu|
+      HandoverFollowUpJob.perform_later(ldu)
+    end
+
+    # Send to 'new' (2021) LDUs
+    LocalDeliveryUnit.enabled.each do |ldu|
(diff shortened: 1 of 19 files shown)


lib/tasks/handover_chase_email.rake
```diff
@@ -5,6 +5,8 @@ require 'rake'
 namespace :handover_chase_emails do
   desc 'Send follow-up emails LDUs two weeks before handover if COM still not assigned'
   task process: :environment do
-    HandoverFollowUpJob.perform_later(Time.zone.today)
+    LocalDivisionalUnit.with_email_address.each do |ldu|
+      HandoverFollowUpJob.perform_later(ldu)
+    end
```
starswan, line 10: shouldn't this do the new LDUs as well?

spec/rails_helper.rb
```diff
@@ -108,3 +108,6 @@
     with.library :rails
   end
 end
+
+# Set the locale for Faker gem to en-GB
+Faker::Config.locale = 'en-GB'
```
starswan, line 113: a comment why might be nice :-)

lib/tasks/handover_chase_email.rake
ollietreend: Fixed

spec/rails_helper.rb
ollietreend: I debated adding one but didn't in the end! I'll add one now..

ollietreend: Fixed

starswan (approved):

PR #1531, opened 2021-01-29
Checks which offenders are inside omic policy

Katy600:
This PR is a refactor. It takes a set of methods that together calculate if an offender is inside the omic policy - and therefore available in the service. These methods are: `over_18?`,  `sentenced?`, `immigration_case?`, `criminal_sentence?` and `convicted?`  Methods such as `sentenced?` have potentially been used incorrectly through out the code base thinking that it was checking if an offender was inside the omic service. 
This PR also tries to make some of these methods private but `sentenced?` has been used in too many places so would require a bigger refactor - I think a ticket has been made for this.

### DIFF
--- a/app/controllers/debugging_controller.rb
+++ b/app/controllers/debugging_controller.rb
@@ -45,7 +45,7 @@ private
 
   def filtered_offenders
     @unfiltered_offenders.group_by { |offender|
-      if offender.age < 18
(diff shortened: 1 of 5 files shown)


ollietreend (approved):
This looks great @Katy600. We've needed a refactor like this for a while – thanks for doing it!

PR #1542, opened 2021-02-02
Remove unused field from the Offender object

ollietreend:
The 'reception date' was never used in our application, so I've removed it from the Offender object to avoid future confusion.


starswan (approved):
One fewer difference between this and OffenderSummary...

PR #1543, opened 2021-02-02
MO-70: Email community when Welsh offender is transferred to Prescoed

theKHutDeveloper:
Set up test structure for determinate and indeterminate welsh offender at Prescoed.
Add mailer and Notify template.
Add to email history any indeterminate welsh offenders going to Prescoed
Add partial to the prisoner show page for welsh offenders going from closed to prescoed
I have also added the partial to the Allocation Information page even though it wasn't implied in the ticket to do so.
I have assumed that we do not want the email notification to remain once a COM has been assigned
Show email history in case history for prescoed welsh event


app/helpers/offender_helper.rb
```diff
@@ -44,4 +44,16 @@ def needs_com_but_ldu_is_uncontactable?(offender)
...
+    # I've assumed that the offender needs to currently be in Prescoed
+    # e.g. this is not valid if the offender has moved to a different prison
+    return false unless offender.prison_id == PrisonService::PRESCOED_CODE
+
+    # if offender has a com assigned then we should not show the notification on the prisoner profile pages
+    return false if offender.allocated_com_name.present?
+
+    email = EmailHistory.welsh_open_prescoed(offender.offender_no)
+    email.present? ? true : false
+  end
```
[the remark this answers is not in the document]
ollietreend: Also: I see what you mean about naming. (Naming is hard 😩)

The naming of this method is very specific, and that certainly feels less than ideal.

I have a feeling this notification should become more generic – e.g. "a COM is needed, but one hasn't been allocated" – followed by "was an email sent to the LDU?" or "are we missing the LDU email address?". At that point, it stops being about Prescoed and instead about OMIC rules – which feels much better.

However: again, I think that's too much to tackle in this PR. We also have Prescoed-specific code in the Handover Date Service, so you wouldn't be creating a 'new' problem by going with the approach you've used here.

The ticket I'm working on MO-394 Open Prison Rules actually has a task to clean up the naming of Prescoed-specific things to make them fit for open prisons in general. So I'll see if I can tackle it in that.

app/jobs/open_prison_transfer_job.rb
```diff
@@ -18,7 +18,21 @@ def perform(movement_json)
...
+    if offender.prison_id == PrisonService::PRESCOED_CODE && offender.welsh_offender
+      return unless offender.indeterminate_sentence?
+
+      if offender.ldu_email_address.present?
+        prescoed_send_email(offender, movement)
+
+        EmailHistory.create! nomis_offender_id: offender.offender_no, name: offender.ldu_name,
+                             email: offender.ldu_email_address,
+                             event: EmailHistory::WELSH_TRANSFER_TO_PRESCOED,
+                             prison: offender.prison_id
```
ollietreend, line 31: I think the `EmailHistory` event name would ideally be non-Prescoed specific.

The email itself is really about the LDU needing to allocate a COM to an offender who's moved into an open prison where OMIC rules apply. The email template itself makes no reference to Prescoed.

The same goes for the method name `prescoed_send_email` – I think it could be more generic, since it actually sends the email called `omic_open_prison_community_allocation`.

Finally, I think it might be nice to keep the `EmailHistory.create!` in the same method as the call to `CommunityMailer.omic_open_prison_community_allocation` – so it all happens in the same place, and it's really clear to see that the EmailHistory record directly relates to the email being sent.

app/models/email_history.rb
```diff
@@ -4,15 +4,22 @@ class EmailHistory < ApplicationRecord
...
-                                      SUITABLE_FOR_EARLY_ALLOCATION], allow_nil: false }
+                                      SUITABLE_FOR_EARLY_ALLOCATION,
+                                      WELSH_TRANSFER_TO_PRESCOED], allow_nil: false }
   validates :prison, inclusion: { in: PrisonService.prison_codes, allow_nil: false }
   validates :email, presence: true, 'valid_email_2/email': true
 
+  # offender could have been at this prison multiple times, so order by newest first
+  scope :welsh_open_prescoed, lambda { |nomis_offender_id|
+    where(nomis_offender_id: nomis_offender_id, event: EmailHistory::WELSH_TRANSFER_TO_PRESCOED).order('created_at DESC')
+  }
```
ollietreend, line 21: ### Naming

Assuming the event name is changed to avoid being Prescoed-specific, I think the scope name should reflect the name of the event.

### Within offender sentence?

Is there also a need to check whether the email was sent within the offender's current sentence?

I remember we did this in the 'might be suitable for Early Allocation' email:

https://github.com/ministryofjustice/offender-management-allocation-manager/blob/883f6c38e04d20b508d61aeb7d2e97b2bb32821c/app/jobs/suitable_for_early_allocation_email_job.rb#L16-L18

In order to do that, we need to have the `Offender` object in our context. This would be possible of the helper method `welsh_offender_from_closed_to_prescoed?` was moved into the `Offender` object itself – you'd have the sentence start date directly in your hands.

And then since the query will only be used in one place, it might not actually be worth having this as a reuable scope. The query could live directly in the Offender method itself.

ollietreend (commented):
A few comments – very happy to talk them through!

I think this is good. Mostly my comments are around the naming and location of code.

ollietreend (approved):

app/views/allocations/_email_open_prison_community_allocation.html.erb
```diff
@@ -0,0 +1,15 @@
...
+    <h2 class="moj-timeline__title">Offender transferred to an open prison</h2>
+  </div>
+
+  <p class="moj-timeline__date"><%= format_date_long(email_open_prison_community_allocation.created_at) %>
+    email sent automatically
+  </p>
+
+  <div class="moj-timeline__description">
+    <p>
+      The LDU for <%= email_open_prison_community_allocation.name %> - <%=  format_email(email_open_prison_community_allocation.email) %> - was sent an email asking them to appoint a Responsible COM.
```
ollietreend, line 12: This needs to be changed to say 'supporting COM'

ollietreend (approved):
I've just spotted one content issue – use of 'responsible' instead of 'supporting' in the case history.

But that's the final bit. My approval still stands! 🙂

PR #1544, opened 2021-02-03
Add logging to the Community API import

ollietreend:
I've added logging to the rake task which performs the nightly Community API import. It's quite verbose, but it's helpful for debugging issues and to build confidence that the job is running successfully.

Previously there was no logging, so it was hard to tell if the job was completing successfully. However, we've observed in production that CaseInformation records don't seem to be updated reliably overnight – which suggests that this script is terminating early. There are no logs in Sentry to suggest it's caused by an application level exception.

I've also added a comment to highlight a potential memory leak. This could be one route of investigation if we discover that the script doesn't complete successfully.

### Viewing the logs in Cloud Platform

To see the logs created by this script in the MOJ Cloud Platform, you'll need access to Kibana.

1. Open Kibana and go to the 'Discover' tab
2. Set the following filters on the search query:
    ```
    kubernetes.namespace_name: offender-management-production
    kubernetes.labels.job-name: recalculate-handover-dates
    ```
    This will show log entries for only this cron job.
3. Set the date/time period to at least 24 hours (since this cron job only runs once every 24 hours)
4. Optionally, put "Done" (including quotation marks) into the search text field. This will filter results to *only* show the log entry which indicates successful completion of the rake task.

### Example log output

```
[1/2] Getting offenders for prison HMP Leeds (LEI)
[1/2] Queueing jobs for 5 offenders
[1/2] [1/5] Queued T1000OH
[1/2] [2/5] Queued T1000OI
[1/2] [3/5] Queued T1000OJ
[1/2] [4/5] Queued T1000OK
[1/2] [5/5] Queued T1000OL
[2/2] Getting offenders for prison HMP Pentonville (PVI)
[2/2] Queueing jobs for 3 offenders
[2/2] [1/3] Queued T1000OM
[2/2] [2/3] Queued T1000ON
[2/2] [3/3] Queued T1000OO
Done
```

The first set of square brackets shows which prison we're on.
The second set of square brackets shows which offender we're on within that prison.

### DIFF
--- a/lib/tasks/community_api_import.rake
+++ b/lib/tasks/community_api_import.rake
@@ -5,10 +5,27 @@ namespace :community_api do
   task import: :environment do |_task|
     Rails.logger = Logger.new(STDOUT)
 
-    Prison.active.each do |prison|
-      prison.offenders.each do |offender|
+    # Avoid filling up the in-memory SQL query cache – we're going to be reading lots of database records
+    ActiveRecord::Base.connection.disable_query_cache!
+
+    prison_count = Prison.active.count
+
+    Prison.active.each_with_index do |prison, prison_index|
+      prison_log_prefix = "[#{prison_index + 1}/#{prison_count}]"
+      Rails.logger.info("#{prison_log_prefix} Getting offenders for prison #{prison.name} (#{prison.code})")
+
+      offenders = prison.offenders
+      offender_count = offenders.count
+
+      Rails.logger.info("#{prison_log_prefix} Queueing jobs for #{offender_count} offenders")
+
+      offenders.each_with_index do |offender, offender_index|
         ProcessDeliusDataJob.perform_later offender.offender_no
+        offender_log_prefix = "[#{offender_index + 1}/#{offender_count}]"
+        Rails.logger.info("#{prison_log_prefix} #{offender_log_prefix} Queued #{offender.offender_no}")
       end
     end
+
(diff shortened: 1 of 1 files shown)


lib/tasks/community_api_import.rake
```diff
@@ -4,11 +4,31 @@ namespace :community_api do
   desc 'Import data from Community API'
   task import: :environment do |_task|
     Rails.logger = Logger.new(STDOUT)
+    ActiveRecord::Base.connection.disable_query_cache!
 
-    Prison.active.each do |prison|
-      prison.offenders.each do |offender|
+    # Potential memory leak:
+    # prison.offenders seems to be cached somewhere, because subsequent calls to prison.offenders are always faster than the first call
```
starswan, line 10: This is our cache in redis - so IMHO there isn't a memory leak here. There's also a good chance that we hit the Oracle cache on the far end (Oracle caches all queries based on parameters when it can)

ollietreend: Good point @starswan – thanks 👍🏻

I've just confirmed this with:

```ruby
Prison.new('LEI').count # first call is slow
Prison.new('LEI').count # second call is quick
Rails.cache.clear
Prison.new('LEI').count # is slow again after clearing the cache
```

I'll remove this comment.

starswan (approved):

PR #1546, opened 2021-02-04
Convert :allocation tag into !:push_pom_to_delius

starswan:
Currently if a new test is written that creates an allocation, the :allocation rspec tag has to be added to it to prevent an accidental PushPomToDeliusJob being kicked off - due to its implementation in after_save. 

This PR turns that setting on its head, so that now tests have to enable the 'PushPomToDeliusJob' via the :push_pom_to_delius tag - this is much safer and should result in less confusion and noise in the tests

### DIFF
--- a/spec/controllers/api/allocation_api_controller_spec.rb
+++ b/spec/controllers/api/allocation_api_controller_spec.rb
@@ -3,9 +3,6 @@ require 'rails_helper'
 RSpec.describe Api::AllocationApiController, :allocation, type: :controller do
   describe '#show' do
     let(:rsa_private) { OpenSSL::PKey::RSA.generate 2048 }
-    let(:stub_url) { "#{api_host}/secure/offenders/nomsNumber/#{offender.fetch(:offenderNo)}/prisonOffenderManager" }
-    let(:api_host) { Rails.configuration.community_api_host }
-
     let(:prison) { build(:prison) }
     let!(:co_working_allocation) {
       create(:allocation, :co_working, primary_pom_nomis_id: primary_pom.staff_id,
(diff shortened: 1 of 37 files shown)


spec/controllers/api/allocation_api_controller_spec.rb
```diff
@@ -3,9 +3,6 @@
 RSpec.describe Api::AllocationApiController, :allocation, type: :controller do
```
ollietreend, line 3: This file still has the `:allocation` tag applied

spec/rails_helper.rb
```diff
@@ -95,9 +95,12 @@
     WebMock.disable_net_connect!(allow_localhost: true)
   end
 
-  # This is to prevent test from having to mock this callback due
+  # This is to prevent most tests from having to mock this callback due
   # to an after_save call back in Allocation.
-  config.before(:each, :allocation) do
+  #config.before(:each, type: lambda { |_v, m| puts "m=#{m.inspect}"; m[:push_pom_to_delius] != true } ) do
+  #  allow(PushPomToDeliusJob).to receive(:perform_later)
+  #end
```
ollietreend, line 102: Commented out code could do with being removed

spec/rails_helper.rb
```diff
@@ -95,9 +95,12 @@
...
+  config.before(:each, type: lambda { |_v, m| m[:push_pom_to_delius] != true } ) do
```
ollietreend, line 103: Might be helpful to add a comment explaining how to use this new 'opposite' tag. You've explained that it's mocked by default, but how do I disable the mock if I want to be able to test pushing a POM to nDelius?

Should reduce the need to interpret the meaning of the lambda.

ollietreend (approved):

PR #1554, opened 2021-02-10
MO-265: Display cell location on POM caseload view

theKHutDeveloper:
We already call the endpoints `/prisoners/{offenderNo}` and `/locations/{locationId}/inmates
which holds the cell location in the fields `internalLocation` and `assignedLivingUnitDesc` respectively.
Previously we did not store the value but in this commit it is stored as the `cell_location`.

Added the cell location to both the `caseload#index` and the `caseload#new` views, along with the ability to sort by
this field.

Caseload Index page:
Screen Shot 2021-02-09 at 08 47 29

Caseload New page:
Screen Shot 2021-02-09 at 08 51 04


app/models/allocated_offender.rb
```diff
@@ -22,6 +22,10 @@ def latest_movement_date
     @offender.latest_movement&.create_date_time&.to_date
   end
 
+  def location
+    @offender.cell_location
+  end
```
ollietreend, line 27: I'm not sure this is needed, since we already delegate `cell_location` to the offender object.

This just seems to duplicate that – so `AllocatedOffender` has two methods which return the same value:
- `allocated_offender.location`
- `allocated_offender.cell_location`

We should be able to remove the `location` method and just keep the delegated method `cell_location`.

app/views/caseload/index.html.erb
```diff
@@ -61,12 +61,14 @@
         </a>
         <%= sort_arrow('last_name') %>
       </th>
+
       <th class="govuk-table__header" scope="col">
-        <a href="<%= sort_link('latest_movement_date') %>">
+        <a href="<%= sort_link('location') %>">
```
ollietreend, line 66: If the `location` method is removed from `AllocatedOffender`, then this will need to become `cell_location`

```suggestion
        <a href="<%= sort_link('cell_location') %>">
```

app/views/caseload/index.html.erb
```diff
@@ -61,12 +61,14 @@
...
           Location
         </a>
-        <%= sort_arrow('latest_movement_date') %>
+        <%= sort_arrow('location') %>
```
ollietreend, line 69: Same here:

```suggestion
        <%= sort_arrow('cell_location') %>
```

app/views/caseload/new.html.erb
```diff
@@ -16,6 +16,14 @@
         </a>
         <%= sort_arrow('last_name') %>
       </th>
+
+      <th class="govuk-table__header" scope="col">
+        <a href="<%= sort_link('latest_movement_date') %>">
+          Location
+        </a>
+        <%= sort_arrow('latest_movement_date') %>
```
ollietreend, line 24: Looks like these are sorting by the wrong field

```suggestion
        <a href="<%= sort_link('cell_location') %>">
          Location
        </a>
        <%= sort_arrow('cell_location') %>
```

ollietreend (commented):
@theKHutDeveloper just a couple of comments, otherwise this looks great 👍🏻

ollietreend (approved):

app/controllers/caseload_controller.rb
```diff
@@ -22,7 +22,29 @@ def new
 private
 
   def sort_allocations(allocations)
-    sort_collection(allocations, default_sort: :last_name)
+    if params['sort'].present? && params['sort'].downcase.include?('cell_location')
```
starswan, line 25: I think you should be good to use the 'sort_params' method here to pick up the parameters rather than grabbing them by hand

app/views/caseload/new.html.erb
```diff
@@ -46,30 +54,31 @@
...
-      <th class="govuk-table__header" scope="col">Action</th>
     </tr>
     </thead>
     <tbody class="govuk-table__body">
     <% @new_cases.each_with_index do |allocation, i| %>
       <tr class="govuk-table__row offender_row_<%= i %>">
         <td aria-label="Prisoner name" class="govuk-table__cell">
           <span class='prisoner-name'>
-            <%= allocation.full_name %>
+            <%= link_to(highlight(allocation.full_name, @q), prison_prisoner_path(@prison.code, allocation.nomis_offender_id), class: "govuk-link" ) %>
```
starswan, line 64: the 'highlight' call is superflous here IMHO because there is no search facility on the 'new-cases' screen

app/controllers/caseload_controller.rb
```diff
@@ -22,7 +22,36 @@ def new
...
+          1
+        end
+      elsif a.latest_movement_date.nil? && b.latest_movement_date.present?
+        1
+      elsif a.latest_movement_date.present? && b.latest_movement_date.nil?
+        -1
+      else
+        a.latest_movement_date <=> b.latest_movement_date
+      end
+    end
```
ollietreend, line 51: As we said: this method feels a bit confusing if you don't know what it's about.

It might be worth adding a comment to explain (at a high level) what it does – just to help anyone who stumbles across it in the future.

spec/features/pom_caseload_feature_spec.rb
```diff
@@ -129,11 +145,14 @@
       end
     end
 
-    it 'can be sorted by location' do
+    it 'can be sorted by cell location' do
+      # Rotl's offenders are grouped when sorted by cell location. If sorted in ascending order they appear
+      # at the top otherwise they are grouped at the bottom
```
ollietreend, line 150: In fact you could probably even use this comment ^^
It'd just be nice to have it explained in the implementation, too.

ollietreend (approved):
I'm re-approving 🙂
This looks really good. I've a note about putting an explainer comment on the `cell_location_sort` method.

PR #1562, opened 2021-02-12
MO-478 don't show early allocation entry screen for CRC cases

starswan:


app/models/case_information.rb
```diff
@@ -41,6 +41,10 @@ class CaseInformation < ApplicationRecord
 
   before_validation :set_probation_service
 
+  def nps?
+    case_allocation == NPS
+  end
```
ollietreend, line 46: We have an almost identical method on the `OffenderBase` object, too.

`offender.nps_case?`

If you think the CaseInfo record is the right place for it (it probably is..), then it might be worth delegating `OffenderBase.nps_case?` to `CaseInformation.nps_case?` to keep things DRY.

ollietreend (approved):
One suggestion to keep things DRY. But otherwise this looks good 👍🏻

PR #1567, opened 2021-02-16
MO-394 Part 2: Fix a bug in the existing HMP Prescoed rules

ollietreend:
This is the second in a series of PRs relating to MO-394, which is to implement OMIC rules in Open Prisons.

---

Whilst working to implement the open prison rules, I discovered that there's actually a bug in existing HMP Prescoed pilot of open prison rules. This PR fixes that bug.

For offenders included in the pilot (Welsh offenders entering HMP Prescoed after 19/10/2020), their responsibility calculation was incorrect. It had been hard-coded to be COM responsible, POM supporting. However, the responsibility should actually follow the calculated handover dates (as per existing OMIC rules).

### Summary of changes

- I've introduced new tests to comprehensively describe the expected behaviour of Open Prison rules in HMP Prescoed. These have been implemented in a new test format which consciously moves away from the old 'split brain' tests (where handover dates and responsibility were tested independently in different files) and instead tests handover dates alongside responsibility.
- Fix the bug in HMP Prescoed responsibility calculations.
- Relocate the HMP Prescoed rules to make them fit into the rest of the codebase more cleanly, rather than standing alone. Open prison rules are more similar to closed prison rules than they are different – so it made sense to DRY these rules up to reduce confusion and duplicate code.

**Note:** this PR is based upon the Part 1 PR #1566, so the first 3 commits are safe to ignore. The new commits in this branch are:
1. 094848a: Fix HMP Prescoed bug and refactor related tests
2. 7f14302: Refactor existing Prescoed rule implementation

### DIFF
--- a/app/services/handover_date_service.rb
+++ b/app/services/handover_date_service.rb
@@ -30,19 +30,6 @@ class HandoverDateService
                        start_date: nil, handover_date: nil,
                        reason: 'Immigration Case'
 
-    elsif offender.recent_prescoed_case? && offender.indeterminate_sentence? && offender.nps_case?
-      handover_date = prescoed_handover_date(offender)
-      if handover_date.future?
-        pom = RESPONSIBLE
-        com = SUPPORTING
-      else
-        pom = SUPPORTING
-        com = RESPONSIBLE
-      end
-      HandoverData.new custody: pom, community: com,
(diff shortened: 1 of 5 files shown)


ollietreend:
`prescoed_handover_date` is still in the service – needs to be removed

starswan (approved):

PR #1569, opened 2021-02-16
Suspend the old nDelius import cron job

ollietreend:
We think it could be interacting / racing with the new Community API import job and causing problems.

There is another ticket in our backlog to remove this job entirely – along with all associated code. But this is just a 'quick fix' to see if it stops some of the problems we've been seeing.


starswan (approved):
should probably do preprod also...

PR #1570, opened 2021-02-17
MO-463 fix movement events

starswan:
We have recently discovered that our movement code is wrong, due to our mis-understanding of the movement API.

This PR attempts to fix this.

### DIFF
--- a/app/services/hmpps_api/prison_api/movement_api.rb
+++ b/app/services/hmpps_api/prison_api/movement_api.rb
@@ -8,9 +8,12 @@ module HmppsApi
       def self.movements_on_date(date)
         route = '/movements'
 
+        # the 'fromDateTime' field in the API is the 'earliest creation date' of a movement
+        # obviously we have no idea how early this might be, so set it to last year so that they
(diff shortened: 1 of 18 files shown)


app/services/hmpps_api/prison_api/movement_api.rb
```diff
@@ -40,7 +43,7 @@ def self.admissions_for(offender_nos)
                      .group_by { |x| x['offenderNo'] }.values
 
         movements = data.map { |d|
-          d.sort_by { |k| k['createDateTime'] }.map { |movement|
```
ollietreend, line 43: Should we be sorting by `movementDate` if this can be `nil`?

Any reason why we need to change this sort at this point?

ollietreend (approved):

app/services/hmpps_api/prison_api/movement_api.rb
starswan: turned out that movementDate had been broken for 2+ years, fixed on 19-Jan-2021

PR #1571, opened 2021-02-18
Remove Offender Presenter class - it clearly isn't working

starswan:


spec/helpers/early_allocation_helper_spec.rb
```diff
@@ -9,7 +9,7 @@
   let(:offender) { build(:offender, sentence: offender_sentence) }
   let(:nomis_offender_id) { offender.offender_no }
   let(:case_info) { create(:case_information, nomis_offender_id: nomis_offender_id, early_allocations: early_allocations) }
-  let(:offender_presenter) { OffenderPresenter.new(offender) }
+  let(:offender_presenter) { offender }
```
ollietreend, line 12: It might be nice to DRY this spec up.

We already have an `offender` object available – presumably that should be used throughout this spec rather than the duplicate `offender_presenter`.

ollietreend (approved):
Approved – with one potential DRY optimisation

PR #1572, opened 2021-02-18
MO-423 female missing info journey

starswan:
Create the missing info journey for women's prisons.

### DIFF
--- a/app/controllers/female_missing_infos_controller.rb
+++ b/app/controllers/female_missing_infos_controller.rb
@@ -1,0 +1,89 @@
+# frozen_string_literal: true
+
+class FemaleMissingInfosController < PrisonsApplicationController
+  include Wicked::Wizard
+
+  before_action :load_missing_info, except: :new
+
+  # We have to ignore these 2 keys when loading attributes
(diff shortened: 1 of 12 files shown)


app/controllers/female_missing_infos_controller.rb
```diff
@@ -0,0 +1,87 @@
...
+  end
+
+  def update
+    @missing_info.assign_attributes(step_params)
+    save_session
+
+    if @missing_info.valid?
+      if step == :complexity_level
+        ComplexityMicroService.save @missing_info.nomis_offender_id, @missing_info.complexity_level
+        if CaseInformation.find_by nomis_offender_id: @missing_info.nomis_offender_id
```
ollietreend, line 38: It might be nice to pull this out into its own method:

```ruby
CaseInformation.find_by nomis_offender_id: @missing_info.nomis_offender_id
```

I can see it's been used (at least) a couple of times. Plus I've seen an equivalent call to find out if there's a Complexity level present.

How about refactoring these into private methods?

- `has_complexity_level?`
- `has_case_information?`

ollietreend (approved):

PR #1575, opened 2021-02-22
MO-366 add the PRD to the handover calculation for indeterminate prisoners

starswan:


app/services/handover_date_service.rb
```diff
@@ -53,56 +51,33 @@ def self.handover(raw_offender)
...
       end
     else
       # CRC case
-      responsibility = crc_responsibility(offender)
-      if responsibility == NOT_INVOLVED
-        HandoverData.new custody: NOT_INVOLVED, community: NOT_INVOLVED,
-                         start_date: nil, handover_date: nil,
-                         reason: 'CRC Case - missing dates'
+      crc_date = offender.release_date - 12.weeks
+      if crc_date > Time.zone.today
```
ollietreend, line 73: This could read more nicely as:

```suggestion
      if crc_date.future?
```

ollietreend: Also: since the only thing which changes between these conditional statements is the responsibility, something like this might be a more DRY way to achieve the same thing.

IMHO it seems more intention-revealing and easier to read – but that's just my opinion!

```ruby
if crc_date.future?
  pom = RESPONSIBLE
  com = NOT_INVOLVED
else
  pom = SUPPORTING
  com = RESPONSIBLE
end

HandoverData.new custody: pom, community: com,
                 start_date: crc_date, handover_date: crc_date,
                 reason: 'CRC Case'
```

app/services/handover_date_service.rb
```diff
@@ -307,12 +249,28 @@ def expected_time_in_custody_gt_10_months?
...
     end
 
+    # We can not calculate the handover date for NPS Indeterminate
+    # with parole cases where the TED is in the past as we need
+    # the parole board decision which currently is not available to us.
     def release_date
       if @offender.indeterminate_sentence?
-        @offender.parole_eligibility_date || @offender.tariff_date
-      else
+        if @offender.tariff_date.present? && @offender.tariff_date > Time.zone.today
```
ollietreend, line 249: Ditto here – this may just be a personal preference of mine, so feel free to ignore:

```suggestion
        if @offender.tariff_date.present? && @offender.tariff_date.future?
```

app/services/handover_date_service.rb
```diff
@@ -307,12 +249,28 @@ def expected_time_in_custody_gt_10_months?
...
+          @offender.tariff_date
+        else
+          [
+            @offender.parole_review_date,
+            @offender.parole_eligibility_date
+          ].compact.select { |date| date >= Time.zone.today }.min
```
ollietreend, line 255:
```suggestion
          ].compact.reject(&:past?).min
```

ollietreend (approved):
A couple of comments on code style – please feel free to ignore, as these are opinions.

This is great stuff! 😁🎉

PR #1583, opened 2021-03-01
MO-431 spinoff - controller sorting concern

starswan:
This is a spinoff from MO-431 - whilst re-writing the controller, we discovered a sorting concern hiding inside the old SummaryService / Bucket classes. This PR brings these 2 methods into a mixin for any controller that needs sorting behaviour - and uses it in CaseloadController.

### DIFF
--- a/app/controllers/caseload_controller.rb
+++ b/app/controllers/caseload_controller.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class CaseloadController < PrisonStaffApplicationController
+  include Sorting
+
   def index
     allocations = @pom.allocations
 
@@ -20,18 +22,7 @@ class CaseloadController < PrisonStaffApplicationController
 private
 
   def sort_allocations(allocations)
-    if params['sort'].present?
-      sort_field, sort_direction = params['sort'].split.map(&:to_sym)
-    else
-      sort_field = :last_name
-      sort_direction = :asc
-    end
-
-    # cope with nil values by sorting using to_s - only dates and strings in these fields
-    allocations = allocations.sort_by { |sentence| sentence.public_send(sort_field).to_s }
-    allocations.reverse! if sort_direction == :desc
-
-    allocations
+    sort_collection(allocations, default_sort: :last_name)
   end
 
   def filter_allocations(allocations)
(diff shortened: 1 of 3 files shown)


app/controllers/caseload_controller.rb
```diff
@@ -20,18 +22,7 @@ def new
...
-      sort_field = :last_name
-      sort_direction = :asc
-    end
-
-    # cope with nil values by sorting using to_s - only dates and strings in these fields
-    allocations = allocations.sort_by { |sentence| sentence.public_send(sort_field).to_s }
-    allocations.reverse! if sort_direction == :desc
-
-    allocations
+    sort_collection(allocations, :last_name)
```
ollietreend, line 25: It would be nice for `default_sort` and `default_direction` to be keyword parameters, to make it more obvious what their meaning is.

Otherwise this line of code reads like we're always sorting by `:last_name`. The fact that we actually read the request `params`, and `:last_name` is just a fallback, is hidden within the implementation of `Sorting` concern.

```suggestion
    sort_collection(allocations, default_sort: :last_name)
```

(where `default_direction` is still optional and `:asc` if not given)

app/controllers/concerns/sorting.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+module Sorting
+  def sort_collection(items, default_sort, default_direction = :asc)
```
ollietreend, line 4: Keyword arguments for 'default' parameters:

```suggestion
  def sort_collection(items, default_sort:, default_direction: :asc)
```

app/controllers/concerns/sorting.rb
```diff
@@ -0,0 +1,38 @@
...
+    field, direction = sort_params(default_sort, default_direction)
+    sort_with_public_send items, field, direction
+  end
+
+  def sort_params(default_sort_params, default_direction = :asc)
+    if params['sort']
+      params['sort'].split.map { |s| s.downcase.to_sym }
+    else
+      [default_sort_params, default_direction]
```
ollietreend, line 13: ### Variable name consistency

`default_sort_params` should probably just be `default_sort`. There's no need for the `_params` suffix.

```suggestion
  def sort_params(default_sort, default_direction = :asc)
    if params['sort']
      params['sort'].split.map { |s| s.downcase.to_sym }
    else
      [default_sort, default_direction]
```

ollietreend (approved):

PR #1584, opened 2021-03-01
MO-432 complexity badge and show caseload

starswan:
Added complexity badges and female caseload view.

Haven't added the link for the 'change' journey, as that makes more sense in the ticket it applies to MO-423


app/models/hmpps_api/offender_base.rb
```diff
@@ -25,6 +25,10 @@ def convicted?
       @convicted_status == 'Convicted'
     end
 
+    def complexity_level
+      @complexity_level ||= ComplexityMicroService.get_complexity(offender_no)
+    end
```
ollietreend, line 30: This implementation is sub-optimal because it'll result in N+1 calls to the Complexity API when operating on a collection of offenders (e.g. on list pages, sorting, etc.)

I expect we'll pick this up in MO-421 though – where we'll replace our stub with a proper API client implementation. So no changes needed for this PR.

ollietreend (approved):

PR #1586, opened 2021-03-04
MO-417 improve token handling

starswan:
Improve our token handling a little bit - better naming of methods and better comment on behaviour


spec/services/hmpps_api/oauth/token_spec.rb
```diff
@@ -5,20 +5,20 @@
     access_token = generate_jwt_token
     token = described_class.new(access_token: access_token, expires_in: 4.hours)
 
-    expect(token).not_to be_expired
+    expect(token).not_to be_needs_refresh
```
ollietreend, line 8: Would be nice to make this more eloquent and follow the format of the other tests:

```suggestion
    expect(token.needs_refresh?).to be(false)
```

ollietreend (approved):

PR #1589, opened 2021-03-09
MO-431 Missing info | Updates to the missing information table

Katy600:
**PR to implement the new Women's  'Add missing details' page**

Screenshot 2021-03-09 at 13 01 51

In order to update the design for the missing information table this ticket required a separate female controller to be created. 

It then created a missing_information view. However it shares a lot of the same designs as the mens summary views, so it reuses the _summary_subnav.html.erb partial from the mens summary view. This displays the table with the various sub tabs. Because the females controller displays the member variables within the controller actions differently to the mens summary controller, it meant that this partial had to be refactored to be able to take both. On retrospect I should have done this as a separate PR. This refactoring of the summary_sub nav, resulted in a further refactoring of the summary controller, so that it ended up with the same variable members as the female controller. It also renders the rest of the views directly in the controller as currently they share the same design as the mens. The mens summary controller also refers to 'pending', as opposed to 'missing_information', so this had to be refactored within the routing as well. 

One of the challenges in this ticket was that I needed my offender to know if it had been allocated. Unfortunately the initial model set up meant that a model association between offender and allocation had not been created. This resulted in the summary controller managing this by creating a getter setter method. This tripped me up during my tests, as I was unaware of having to set this allocation_date method. To solve this in a less convoluted way, OffenderWithPomName wrapper was created which allows you to brings allocation and offender together in one space.


kubectl.sha256
```diff
@@ -0,0 +1 @@
+37f593731b8c9913bf2a3bfa36dacb3058dc176c7aeae2930c783822ea03a573
```
ollietreend, line 1: Looks like this file was committed accidentally – it should be safe to delete.

spec/features/summary_feature_spec.rb
```diff
@@ -0,0 +1,76 @@
+require 'rails_helper'
+
+feature 'sub tabs' do
```
ollietreend, line 3: I think this feature spec could benefit from having a more descriptive name.

If this tests the tab navigation on the male prisoners summary page, then something along the lines of this might be more descriptive:

```suggestion
feature 'male prisoners summary navigation tabs' do
```

ollietreend (commented):
This looks excellent @Katy600 🎉 great work!

I've added a few comments – but nothing big. I'll approve once those few things have been addressed 🙂

ollietreend (approved):

app/controllers/female_prisoners_controller.rb
```diff
@@ -1,2 +1,77 @@
...
+
+  def allocated
+    all_offenders = @prison.offenders
+    active_allocations_hash = AllocationService.active_allocations(all_offenders.map(&:offender_no), @prison.code)
+    allocated_offenders = load_offenders :allocated, all_offenders, active_allocations_hash
+
+    offenders = allocated_offenders.map { |offender| OffenderWithPomName.new(offender, active_allocations_hash.fetch(offender.offender_no)) }
+    @offenders = Kaminari.paginate_array(offenders).page(page)
+    render 'summary/allocated'
+  end
```
ollietreend, line 14:
```suggestion
  def allocated
    every_offender = @prison.offenders
    active_allocations_hash = AllocationService.active_allocations(every_offender.map(&:offender_no), @prison.code)
    allocated_offenders = load_offenders :allocated, every_offender, active_allocations_hash

    offenders = allocated_offenders.map { |offender| OffenderWithPomName.new(offender, active_allocations_hash.fetch(offender.offender_no)) }
    @offenders = Kaminari.paginate_array(offenders).page(page)
    render 'summary/allocated'
  end
```

PR #1596, opened 2021-03-10
MO-265 last offender movement refactor

starswan:
Create a latest_temp_movement property accessible from OffenderBase rather than just in OffenderSummary. Create at initialisation rather than via attr_accessor as well

### DIFF
--- a/app/models/allocated_offender.rb
+++ b/app/models/allocated_offender.rb
@@ -18,8 +18,8 @@ class AllocatedOffender
     @offender = offender
   end
 
-  def latest_movement_date
-    @offender.latest_movement&.movement_date
+  def latest_temp_movement_date
+    @offender.latest_temp_movement&.movement_date
   end
 
   def pom_responsibility
--- a/app/models/hmpps_api/offender_base.rb
+++ b/app/models/hmpps_api/offender_base.rb
@@ -17,7 +17,7 @@ module HmppsApi
 
     attr_accessor :category_code, :date_of_birth, :prison_arrival_date, :sentence, :allocated_pom_name
 
-    attr_reader :first_name, :last_name, :booking_id, :offender_no, :sentence_type
+    attr_reader :first_name, :last_name, :booking_id, :offender_no, :sentence_type, :latest_temp_movement
 
     # This is needed (sadly) because although when querying by prison these are filtered out,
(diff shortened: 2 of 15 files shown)


app/models/allocated_offender.rb
```diff
@@ -19,7 +19,7 @@ def initialize(staff_id, allocation, offender)
   end
 
   def latest_movement_date
```
ollietreend, line 21: Could this also be renamed to include temp?

```suggestion
  def latest_temp_movement_date
```

spec/support/helpers/api_helper.rb
```diff
@@ -6,6 +6,7 @@ module ApiHelper
   T3_SEARCH = "#{Rails.configuration.prisoner_search_host}/prisoner-search"
   KEYWORKER_API_HOST = ENV.fetch('KEYWORKER_API_HOST')
   COMMUNITY_HOST = "#{Rails.configuration.community_api_host}/secure"
+  ELITE2LATESTMOVE = "#{T3}/movements/offenders?latestOnly=true&movementTypes=TAP"
```
ollietreend, line 9: It would be nice remove the 'Elite2' reference from this variable name.

I thought we'd done a pretty good job of replacing all Elite2 references with Prison API – to avoid confusion to future devs who don't know that API's history. Looks like we missed a this in the API helper.

app/models/allocated_offender.rb
ollietreend: In fact – presumably the next PR will be to relocate this method (into OffenderBase) or to use AllocatedOffender more widely. So presumably this isn't going to be hanging around for long..?

app/models/hmpps_api/offender_base.rb
```diff
@@ -17,7 +17,7 @@ class OffenderBase
 
     attr_accessor :category_code, :date_of_birth, :prison_arrival_date, :sentence, :allocated_pom_name
 
-    attr_reader :first_name, :last_name, :booking_id, :offender_no, :sentence_type
+    attr_reader :first_name, :last_name, :booking_id, :offender_no, :sentence_type, :latest_temp_movement
```
ollietreend, line 20: ...which this attr reader won't be needed for long, either..?

ollietreend (approved):

PR #1601, opened 2021-03-15
Summary feature failure 15 mar 21

starswan:
Remove Monday specific bahaviour and tests - the NART report is long-gone


ollietreend (approved):
@starswan I've added one idea to avoid freezing time entirely – but it's just an opinion, so I'll leave it up to you 🙂

Also: 👍🏻 to reducing the magic `enrich_offenders` method.

PR #1602, opened 2021-03-16
Support ticket:  17 identical automated emails sent

Katy600:
This PR fixes a support request. 

_I monitor the northants-omic@justice.gov.uk mailbox. This weekend we have had a glitch with 17 identical automated emails sent from Saturday to Monday regarding one offender (E229139 Kenneth Smith A2361ED). Please could this glitch be investigated so that it does not reoccur?_ 

This was a result of Allocation not being deleted when a prisoner no longer has an allocation. When it tries to deallocate an an allocated offender it crashes when the allocation is not active.

### DIFF
--- a/app/models/allocation.rb
+++ b/app/models/allocation.rb
@@ -127,12 +127,12 @@ class Allocation < ApplicationRecord
             :event_trigger,
             :prison, presence: true
 
-  def offender_released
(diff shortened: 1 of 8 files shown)


ollietreend (approved):
Great work @Katy600 🎉

PR #1605, opened 2021-03-17
Fix a UI bug where the wrong tab was highlighted

ollietreend:
The 'See allocations' tab was highlighted as the active tab, when actually on the 'Make allocations' page.

Also: fix an invalid HTML attribute


starswan (approved):
would have been nice to have written a test? I wonder if this is broken in Women's as well?

PR #1606, opened 2021-03-17
MO-429 Women's OMIC rules

starswan:
Women's OMIC rules

### DIFF
--- a/spec/services/handover_date_service_spec.rb
+++ b/spec/services/handover_date_service_spec.rb
@@ -5,6 +5,7 @@ require 'rails_helper'
 describe HandoverDateService do
   subject { described_class.handover(offender) }
 
+  let(:test_strategy) { Flipflop::FeatureSet.current.test! }
   let(:pom) { subject.custody }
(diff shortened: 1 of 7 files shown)


spec/services/handover_date_service_spec.rb
```diff
@@ -448,6 +495,50 @@
...
+
+        context 'when entering on or after the policy start date' do
+          let(:sentence_start_date) { womens_policy_start_date }
+
+          it 'follows POM policy rules' do
+            expect(pom).to be_responsible
+          end
+
+          it 'follows COM policy rules' do
+            expect(com).not_to be_supporting
```
ollietreend, line 529: This might be a 'stronger' test:

```suggestion
            expect(com).not_to be_involved
```

i.e. asserting that the COM isn't involved whatsoever (rather than just 'not supporting')

I know that POM & COM can't both be responsible at the same time, but I don't think that's explicitly enforced unless we test for it.

ollietreend (approved):
Looks good 👍🏻

PR #1607, opened 2021-03-17
MO-421 complexity client

starswan:
This PR adds the last piece of the 'complexity level' jigsaw - the ability to talk to the complexity micro-service via an HTTP client


app/services/hmpps_api/prison_api/offender_api.rb
```diff
@@ -29,32 +29,35 @@ def self.list(prison, page = 0, page_size: 20)
...
-          temp_movements = HmppsApi::PrisonApi::MovementApi.latest_temp_movement_for([url_offender_no])
-          HmppsApi::Offender.from_json(response.first,
-                                       search_data.fetch(url_offender_no, {}),
-                                       latest_temp_movement: temp_movements[url_offender_no]).tap do |offender|
+          search_data = get_search_payload([offender_no])
+          temp_movements = HmppsApi::PrisonApi::MovementApi.latest_temp_movement_for([offender_no])
+          HmppsApi::Offender.new(response.first,
+                                 search_data.fetch(offender_no, {}),
+                                 latest_temp_movement: temp_movements[offender_no],
+                                 complexity_level: ComplexityMicroService.get_complexity(offender_no)).tap do |offender|
```
ollietreend, line 60: Just pointing this out, but I don't have a perfect solution to it:

Taking this approach means we'll end up performing needless calls to the Complexity API for **most offenders in the service** – since male offenders don't have a Complexity of Need level. It seems a bit wasteful.

Let's not forget the fact there will be a (possibly very small) performance hit for performing these calls – it's far cheaper to simply not make the call at all when we know it's not needed.

Should this class have knowledge of which conditions must be met for a Complexity of Need level to be needed? It feels wrong to do that – but it'd be great to avoid these extra API calls if at all possible.

spec/fixtures/vcr_cassettes/complexity/get_complexities.yml
```diff
@@ -0,0 +1,107 @@
+---
+http_interactions:
+- request:
+    method: post
+    uri: https://hmpps-complexity-of-need-staging.apps.live-1.cloud-platform.service.justice.gov.uk/v1/complexity-of-need/multiple/offender-no
```
ollietreend, line 5: The API hostname change presumably means that these VCR cassettes will change too.

spec/support/helpers/api_helper.rb
```diff
@@ -50,6 +50,11 @@ def stub_offender(offender)
     stub_request(:post, T3_LATEST_MOVE_URL).with(
       body: [offender_no].to_json).
       to_return(body: [].to_json)
+
+    allow(ComplexityMicroService).
+      to receive(:get_complexity).
+        with(offender_no).
+        and_return(offender.fetch(:complexityLevel))
```
ollietreend, line 57: Is there any reason for using this RSpec mock over a WebMock mock?

Doesn't necessarily need to be changed – just interested in the reason.

spec/support/helpers/api_helper.rb
```diff
@@ -146,11 +152,15 @@ def stub_offenders_for_prison(prison, offenders, movements = [])
...
 
     stub_request(:post, T3_LATEST_MOVE_URL).with(
-      body: offenders.map { |offender| offender.fetch(:offenderNo) }.to_json).
+      body: offender_nos.to_json).
       to_return(body: movements.to_json)
 
     stub_movements
 
+    allow(ComplexityMicroService).to receive(:get_complexities).with(offender_nos).and_return(
+      offenders.map { |offender| [offender.fetch(:offenderNo), offender.fetch(:complexityLevel)] }.to_h
```
ollietreend, line 161: Ditto

app/services/complexity_micro_service.rb
```diff
@@ -1,15 +1,47 @@
...
+    def get_complexities offender_nos
+      route = '/complexity-of-need/multiple/offender-no'
+      result = client.post route, [offender_nos], cache: false
+      result.map { |complexity| [complexity.fetch('offenderNo'), complexity.fetch('level')] }.to_h
+    end
+
+    def get_history offender_no
+      route = "/complexity-of-need/offender-no/#{offender_no}/history"
+      results = client.get route, cache: false
+      results.sort_by { |item| item.fetch('createdTimeStamp') }.map do |result|
```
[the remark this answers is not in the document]
starswan: The problem is that I wanted the list in forward order, not reversed. The reversing is done very late (deliberately) in the allocation history code - I think it wants forward ordering

app/services/hmpps_api/prison_api/offender_api.rb
starswan: I think we're stuffed here - as we don't really know which prison we're in. I suppose we could call PrisonService.womens_prison? on the prison_id returned...

spec/support/helpers/api_helper.rb
starswan: It seemed easier (c.f. Community API where we do the same thing)

.circleci/config.yml
```diff
@@ -21,6 +21,7 @@ references:
           RACK_ENV: test
           VCR: 1
           PARALLEL_TEST_PROCESSORS: 7
+          COMPLEXITY_API_HOST: https://hmpps-complexity-of-need-staging.apps.live-1.cloud-platform.service.justice.gov.uk
```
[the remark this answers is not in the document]
ollietreend: @starswan this still needs changing 🙂

app/services/hmpps_api/complexity_api.rb
```diff
@@ -0,0 +1,49 @@
...
+      end
+
+      def get_history offender_no
+        route = "/complexity-of-need/offender-no/#{offender_no}/history"
+        results = client.get route, cache: false
+        results.sort_by { |item| item.fetch('createdTimeStamp') }.map do |result|
+          {
+          level: result.fetch('level'),
+          createdTimeStamp: DateTime.parse(result.fetch('createdTimeStamp')),
+          sourceUser: result.fetch('sourceUser')
```
ollietreend, line 34: `sourceUser` is an optional field – it will be omitted from the response if one wasn't given for that record.

To avoid `KeyError (key not found: "sourceUser")` we should default to `nil`:

```suggestion
          sourceUser: result['sourceUser']
```

(or follow the same pattern as the `notes` field...)

app/services/hmpps_api/prison_api/offender_api.rb
```diff
@@ -29,32 +29,43 @@ def self.list(prison, page = 0, page_size: 20)
...
+                       HmppsApi::Offender.new api_response,
+                                              search_data.fetch(offender_no, {}),
+                                              latest_temp_movement: temp_movements[offender_no],
+                                              complexity_level: HmppsApi::ComplexityApi.get_complexity(offender_no)
+                     else
+                       HmppsApi::Offender.new api_response,
+                                              search_data.fetch(offender_no, {}),
+                                              latest_temp_movement: temp_movements[offender_no],
+                                              complexity_level: nil
+                     end
```
ollietreend, line 67: We could DRY this up by only making `complexity_level` conditional – since that's the only thing that changes between the two branches of `prisoner`.

IMHO this is a little easier to read/follow because it's obvious what changes and what stays the same.

```suggestion
          complexity_level = if PrisonService::womens_prison?(api_response.fetch('latestLocationId'))
                               HmppsApi::ComplexityApi.get_complexity(offender_no)
                             end
          prisoner = HmppsApi::Offender.new api_response,
                                            search_data.fetch(offender_no, {}),
                                            latest_temp_movement: temp_movements[offender_no],
                                            complexity_level: complexity_level
```

app/services/hmpps_api/prison_api/offender_api.rb
```diff
@@ -29,32 +29,43 @@ def self.list(prison, page = 0, page_size: 20)
         booking_ids = data.map { |o| o.fetch('bookingId') }
         sentence_details = HmppsApi::PrisonApi::OffenderApi.get_bulk_sentence_details(booking_ids)
 
+        complexities = HmppsApi::ComplexityApi.get_complexities offender_nos
```
ollietreend, line 32: Any chance we could apply the same `womens_prison?` logic here, to avoid calling out when listing offenders in a male prison?

I suppose the call to `PrisonService::womens_prison?(prison_code)` could even be split out into a private helper method within this class – e.g. `get_complexity?` – which would better explain what the intention is of this conditional logic.

spec/services/hmpps_api/complexity_api_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+  describe '#get_history', vcr: { cassette_name: 'complexity/get_history' } do
+    scenario 'happy path' do
+      expect(described_class.get_history('S0004FT')).to eq(
+        [
+          { createdTimeStamp: DateTime.parse('2021-03-18T14:33:28.364Z'), level: 'low', sourceUser: 'SDICKS_GEN', notes: 'Happy Feet' },
+          { createdTimeStamp: DateTime.parse('2021-03-18T14:34:58.551Z'), level: 'high', sourceUser: 'SDICKS_GEN', notes: 'Happy Feet' },
+          { createdTimeStamp: DateTime.parse('2021-03-18T14:35:20.046Z'), level: 'high', sourceUser: 'SDICKS_GEN', notes: 'Happy Feet' },
+        ])
+    end
+  end
```
ollietreend, line 46: Might be worth testing that it can handle responses which don't include the optional fields:

- `sourceUser`
- `notes`

app/services/hmpps_api/prison_api/offender_api.rb
ollietreend:
```suggestion
        complexities = if PrisonService::womens_prison?(prison)
                         HmppsApi::ComplexityApi.get_complexities(offender_nos)
                       else
                         {}
                       end
```

ollietreend (approved):

PR #1608, opened 2021-03-18
MO-421 move nomis vcr location

starswan:
We need to capture a few VCR interactions between us and the Complexity service.

This PR moves the existing VCR cassettes into a new directory (which is then added to .gitignore) so that there is space to put new VCR cassettes into a different directory which can be checked into source control

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,10 +28,11 @@
 
 # Ignore IDEs
 .idea/
+*.iml
 .vscode/
 
-# Ignore VCR cassettes
-/spec/fixtures/vcr_cassettes
+# Ignore Prison API / NOMIS VCR cassettes
+/spec/fixtures/vcr_cassettes/prison_api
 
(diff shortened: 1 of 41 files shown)


.gitignore
```diff
@@ -30,12 +30,13 @@
 .idea/
 .vscode/
 
-# Ignore VCR cassettes
-/spec/fixtures/vcr_cassettes
+# Ignore Prison API / NOMIS VCR cassettes
+/spec/fixtures/vcr_cassettes/nomis
```
ollietreend, line 34: Is `nomis` the most meaningful name for this directory?

I'm assuming lots of these VCRs also pull in data from the Community API, Keyworker API, etc.

Naming is hard, and I'm struggling to come up with a suggestion that makes sense. Maybe `local` or `untracked`? 🤔 I'm not sure.

ollietreend (approved):

PR #1614, opened 2021-03-23
MO-425 pre-factor - bind Allocation to the CaseInformation record

starswan:
This PR starts to create a data model for the service by attaching the Allocation model to the CaseInformation model via a Rails has_one relationship. It is perfectly possible to destroy Allocation records - they show up in the AllocationHistory just fine - although because of the metadata currently recorded on a release, we ignore the PaperTrail::Version record associated with a destroy in favour of our own change record.

### DIFF
--- a/app/models/allocation.rb
+++ b/app/models/allocation.rb
@@ -89,7 +89,9 @@ class Allocation < ApplicationRecord
   # Gets the versions in *forward* order - so often we want to reverse
   # this list as we're interested in recent rather than ancient history
   def history
-    get_old_versions.append(self).zip(papertrail_versions).map do |alloc, raw_version|
+    # ignore 1st entry, as initial create will reify to nil
+    # we have to keep the :destroy event and filter the next 'create' as it contains the last (often release) event information
+    papertrail_versions[1..].map(&:reify).append(self).zip(papertrail_versions).reject { |a, _v| a.nil? }.map do |alloc, raw_version|
       AllocationHistory.new(alloc, raw_version)
     end
   end
@@ -187,6 +189,6 @@ private
   end
 
   def papertrail_versions
-    PaperTrail::Version.where(item_type: 'Allocation', nomis_offender_id: nomis_offender_id).where.not(event: 'destroy').order(:created_at)
+    PaperTrail::Version.where(item_type: 'Allocation', nomis_offender_id: nomis_offender_id).order(:created_at)
   end
 end
--- a/app/models/allocation.rb
+++ b/app/models/allocation.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class Allocation < ApplicationRecord
-  has_paper_trail
+  has_paper_trail meta: { nomis_offender_id: :nomis_offender_id }
 
   ALLOCATE_PRIMARY_POM = 0
   REALLOCATE_PRIMARY_POM = 1
@@ -37,6 +37,8 @@ class Allocation < ApplicationRecord
     offender_transferred: OFFENDER_TRANSFERRED,
     offender_released: OFFENDER_RELEASED
   }
+
+  belongs_to :case_information, primary_key: :nomis_offender_id, foreign_key: :nomis_offender_id, inverse_of: :allocation
 
   scope :allocations, lambda { |nomis_offender_ids|
     where(nomis_offender_id: nomis_offender_ids)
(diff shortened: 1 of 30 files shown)


spec/factories/allocations.rb
```diff
@@ -24,9 +28,9 @@
       Faker::Number.number(digits: 7)
     end
 
-    nomis_offender_id do
-      Faker::Alphanumeric.alpha(number: 10)
-    end
+    # nomis_offender_id do
+    #   Faker::Alphanumeric.alpha(number: 10)
+    # end
```
Katy600, line 34: This is the only feed back I have so far! Did you mean to leave this commented out?

app/models/allocation.rb
```diff
@@ -135,6 +137,21 @@ def dealloate_offender_after_transfer
...
+  # (item[1] in the array) is our staff_id
+  def new_case_for(staff_id)
+    papertrail_versions.where('created_at >= ?', 7.days.ago).map { |c|
+      YAML.load(c.object_changes)
+    }.select { |c|
+      c.key?('primary_pom_nomis_id') && c['primary_pom_nomis_id'][1] == staff_id ||
+        c.key?('secondary_pom_nomis_id') && c['secondary_pom_nomis_id'][1] == staff_id
+    }.any?
+  end
+
```
Katy600, line 154: Just for my own understanding, what is the benefit of moving this method out of allocated_offender model?

Katy600 (commented):
Very interesting. Thank you Stephen. It seems really exciting to finally have a model relationship between Allocation and Case information! I have just one comment and one question!  I am just hoping I grasped the basics. So probably good to get this reviewed by someone else as well.  :)

app/models/allocation.rb
starswan: Becuase it was only dependent on the 'allocation' object. This class is trying (along with several others) to work around our data model deficiencies, so it should probably not exist. This method is useful, so moving it into allocation is  one step towards removing the class entirely

spec/factories/case_information.rb
```diff
@@ -2,6 +2,8 @@
 
 FactoryBot.define do
   factory :case_information do
+    # association :allocation
+
```
ollietreend, line 6: Unneeded comment?

spec/services/movement_service_spec.rb
```diff
@@ -132,8 +134,9 @@
         updated_allocation = Allocation.find_by(nomis_offender_id: valid_release.offender_no)
 
         expect(CaseInformationService.get_case_information([valid_release.offender_no])).to be_empty
-        expect(updated_allocation.event_trigger).to eq 'offender_released'
-        expect(updated_allocation.prison).to eq 'LEI'
+        expect(updated_allocation).to be_nil
+        # expect(updated_allocation.event_trigger).to eq 'offender_released'
+        # expect(updated_allocation.prison).to eq 'LEI'
```
ollietreend, line 139: Not needed?

spec/services/movement_service_spec.rb
```diff
@@ -200,7 +203,7 @@
             processed = described_class.process_movement(immigration_movement)
 
             expect(CaseInformationService.get_case_information([immigration_movement.offender_no])).to be_empty
-            expect(allocation.event_trigger).to eq 'offender_released'
+            # expect(allocation.event_trigger).to eq 'offender_released'
```
ollietreend, line 206: Not needed?

spec/services/movement_service_spec.rb
```diff
@@ -216,7 +219,7 @@
...
```
ollietreend, line 222: Not needed?

spec/services/movement_service_spec.rb
```diff
@@ -247,7 +250,7 @@
...
```
ollietreend, line 253: Not needed?

spec/services/movement_service_spec.rb
```diff
@@ -267,7 +270,7 @@
...
```
ollietreend, line 273: Not needed?

spec/services/movement_service_spec.rb
```diff
@@ -283,7 +286,7 @@
...
```
ollietreend, line 289: Not needed?

ollietreend (commented):
I've highlighted a few redundant comments to clean up.

But my main concern is: won't we need to migrate existing Allocation records to include the new `nomis_offender_id` metadata field? Surely the `Allocation#papertrail_versions` method won't be able to find existing history for that offender if the metadata isn't backfilled.

PR #1629, opened 2021-03-29
MO-434 Female prison Staff management view

theKHutDeveloper:
In this commit the staff management view has been split dependant on whether in a male or
women's prison. The female offenders have a complexity level, which is not present for
male offenders. So the view for the staff at women's prisons have a completely different
design which shows the complexity level.

There is a future ticket that will show the new staff page designs on the male estate
(omitting the complexity level), which will likely rejoin the views for the male and
female estates.

Screenshot 2021-03-29 at 09 48 36

Screenshot 2021-03-29 at 09 48 54

Screenshot 2021-03-29 at 09 49 06


app/controllers/poms_controller.rb
```diff
@@ -10,8 +10,16 @@ def index
     active_poms, inactive_poms = poms.partition { |pom|
       %w[active unavailable].include? pom.status
     }
-    @active_poms = active_poms.map { |p| PomPresenter.new(p) }
-    @inactive_poms = inactive_poms.map { |p| PomPresenter.new(p) }
+
+    if @prison.womens?
+      # poms are split into groups - 1) probation 2) prison & are sorted by last name
+      @active_probation_poms = active_poms.sort_by(&:last_name).select { |pom| pom.position == 'PO' }.map { |po| PomPresenter.new(po) }
```
starswan, line 16: There are methods called prison_officer? and probation_officer? which already do this - no need to surface position

app/controllers/poms_controller.rb
```diff
@@ -10,8 +10,16 @@ def index
     active_poms, inactive_poms = poms.partition { |pom|
       %w[active unavailable].include? pom.status
     }
-    @active_poms = active_poms.map { |p| PomPresenter.new(p) }
-    @inactive_poms = inactive_poms.map { |p| PomPresenter.new(p) }
+
```
starswan, line 13: We have been creating new controllers using routing rather than splitting behaviours like this.

app/controllers/poms_controller.rb
```diff
@@ -10,8 +10,16 @@ def index
...
+    if @prison.womens?
+      # poms are split into groups - 1) probation 2) prison & are sorted by last name
```
starswan, line 15: We have been creating new controllers using routing rather than splitting behaviours like this.

app/presenters/pom_presenter.rb
```diff
@@ -1,39 +1,53 @@
...
+    @pom_allocated_offenders.count
+  end
+
+  def high_complexity
+    @pom_allocated_offenders.count { |o| o.complexity_level == 'high' }
+  end
+
+  def state
+    @pom.status == 'active' ? 'available' : @pom.status
+  end
```
starswan, line 42: This method looks a bit strange - if we really want to do this sort of display behaviour, it belongs in a view helper. But seems odd that we would only alter the 'active' state.

spec/features/pom_staff_female_feature_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+
+feature "get poms list" do
+  before do
+    signin_spo_user
+
+    # mock to ensure we are in a women's prison
+    allow(PrisonService).to receive(:womens_prison?).and_return(true)
+    visit prison_poms_path('LEI')
+  end
+
```
starswan, line 11: This feels wrong to me - ideally we should not be using VCR for new tests. Our pattern is to use WebMock stubs and build(:womens_prison) to make a Women's prison.

spec/features/pom_staff_female_feature_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+        first('td.govuk-table__cell > a').click
+      end
+
+      expect(page).to have_css(".govuk-button", count: 1)
+      expect(page).to have_content('POM level')
+      expect(page).to have_content('Working pattern')
+      expect(page).to have_content('Status')
+    end
+  end
+
```
starswan, line 32: This feels much more like a view test than a feature test, and the assertions feel a bit weak

app/controllers/poms_controller.rb
starswan, line 16: Rather than do this, our pattern so far has been to create a brand-new controller using routing as the switch. It's really confusing having a switch on women's in both controller and view IMHO

app/presenters/pom_presenter.rb
```diff
@@ -1,39 +1,53 @@
...
 
   def total_cases
-    [tier_a, tier_b, tier_c, tier_d, no_tier].sum
+    @pom_allocated_offenders.count
+  end
+
+  def high_complexity
+    @pom_allocated_offenders.count { |o| o.complexity_level == 'high' }
+  end
+
```
starswan, line 39: IMHO this method might be better if it was called high_complexity_case_count or something similar . It doesn't read very well in the view

spec/features/pom_staff_female_feature_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+  context 'with inactive poms', vcr: { cassette_name: 'prison_api/pom_staff_female_inactive_list' } do
+    it 'shows the page' do
+      click_on('Inactive staff')
+
+      expect(page).to have_content('POM')
+      expect(page).to have_content('POM type')
+      expect(page).to have_content('Total cases')
+    end
+  end
+end
```
starswan, line 42: ditto to above

app/controllers/poms_controller.rb
ollietreend: In case this is of any interest, there's a nice example of succinctly splitting Prison Officer / Probation Officer in the allocations controller using the `#partition` method:

https://github.com/ministryofjustice/offender-management-allocation-manager/blob/32a6a5ef9aa5ee3ecbfdc867bc0452ed534e4822/app/controllers/allocations_controller.rb#L184-L190

Which could be adapted to something like:

```suggestion
      @active_probation_poms, @active_prison_poms = active_poms.sort_by(&:last_name).map { |p| PomPresenter.new(p) }.partition(&:probation_officer?)
```

One advantage of this approach is performance: you only end up sorting and looping through the array once, rather than twice, to end up with two arrays. In this case I expect the arrays are very small, so the performance gain is negligible, but it might also DRY up the code a little.

starswan (approved):

PR #1636, opened 2021-03-31
Remove superfluous nomis_booking_id from Allocation object.

starswan:
Whilst working on the allocation story, I came across the fact that we store the nomis_booking_id in our allocation object.

However...its never used (our public interface to the Prison API/NOMIS is always by offender_no, although we do have to retrieve sentence information via booking id) and it's just noise in an already over-complicated object. 

... reviewers - slightly worried that this might break existing allocation histories?


Katy600 (approved):
Wow! I can't believe it isn't being used and it is everywhere!

PR #1638, opened 2021-04-01
MO-576 set Rails Timezone to London time 

starswan:
so that times don't come out in GMT


spec/features/allocation_history_spec.rb
```diff
@@ -2,7 +2,7 @@
 
 require 'rails_helper'
 
-feature 'Allocation History' do
+xfeature 'Allocation History' do
   let(:probation_pom) do
```
Katy600: Is this feature test pending?

spec/features/allocation_history_spec.rb
```diff
@@ -222,7 +222,15 @@
               expect(page).to have_css(key, text: val)
             end
           end
+        end
+      end
+
+      it 'has a Pentonville section', :js do
+        # 2nd Prison - Pentonville. This contains 7 events
```
Katy600, line 229: Now you have split the test into 2 does this test need this comment now? Is this now covered in the first test?

Katy600 (approved):
Hi Stephen, 
I have tried really hard to find things to say. I think my comments are mostly about trying to understand the tests but these are not actually your tests! So feel free to ignore them.

spec/features/allocation_history_spec.rb
starswan: reverted

PR #1639, opened 2021-04-01
MO-557 (take II) added Prison.all method

starswan:
The previous fix for this issue didn't take into account that there was no Prison.all method.

This PR addresses that issue


Katy600 (approved):
I have just approved this! I can't think of anything to comment on. :)

PR #1653, opened 2021-04-08
MO-64 Part 3 - move push handover to delius out to its own (retriable) job

starswan:
This change moves the push handover to delius operation out to its own job (and out of the save action of the model). 

This allows the job to be retry-able once more and not to catch exceptions - if after N retries the job fails, its not going to work so it is fine for it to be discarded

### DIFF
--- a/app/jobs/recalculate_handover_date_job.rb
+++ b/app/jobs/recalculate_handover_date_job.rb
@@ -5,16 +5,34 @@ class RecalculateHandoverDateJob < ApplicationJob
 
   def perform(nomis_offender_id)
     offender = OffenderService.get_offender(nomis_offender_id)
-    return if offender.nil? || offender.sentenced? == false
-
-    begin
+    if offender&.inside_omic_policy?
       # Recalculate handover dates, which will trigger a push to the Community API after_save
-      CalculatedHandoverDate.recalculate_for(offender)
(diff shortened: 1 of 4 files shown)


app/jobs/recalculate_handover_date_job.rb
```diff
@@ -5,16 +5,35 @@ class RecalculateHandoverDateJob < ApplicationJob
...
+    record.save! if record.changed?
+    push_to_delius record
+  end
+
+  def push_to_delius record
+    # Don't push if the CaseInformation record is a manual entry (meaning it didn't match against nDelius)
+    # This avoids 404 Not Found errors for offenders who don't exist in nDelius (they could be Scottish, etc.)
+    # Don't push if the dates haven't changed
+    if !record.case_information.manual_entry &&
+      record.saved_change_to_start_date? || record.saved_change_to_handover_date?
```
ollietreend, line 35: I think you might need to add brackets in here:

```suggestion
    if !record.case_information.manual_entry &&
      ( record.saved_change_to_start_date? || record.saved_change_to_handover_date? )
```

Without them, I think it could incorrectly push handover dates on manual entries.

According to `irb`:

```ruby
false && true || true
# => true

false && ( true || true )
# => false
```

Looks like this combination isn't covered by any tests..?

ollietreend (approved):

PR #1657, opened 2021-04-09
MO-64 Part 4 send chaser emails every 48 hours when no COM and < 10 months left to serve

starswan:


app/models/hmpps_api/offender_base.rb
```diff
@@ -255,15 +255,8 @@ def criminal_sentence?
...
-    # Take either the new LocalDeliveryUnit (if available and enabled) and
-    # fall back to the old local_divisional_unit if not. This should all go away
-    # in Feb 2021 after the PDU changes have been rolled out in nDelius
     def ldu
-      if @case_information&.local_delivery_unit&.enabled?
-        @case_information.local_delivery_unit
-      else
-        @case_information&.local_divisional_unit
-      end
+      @case_information&.ldu
```
ollietreend, line 259: This entire `Offender#ldu` method might be a bit pointless now. Could we lose this and simply use `@case_information&.ldu` in the `ldu_email` and `ldu_name` methods?

db/migrate/20210407101956_add_responsibility_to_calculated_handover_dates.rb
```diff
@@ -0,0 +1,10 @@
+class AddResponsibilityToCalculatedHandoverDates < ActiveRecord::Migration[6.0]
+  def change
+    change_table :calculated_handover_dates do |t|
+      # This ought to be non-nullable, but we have existing data
+      t.boolean :pom_responsible
+      # This is deliberately nullable as the COM is sometimes not involved
+      t.boolean :com_responsible
```
ollietreend, line 7: See my previous comments.

It'd be really nice if we could be consistent with these fields. I think they would be better suited as representing the entire `pom_responsibility` and `com_responsibility` `HandoverData` objects, rather than just one field (`responsible?`).

I don't like the feel of these special comments floating around continually needing to explain the different between the POM/COM responsibilities (i.e. one can be true/false, the other can be true/false/nil)

ollietreend:
@

ollietreend:
@starswan I like where you're going with this, but have added a couple of comments around the naming and values stored in the new responsibility fields of `CalculatedHandoverDate`.

ollietreend (approved):

PR #1658, opened 2021-04-12
MO-599 - stop using cta_for_offender in search results

starswan:
SearchHelper#cta_for_offender duplicates the logic of splitting out the offenders in a prison into 4 seperate 'buckets'.

This logic  changed with women's prisons, and so cta_for_offender broke. This PR stops using cta_for_offender in the search results, instead using the 4 'buckets' to make the decision about which link to display


spec/controllers/summary_controller_spec.rb
```diff
@@ -362,4 +362,89 @@
...
+        expect(response).to be_successful
+
+        expect(assigns(:q)).to eq('Cal')
+        expect(assigns(:offenders).size).to eq(0)
+      end
+
+      context "with 3 offenders", :allocation do
+        let(:offenders) { build_list(:nomis_offender, 3, lastName: 'Bloggs', firstName: 'Alice') }
+        let(:updated_offenders) { assigns(:offenders) }
+        let(:alloc_offender) { updated_offenders.detect { |o| o.offender_no == offenders.first.fetch(:offenderNo) } }
```
Katy600, line 405: Stephen, I am trying to understand what alloc_offender does. It compares the first offender from offenders list with the updated_offenders and returns the updated_offenders match. This is then assigns to the name alloc_offender. Oh I think I am confused!! How does this make it allocated?

starswan: It's not great I agree It's trying to pick out the first one as an 'offender to be allocated' I think. Let me try and re-work it

Katy600 (approved):
Oh it took me so long! I found it really helpful to try and get my head around though. I really like what you have done with the FemalePrisonersController. 
I shall approve it but I imagine you will want Ollie's eyes over it too.

PR #1661, opened 2021-04-14
Fixup tests as a result of Andrien leaving

starswan:


spec/services/allocation_service_spec.rb
```diff
@@ -71,12 +71,21 @@
     context 'without an existing' do
       before do
         create(:case_information, nomis_offender_id: nomis_offender_id)
+        stub_auth_token
+        stub_request(:get, "#{ApiHelper::T3}/users/MOIC_POM").
+          to_return(body: { staffId: 1, firstName: "MOIC", lastName: 'POM' }.to_json)
+        stub_pom_emails 1, []
```
ollietreend, line 77: I think we have a helper for this:

`stub_spo_user`

https://github.com/ministryofjustice/offender-management-allocation-manager/blob/f6163fa281f473c283f82a1444dcb82469d706ad/spec/support/helpers/features_helper.rb#L14-L19

ollietreend (approved):

PR #1665, opened 2021-04-16
MO-607 Error during women's missing info journey

Katy600:
This PR fixes a bug where HOMD had been opening multiple tabs to allow them to complete the missing info for several prisoners at once. 
Because the session keys in the new action were static this meant that having multiple tabs open overrode the session keys causing a crash. 

The fix was to create dynamic session keys for complexity and case info in the new action.

### DIFF
--- a/spec/features/female_missing_info_feature_spec.rb
+++ b/spec/features/female_missing_info_feature_spec.rb
@@ -5,15 +5,17 @@ require "rails_helper"
 feature "womens missing info journey" do
   let(:test_strategy) { Flipflop::FeatureSet.current.test! }
   let(:prison) { build(:womens_prison) }
-  let(:offender) { build(:nomis_offender, agencyId: prison.code, complexityLevel: complexity) }
+  let(:offenders) { build_list(:nomis_offender, 2, agencyId: prison.code, complexityLevel: complexity) }
+  let(:offender) { offenders.first }
   let(:prisoner_id) { offender.fetch(:offenderNo) }
+  let(:second_prisoner_id) { offenders.last.fetch(:offenderNo) }
   let(:user) { build(:pom) }
   let(:username) { 'MOIC_POM' }
 
   before do
     stub_signin_spo user, [prison.code]
     stub_poms(prison.code, [user])
-    stub_offenders_for_prison(prison.code, [offender])
+    stub_offenders_for_prison(prison.code, offenders)
     test_strategy.switch!(:womens_estate, true)
   end
 
@@ -30,8 +32,9 @@ feature "womens missing info journey" do
 
     it 'has a happy path' do
       visit missing_information_prison_prisoners_path prison.code
-      click_link 'Add missing details'
-
+      within "#edit_#{prisoner_id}" do
+        click_link 'Add missing details'
+      end
       # bang the update button - expect a big red error message
       click_button 'Update'
       within '.govuk-error-summary' do
@@ -40,6 +43,7 @@ feature "womens missing info journey" do
       end
 
       find('label[for=complexity-form-complexity-level-low-field]').click
+
       click_button 'Update'
       expect {
         #  again should have lots of error messages
@@ -84,7 +88,9 @@ feature "womens missing info journey" do
       expect(HmppsApi::ComplexityApi).to receive(:save).with(prisoner_id, level: 'medium', username: username, reason: nil)
 
       visit missing_information_prison_prisoners_path prison.code
-      click_link 'Add missing details'
+      within "#edit_#{prisoner_id}" do
+        click_link 'Add missing details'
+      end
 
       find('label[for=complexity-form-complexity-level-medium-field]').click
     end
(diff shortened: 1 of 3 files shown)


spec/features/female_missing_info_feature_spec.rb
```diff
@@ -119,4 +126,39 @@
...
+      switch_to_window(current)
+
+      find('label[for=case-information-probation-service-england-field]').click
+      find('label[for=case-information-case-allocation-nps-field]').click
+      find('label[for=case-information-tier-a-field]').click
+      click_button 'Update'
+
+      # check both updates work
+      expect(CaseInformation.find_by(nomis_offender_id: prisoner_id)).not_to be_nil
+      expect(CaseInformation.find_by(nomis_offender_id: second_prisoner_id)).not_to be_nil
```
starswan, line 161: might be nice to tear down the second tab after you've finished...

spec/features/female_missing_info_feature_spec.rb
```diff
@@ -119,4 +126,39 @@
       wait_for { page.current_path == missing_information_prison_prisoners_path(prison.code) }
     end
   end
+
+  context 'when multiple tabs are opened', :js do
+    let(:complexity) { 'medium' }
```
starswan, line 131: given the change, I think this test should have complexity nil so that both screens are filled out both times...

spec/features/female_missing_info_feature_spec.rb
```diff
@@ -119,4 +126,39 @@
...
+      find('label[for=case-information-case-allocation-nps-field]').click
+      find('label[for=case-information-tier-a-field]').click
+      click_button 'Update'
+
+      switch_to_window(current)
+
+      find('label[for=case-information-probation-service-england-field]').click
+      find('label[for=case-information-case-allocation-nps-field]').click
+      find('label[for=case-information-tier-a-field]').click
+      click_button 'Update'
```
starswan, line 157: would be good to DRY this repeated block into a method

spec/features/female_missing_info_feature_spec.rb
```diff
@@ -119,4 +126,39 @@
...
+    let(:complexity) { 'medium' }
+
+    it 'can complete forms in parallel' do
+      visit missing_information_prison_prisoners_path prison.code
+      window = page.open_new_window
+      current = page.current_window
+      within "#edit_#{prisoner_id}" do
+        click_link 'Add missing details'
+      end
+      switch_to_window(window)
```
starswan, line 140: As you wanted a 'do things in current tab' method in capybara, would be really nice to create a helper that does implement this interface i.e. creates a new window/tab, switches to it, yields to a block, and then switches back and destroys the tab. Would make the test really nice and readable too.

spec/features/female_missing_info_feature_spec.rb
```diff
@@ -119,4 +126,39 @@
...
     end
   end
+
+  context 'when multiple tabs are opened', :js do
+    let(:complexity) { 'medium' }
+
+    it 'can complete forms in parallel' do
+      visit missing_information_prison_prisoners_path prison.code
+      window = page.open_new_window
+      current = page.current_window
```
ollietreend, line 136: These variable names are a bit unclear.

Would it be possible to name them `tab_1` and `tab_2` or something – to make it really obvious that we're switching between 2 tabs in this test?

ollietreend (commented):
This looks great @Katy600 – the parallel / multiple tab test is new to me! 🤓

spec/features/female_missing_info_feature_spec.rb
```diff
@@ -119,4 +127,50 @@
...
+        end
+
+        find('label[for=complexity-form-complexity-level-high-field]').click
+
+        click_button 'Update'
+
+        fill_in_missing_case_information
+
+        click_button 'Update'
+      end
```
ollietreend, line 163: Unless I'm misunderstanding what's going on here – aren't these two things happening one after the other? It doesn't look like they're happening in parallel any more.

Really we need both tabs to be on the "Add missing details" page **at the same time** – and then "Update" one before the other. Ultimately: replicating the behaviour of a user who opens multiple tabs and then updates them one by one.

I think without that, we're not actually proving that the bug is fixed..?

Happy to be corrected if I've got that wrong!

starswan: Yes you're right @ollietreend the second journey should happen *after* the first one (which is what it used to do)

Katy600: Oh I see. I have put it in the wrong order! Thank you @ollietreend

spec/support/helpers/features_helper.rb
```diff
@@ -50,4 +50,17 @@ def stub_user(username: 'MOIC_POM', staff_id:)
...
+    tab_one = page.open_new_window
+    tab_two = page.current_window
+
+    switch_to_window(tab_one)
+
+    yield
+
+    switch_to_window(tab_two)
+
+    tab_one.close
```
starswan, line 64: Think these names might be better as current_window and new_window...

spec/support/helpers/api_helper.rb
```diff
@@ -56,6 +56,9 @@ def stub_offender(offender)
 
     stub_request(:get, "#{Rails.configuration.complexity_api_host}/v1/complexity-of-need/offender-no/#{offender_no}").
       to_return(body: { level: offender.fetch(:complexityLevel) }.to_json)
+
+    stub_request(:post, "#{Rails.configuration.complexity_api_host}/v1/complexity-of-need/offender-no/#{offender_no}").
+      to_return(status: 200, body: { level: offender.fetch(:complexityLevel) }.to_json, headers: {})
```
starswan, line 61: This isn't needed - it is a copy of the line above it with some of the defaults explictly added

starswan (approved):

PR #1669, opened 2021-04-19
MO-486 rewrite staff/poms controller

starswan:


app/views/poms/_poms_table.html.erb
```diff
@@ -1,31 +1,30 @@
...
-      <td aria-label="Total" class="govuk-table__cell "><%= pom.total_cases %></td>
+      <td aria-label="POM" class="govuk-table__cell"><%= link_to full_name_ordered(pom), prison_pom_path(@prison.code, nomis_staff_id: pom.staff_id) %></td>
       <td aria-label="Working pattern" class="govuk-table__cell"><%= format_working_pattern(pom.working_pattern) %></td>
-      <td aria-label="Status" class="govuk-table__cell "><%= pom.status.capitalize %></td>
-      <td aria-label="Action" class="govuk-table__cell "><%= link_to 'View', prison_pom_path(@prison.code, nomis_staff_id: pom.staff_id) %></td>
+      <td aria-label="Status" class="govuk-table__cell"><%= status(pom).capitalize %></td>
+      <td aria-label="Case mix by tier" class="govuk-table__cell"><%= case_mix_bar(pom) %></td>
+      <% if @prison.womens? %>
+        <td aria-label="High complexity cases" class="govuk-table__cell govuk-table__cell--numeric">
+          <%= @offenders.count { |o| pom.allocations.map(&:nomis_offender_id).include?(o.offender_no) && o.complexity_level == 'high' } %>
```
ollietreend, line 24: This is a bit complex for view logic.

But if we did implement `PomWrapper#count_offenders` then we could end up simplifying this to something as simple as:

```ruby
pom.count_offenders { |o| o.complexity_level == 'high' }
```

app/views/poms/_poms_table.html.erb
```diff
@@ -1,31 +1,30 @@
...
-      <td aria-label="Tier C cases" class="govuk-table__cell "><%= pom.tier_c %></td>
-      <td aria-label="Tier D cases" class="govuk-table__cell "><%= pom.tier_d %></td>
-      <td aria-label="No tier cases" class="govuk-table__cell "><%= pom.no_tier %></td>
-      <td aria-label="Total" class="govuk-table__cell "><%= pom.total_cases %></td>
+      <td aria-label="POM" class="govuk-table__cell"><%= link_to full_name_ordered(pom), prison_pom_path(@prison.code, nomis_staff_id: pom.staff_id) %></td>
       <td aria-label="Working pattern" class="govuk-table__cell"><%= format_working_pattern(pom.working_pattern) %></td>
-      <td aria-label="Status" class="govuk-table__cell "><%= pom.status.capitalize %></td>
-      <td aria-label="Action" class="govuk-table__cell "><%= link_to 'View', prison_pom_path(@prison.code, nomis_staff_id: pom.staff_id) %></td>
+      <td aria-label="Status" class="govuk-table__cell"><%= status(pom).capitalize %></td>
+      <td aria-label="Case mix by tier" class="govuk-table__cell"><%= case_mix_bar(pom) %></td>
```
ollietreend, line 21: Likewise, the `case_mix_bar` helper method could perform its own counts rather than depending on `PomWrapper#tier_a`, `PomWrapper#tier_b`, and so on...

```ruby
pom.count_offenders { |o| o.tier == 'A' }
# and so on...
```

app/views/poms/index.html.erb
```diff
@@ -1,9 +1,41 @@
...
-<% else %>
-  <%= render 'male_index' %>
-<% end %>
+<div class="govuk-tabs" data-module="govuk-tabs">
+  <h2 class="govuk-tabs__title">Manage your staff</h2>
+
+  <ul class="govuk-tabs__list" role='tablist'>
+    <li class="govuk-tabs__list-item govuk-tabs__list-item--selected">
+      <a class="govuk-tabs__tab" href="#active_probation_poms">
+        Active Probation officer POMs (<%= @poms.count { |pom| %w[active unavailable].include?(pom.status) && pom.probation_officer? } %>)
```
ollietreend, line 12: ### Opportunity to DRY up

It seems a shame to be duplicating this special filtering logic throughout this view. We perform the same filter twice: once for the count, and once to render the list.

Previously splitting was performed in the Controller, so we had:
`@active_probation_poms`, `@active_prison_poms` and `@inactive_poms`
These were nice & easy to use in the view – but I imagine your intention is for the View layer to perform this filtering (since it's only done for display purposes)?

Could this filtering logic be put into helper methods instead of cluttering the view?

`active_probation_poms`, `active_prison_poms` and `inactive_poms` could conceivably become helper methods...

db/seeds.rb
```diff
@@ -1,33 +1,38 @@
 # Jay Heal
 PomDetail.find_or_create_by!(
+           prison_code: 'LEI',
```
ollietreend, line 3: The indentation looks a bit weird here 🤷🏻

spec/features/edit_a_pom_spec.rb
```diff
@@ -43,17 +45,16 @@
   end
 
   it "makes an inactive POM active", vcr: { cassette_name: 'prison_api/edit_poms_activate_pom_feature' } do
+    # This doesn't do what it appears to - the URL is wrong so we're not editing an inactive POM... :-(
```
ollietreend, line 48: If this is a non-JS test, the `#inactive` will have no effect anyway.

Could we just remove it to avoid the confusing URL?

ollietreend (commented):
I've added some comments – but don't take that as a rejection.

I think this PR looks great – it's moving us in the right direction! 🙌🏻

PR #1670, opened 2021-04-20
MO-199 remove dependency on imprisonmentStatus to decide indeterminate/determinate sentences

starswan:
This changes our sentence detail object so that rather than relying on sentenceType  / imprisonmentStatus to decide whether the offence is determinate or indeterminate, it uses a value provided by the PrisonerSearch API instead


ollietreend (dismissed):

ollietreend (commented):
Apparently there's a subtle flaw in this somewhere – so un-approving for now.

ollietreend (approved):

PR #1672, opened 2021-04-21
MO-629 make PrisonOffenderManageService a bit smaller

starswan:
remove small single-use methods that do nothing to help abstraction for the code base.

This PR helps PrisonOffenderManagerService tremendously, in that it shrinks it down to 2 methods - get_pom_at and get_poms_for. It should be reviewed and merged ahead of MO-486, which struggles to cope with the complexity of this class - simplifying it should make some of the issues with that PR go away

### DIFF
--- a/app/helpers/pom_helper.rb
+++ b/app/helpers/pom_helper.rb
@@ -36,9 +36,8 @@ module PomHelper
   end
 
   def fetch_pom_name(staff_id)
-    pom_firstname, pom_secondname =
-      PrisonOffenderManagerService.get_pom_name(staff_id)
-    "#{pom_secondname}, #{pom_firstname}"
+    staff = HmppsApi::PrisonApi::PrisonOffenderManagerApi.staff_detail(staff_id)
+    "#{staff.last_name}, #{staff.first_name}"
   end
 
   def status(pom)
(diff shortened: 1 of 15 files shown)


app/helpers/pom_helper.rb
```diff
@@ -36,9 +36,8 @@ def grade(pom)
   end
 
   def fetch_pom_name(staff_id)
-    pom_firstname, pom_secondname =
-      PrisonOffenderManagerService.get_pom_name(staff_id)
-    "#{pom_secondname}, #{pom_firstname}"
+    staff = HmppsApi::PrisonApi::PrisonOffenderManagerApi.staff_detail(staff_id)
+    "#{staff.last_name}, #{staff.first_name}"
```
ollietreend, line 40: I'm not convinced our View layer should be calling out to our API... 😕
But I realise you haven't introduced that behaviour in this PR. It was always there, just hidden in the Service object.

ollietreend (approved):

app/helpers/pom_helper.rb
starswan: Thanks for the comment - turns out it isn't. Moved this 'helper' method in the other direction into the service

PR #1674, opened 2021-04-22
MO-426 pre-factor: retrieve the date an offender's category was set

ollietreend:
This commit refactors the way offender categories are retrieved from the Prison API, and adds a new 'active since' date to indicate when the offender's current category was set.

We'd previously used the `SUP_LVL_TYPE` endpoint to retrieve a dictionary mapping category codes => human-friendly labels. However it turns out that we didn't need to do this, since we already retrieved the category label when calling the `/offender-assessments/CATEGORY` endpoint – we just weren't using it.

So I've refactored the codebase to make better use of the data we receive from the `/offender-assessments/CATEGORY` endpoint.

This involved introducing a new `HmppsApi::OffenderCategory` model which holds information relating to an offender's categorisation. This is made available on the Offender model as `category_code`, `category_label` and `category_active_since`.

---

⚠️ **Note to devs**

This will break your VCR cassettes. The first time you RSpec from this branch, you'll need to run with `VCR=2` to re-write your local VCR cache.

### DIFF
--- a/app/models/hmpps_api/offender.rb
+++ b/app/models/hmpps_api/offender.rb
@@ -9,12 +9,12 @@ module HmppsApi
     attr_reader :prison_id
 
     # This must only reference fields that are present in
-    # https://https://api-dev.prison.service.justice.gov.uk/swagger-ui.html#//prisoners/getPrisonersOffenderNo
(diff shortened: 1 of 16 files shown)


ollietreend:
Closed as incorporated into #1676

PR #1676, opened 2021-04-23
MO-426 pre-factor: Use female category for start of handover date

ollietreend:
In the women's estate, use the date that their category became "female open" to determine when their handover starts.
    
This replicates the behaviour of the men's estate, where their start of handover date is the day that they moved into a male open prison.

---

Based upon #1674 pre-factor. The only commit of interest in this branch is 11b788c.

### DIFF
--- a/app/models/hmpps_api/offender_base.rb
+++ b/app/models/hmpps_api/offender_base.rb
@@ -13,7 +13,9 @@ module HmppsApi
              :parole_review_date, :victim_liaison_officers, :updated_at,
              to: :@case_information, allow_nil: true
 
-    attr_accessor :category_code, :date_of_birth, :prison_arrival_date, :sentence, :allocated_pom_name
+    delegate :code, :label, :active_since, to: :@category, prefix: :category, allow_nil: true
+
+    attr_accessor :date_of_birth, :prison_arrival_date, :sentence, :allocated_pom_name
 
     attr_reader :first_name, :last_name, :booking_id, :offender_no, :sentence_type, :cell_location, :complexity_level
 
@@ -152,7 +154,7 @@ module HmppsApi
     # https://api-dev.prison.service.justice.gov.uk/swagger-ui.html#//prisoners/getPrisonersOffenderNo
     # and also by
     # https://api-dev.prison.service.justice.gov.uk/swagger-ui.html#//locations/getOffendersAtLocationDescription
-    def initialize(api_payload, search_payload, latest_temp_movement:, complexity_level:)
+    def initialize(api_payload, search_payload, category:, latest_temp_movement:, complexity_level:)
       # It is expected that this method will be called by the subclass which
       # will have been given a payload at the class level, and will call this
       # method from it's own internal from_json
@@ -160,11 +162,11 @@ module HmppsApi
       @last_name = api_payload.fetch('lastName')
       @offender_no = api_payload.fetch('offenderNo')
       @convicted_status = api_payload['convictedStatus']
-      @category_code = api_payload['categoryCode']
       @date_of_birth = Date.parse(api_payload.fetch('dateOfBirth'))
       @latest_temp_movement = latest_temp_movement
       @cell_location = search_payload['cellLocation']
       @complexity_level = complexity_level
+      @category = category
     end
 
     def handover_start_date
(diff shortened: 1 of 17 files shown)


app/models/hmpps_api/offender_base.rb
```diff
@@ -13,7 +13,9 @@ class OffenderBase
              :parole_review_date, :victim_liaison_officers, :updated_at,
              to: :@case_information, allow_nil: true
 
-    attr_accessor :category_code, :date_of_birth, :prison_arrival_date, :sentence, :allocated_pom_name
+    delegate :code, :label, :active_since, to: :@category, prefix: :category, allow_nil: true
```
starswan, line 16: not sure about the allow_nil - surely category will always be filled in?

app/services/hmpps_api/prison_api/offender_api.rb
```diff
@@ -40,6 +41,7 @@ def self.list(prison, page = 0, page_size: 20)
           search_payload = search_data.fetch(offender_no)
           HmppsApi::OffenderSummary.new(payload,
                                         search_payload,
+                                        category: offender_categories[offender_no],
```
starswan, line 44: this allows nils - think it should be a fetch? If we're worried about inconsistent caches again, then offender doesn't exist if they don't have a category?

app/services/hmpps_api/prison_api/offender_api.rb
```diff
@@ -63,11 +65,13 @@ def self.get_offender(raw_offender_no)
           nil
         else
           temp_movements = HmppsApi::PrisonApi::MovementApi.latest_temp_movement_for([offender_no])
+          offender_categories = get_offender_categories([offender_no])
           complexity_level = if api_response.fetch('currentlyInPrison') == 'Y' && PrisonService::womens_prison?(api_response.fetch('latestLocationId'))
                                HmppsApi::ComplexityApi.get_complexity(offender_no)
                              end
           prisoner = HmppsApi::Offender.new api_response,
                                             search_payload,
+                                            category: offender_categories[offender_no],
```
starswan, line 74: ditto as above

spec/factories/offender_categories.rb
```diff
@@ -0,0 +1,56 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :offender_category, class: HmppsApi::OffenderCategory do
+    initialize_with { HmppsApi::OffenderCategory.new(attributes.stringify_keys) }
+
+    offenderNo { '' }
```
starswan, line 7: not sure about defaulting this to blank - maybe better to force callers to fill it in? Feels like a mandatory has_one relationship

spec/factories/offenders.rb
```diff
@@ -28,6 +28,7 @@
     # in tests, as ruby sort isn't stable by default
     sequence(:lastName) { |c| "#{Faker::Name.last_name.titleize}_#{c}" }
     categoryCode { 'C' }
```
starswan, line 30: categoryCode now goes away from this factory

spec/factories/offenders.rb
```diff
@@ -84,6 +86,7 @@
...
```
starswan, line 88: again, categpryCode can be removed

app/models/hmpps_api/offender_base.rb
ollietreend: It looks like offenders don't _always_ have a category. If they're new in prison and an offender categorisation assessment hasn't been completed yet, then their category will legitimately be `nil`.

I've added a test to `offender_spec.rb` to cover this scenario, to make it clear that this is expected behaviour.

app/services/hmpps_api/prison_api/offender_api.rb
ollietreend: As above: there might not be an active category assessment for every offender, so sometimes this can be `nil` (and that's expected behaviour).

spec/factories/offender_categories.rb
ollietreend: Good point. I've refactored this factory to remove the need for `offenderNo` to be explicitly specified.

starswan (approved):

PR #1677, opened 2021-04-26
MO-584 unify Offender objects

starswan:
This PR unifies the Offender and OffenderSummary classes by combining the last set of differences into one Offender class. This should make progress in the codebase a lot easier.
However it doesn't address the attr_accessor for :main_offence left over in Offender


app/models/hmpps_api/offender.rb
```diff
@@ -1,20 +1,289 @@
...
+      @convicted_status = api_payload['convictedStatus']
+      @date_of_birth = Date.parse(api_payload.fetch('dateOfBirth'))
+      @latest_temp_movement = latest_temp_movement
+      @cell_location = search_payload['cellLocation']
+      @complexity_level = complexity_level
+      @category = category
+
+      @booking_id = booking_id
+      @prison_id = prison_id
+    end
```
ollietreend, line 185: I tend to prefer having the constructor method at the top of the class file – making it really obvious what goes 'in' to the object, before seeing the methods that come 'out' of it.

Right now it's just mixed in amongst all the other methods, which all happen to be in no particular order.

This could just be personal preference, so feel free to ignore.

app/models/hmpps_api/offender.rb
```diff
@@ -1,20 +1,289 @@
...
+
+    def load_case_information(record)
+      return if record.blank?
+
+      # Hold on to the CaseInformation record so we can reference it later
+      @case_information = record
+
+      # This is separate from @case_information so the rest of this class doesn't need to know about the Active Record association
+      @responsibility_override = record.responsibility
+    end
```
ollietreend, line 207: It might also be nice to consider this another 'constructor' method and keep it at the top of the file alongside the `initialize` method. It's another method that puts something 'in' to the object for the other methods to use.

ollietreend (approved):
Looks good 👍🏻 

I've added a couple of comments which are just opinions. Feel free to ignore if you don't agree!

PR #1678, opened 2021-04-26
MO-298 Remove old LDU and team data

starswan:
Ths PR drops the old team and LocalDevisonalUnit tables. 

It also removes all the old code to do with parsing spreadsheets and downloading same via IMAP


db/migrate/20201014082050_add_welsh_flag_to_ldu.rb
```diff
@@ -1,6 +1,10 @@
 # frozen_string_literal: true
 
 class AddWelshFlagToLdu < ActiveRecord::Migration[6.0]
+  # Need to define this for Heroku review apps which run all the migrations
+  class LocalDivisionalUnit < ApplicationRecord
+  end
+
```
ollietreend, line 7: Why does Heroku run all migrations rather than building the database from `schema.rb`?

This seems a bit weird? 🤔 Surely that means any time we remove/rename a model, we'll need to add this 'workaround' to keep Heroku happy. That seems wrong to me.

starswan: It seems that it wants to 'update' the schema (seems reasonable) but for review apps the schema is empty - hence it runs all the migrations

spec/models/hmpps_api/offender_spec.rb
```diff
@@ -325,9 +325,9 @@
         end
 
         context 'without a new LDU mapping' do
```
ollietreend, line 327: We should probably remove references to a "new LDU" mapping.

There should be no such thing as "old" and "new" LDUs after this PR.

spec/services/offender_service_spec.rb
```diff
@@ -71,6 +71,7 @@
                   offender_manager: nil,
                   service_provider: "NPS",
                   team_code: "N07UAT",
```
ollietreend, line 73: As above: I don't think `team_code` is needed any more.

If we don't have any actual code using it, then we should remove it.

spec/services/view_handover_email_addresses_summary_spec.rb
```diff
@@ -48,50 +48,11 @@
...
   end
 
-  context 'with an offender that has case information but no team link' do
-    let(:offenders) { [double(offender_no: 1)] }
-
-    before { create(:case_information, nomis_offender_id: 1, team: nil) }
-
-    it 'returns the right counts' do
-      expect(result).to eq(
-        has_email_address: 0,
```
ollietreend, line 58: As above: we can safely delete this entire service object and its associated rake task. It was never needed, and has never been used.

ollietreend:
@starswan it'd also be good to remove the data files in `app/data`. None of them are needed, and they relate to these old LDUs/teams.

PR #1681, opened 2021-04-27
MO-621 create prisoners model

starswan:
Note - PRd not done (yet) in this spike


app/services/offender_service.rb
```diff
@@ -5,8 +5,8 @@ def self.get_offender(offender_no)
     HmppsApi::PrisonApi::OffenderApi.get_offender(offender_no).tap { |o|
       next false if o.nil?
 
-      record = CaseInformation.find_by(nomis_offender_id: offender_no)
-      o.load_case_information(record)
+      offender = Offender.find_or_create_by!(nomis_offender_id: offender_no)
+      o.load_case_information(offender.case_information)
```
ollietreend, line 9: Shouldn't we be doing the same in the `prison.offenders` method?

The method `Prison::OffenderEnumerator#enrich_offenders` currently uses `CaseInformationService` to get offenders. But it doesn't auto-create offender records – meaning 'full fat' offender objects will probably end up without case information being loaded.

app/services/case_information_service.rb
```diff
@@ -2,7 +2,7 @@
 
 class CaseInformationService
   def self.get_case_information(offender_ids)
-    CaseInformation.includes(:responsibility, :early_allocations, :local_delivery_unit).
-      where(nomis_offender_id: offender_ids).index_by(&:nomis_offender_id)
+    Offender.includes(case_information: [:responsibility, :early_allocations, :local_delivery_unit]).
+      where(nomis_offender_id: offender_ids).map(&:case_information).compact.index_by(&:nomis_offender_id)
   end
 end
```
ollietreend, line 8: This entire service object is only used by one consumer – `Prison::OffenderEnumerator#enrich_offenders`.

We should probably just roll it into that. It doesn't make sense to have a whole service object just for this.

ollietreend (approved):

PR #1685, opened 2021-04-29
MO-670 Category#active_since falls back to assessmentDate if approvalDate not present

starswan:


Katy600 (approved):
It makes sense to me :)

PR #1699, opened 2021-04-30
Mo 642 create a database driven prison model

Katy600:
This PR replaces the old Prison model, which was a plain-old Ruby object, with new Prison model, which is a database-backed ActiveRecord model.

1. Creates a new ActiveRecord model called Prison

2. Creates a database seed file to automatically populate the prisons table using the existing data held in the PrisonService.

3. Replaces the old Prison model but keeps a number of the previous methods such as PrisonEmumerator and OffenderEnumerator. It removes methods such as _all_ that are now done automatically as part of a rails model, and modifies the method womens for the same reason.

4. It tries to remove a lot of PrisonService service object but this is very tricky as the method PrisonService.prison_codes is used in email history, which is part of AllocationHistory. AllocationHistory doesn't have an association with Prison but will be able to get Prison through the Offender, via the PrisonTime line eventually. But the former has not yet been refactored and the latter has not been created yet. So once these things have been done it will hopefully make it easier to remove more of the PrisonService object and eventually the whole of PrisonService object. 

5. There is a lot of tidying up in the tests as now Prison is no longer an object but is a database backed model the tests set up requires you to use create(:prison) as opposed to build(:prison) This caused an issue with the VCR tests, as changing all the prison set up to _create_ resulted in errors validation errors due to duplicated prisons. This was solved by creating a prison, once, inside rails_helper as part of Webmock.


db/seeds.rb
```diff
@@ -1,166 +1,23 @@
...
-                           case_allocation:'NPS',
-                           probation_service: 'Wales',
-                           manual_entry: true,
-                           team: team2)
+PrisonService::PRISONS.values.each do |prison|
+  # If prison data is already present it runs update instead to prevent a crash
+  if Prison.where(code: prison.code).exists?
+    Prison.update(code: prison.code, name: prison.name, prison_type: prison_type(prison))
+  else
+    Prison.create!(code: prison.code, name: prison.name, prison_type: prison_type(prison))
```
ollietreend, line 20: Is the intention that this runs on every deployment in production environments?

This used to only be run manually by devs setting up their local installs – so I'm not sure that it's automatically running on production deploys.

app/controllers/debugging_controller.rb
```diff
@@ -18,7 +18,7 @@ def debugging
   end
 
   def prison_info
-    @prison_title = PrisonService.name_for(active_prison_id)
+    @prison_title = Prison.find_by!(code: active_prison_id).name
```
starswan, line 21: This is just @prison.name

app/controllers/pages_controller.rb
```diff
@@ -8,8 +8,8 @@ def contact_us
       @user = HmppsApi::PrisonApi::UserApi.user_details(current_user)
       @contact = ContactSubmission.new(email_address: @user.email_address.first,
                                        name: @user.full_name_ordered,
-                                       prison: PrisonService.name_for(
-                                         @user.active_case_load_id)
+                                       prison: Prison.find_by!(code:
```
starswan, line 11: again @prison.name

app/controllers/responsibilities_controller.rb
```diff
@@ -34,7 +34,7 @@ def create
         message: params[:message],
         prisoner_number: @responsibility.nomis_offender_id,
         prisoner_name: @offender.full_name,
-        prison_name: PrisonService.name_for(@prison.code),
+        prison_name: Prison.find_by!(code: @prison.code).name,
```
starswan, line 37: @prison.name

app/jobs/auto_early_allocation_email_job.rb
```diff
@@ -15,7 +15,7 @@ def perform(prison, offender_no, encoded_pdf)
                                     prisoner_number: offender.offender_no,
                                     pom_name: allocation.primary_pom_name,
                                     pom_email: pom.email_address,
-                                    prison_name: PrisonService.name_for(prison),
+                                    prison_name: Prison.find_by!(code: prison).name,
```
starswan, line 18: would be nice to call this prison_code

app/models/allocation.rb
```diff
@@ -137,14 +137,14 @@ def dealloate_offender_after_transfer
...
-    primary_pom = StaffMember.new Prison.new(prison), primary_pom_nomis_id
+    primary_pom = StaffMember.new Prison.find_by!(code: prison), primary_pom_nomis_id
     offender = OffenderService.get_offender(nomis_offender_id)
     mail_params = {
         email: primary_pom.email_address,
         pom_name: primary_pom.first_name,
         offender_name: offender.full_name,
         nomis_offender_id: nomis_offender_id,
-        prison_name: PrisonService.name_for(prison),
+        prison_name: Prison.find_by!(code: prison).name,
```
starswan, line 147: again prison_code would be a nice rename

app/models/allocation.rb
```diff
@@ -137,14 +137,14 @@ def dealloate_offender_after_transfer
 private
 
   def deallocate_offender event:, event_trigger:
-    primary_pom = StaffMember.new Prison.new(prison), primary_pom_nomis_id
+    primary_pom = StaffMember.new Prison.find_by!(code: prison), primary_pom_nomis_id
```
starswan, line 140: This looks wrong - we only need email_address and first_name

app/models/pom_detail.rb
```diff
@@ -11,7 +11,7 @@ class PomDetail < ApplicationRecord
   def allocations
     @allocations ||= begin
       allocations = Allocation.active_pom_allocations(nomis_staff_id, prison_code).pluck(:nomis_offender_id)
-      Prison.new(prison_code).offenders.select { |o| allocations.include? o.offender_no }
+      Prison.find_by!(code: prison_code).offenders.select { |o| allocations.include? o.offender_no }
```
starswan, line 14: pom_detail should belong_to prison

starswan: prison.allocations.where(staff_id: my_staff_id)

app/services/email_service.rb
```diff
@@ -89,7 +89,7 @@ def previous_pom
       }
       return nil if previous.blank?
 
-      StaffMember.new(Prison.new(previous.prison), previous.primary_pom_nomis_id)
+      StaffMember.new(Prison.find_by!(code: previous.prison), previous.primary_pom_nomis_id)
```
starswan, line 92: do we really need a staff member here?

deploy/staging/deployment.yaml
```diff
@@ -27,7 +27,7 @@ spec:
         - name: allocation-manager
           image: 754256621582.dkr.ecr.eu-west-2.amazonaws.com/offender-management/offender-management-allocation-manager:latest
           imagePullPolicy: Always
-          command: ['sh', '-c', 'bundle exec rails db:migrate && bundle exec puma -p 3000 -C ./config/puma_prod.rb --pidfile /tmp/server.pid']
+          command: ['sh', '-c', 'bundle exec rake import_prison','bundle exec rails db:migrate && bundle exec rake import:prison && bundle exec puma -p 3000 -C ./config/puma_prod.rb --pidfile /tmp/server.pid']
```
starswan, line 30: this is not aligned with the others

lib/tasks/fix_data.rake
```diff
@@ -9,7 +9,7 @@ namespace :fix_data do
 
     prisons.each { |prison|
       puts "Attempting to process #{prison}"
-      if PrisonService.name_for(prison).nil?
+      if Prison.find_by!(code: prison).nil?
```
starswan, line 12: find_by (without the !)

app/services/email_service.rb
Katy600: Hi Stephen, I have just been trying to refactor this but unfortunately it gets a bit messy when I try to. Unless I am tackling it wrong it feels better to leave it as it is.

app/models/allocation.rb
```diff
@@ -131,14 +131,14 @@ def dealloate_offender_after_transfer
 private
 
   def deallocate_offender event:, event_trigger:
-    primary_pom = StaffMember.new Prison.new(prison), primary_pom_nomis_id
+    primary_pom = HmppsApi::PrisonApi::PrisonOffenderManagerApi.staff_detail(primary_pom_nomis_id)
     offender = OffenderService.get_offender(nomis_offender_id)
     mail_params = {
         email: primary_pom.email_address,
-        pom_name: primary_pom.first_name,
+        pom_name: primary_pom.first_name.capitalize,
```
starswan, line 138: should be titleize

app/models/prison.rb
```diff
@@ -1,55 +1,30 @@
-# frozen_string_literal: true
-
-class Prison
-  attr_reader :code
+class Prison < ApplicationRecord
+  validates :prison_type, inclusion: ['womens', 'mens_open', 'mens_closed']
```
starswan, line 2: think allow_nil: false would be nice and explicit

db/migrate/20210421134556_create_prisons.rb
```diff
@@ -0,0 +1,11 @@
+class CreatePrisons < ActiveRecord::Migration[6.0]
+  def change
+    create_table :prisons do |t|
+      t.string :prison_type
```
starswan, line 4: non-nullability would be nice here too

db/migrate/20210518134556_create_prisons.rb
```diff
@@ -0,0 +1,11 @@
+class CreatePrisons < ActiveRecord::Migration[6.0]
+  def change
+    create_table :prisons, id: false do |t|
+      t.string :prison_type, null: false
+      t.string :code, primary_key: true, null: false, index: {unique: true}
```
starswan, line 5: You don't need the index or the uniqueness check now that code is the primary key.

starswan (approved):

PR #1708, opened 2021-05-07
Use cache-backed sessions in tests

ollietreend:
This commit adds cache-backed sessions to the RSpec test environment. This makes our tests better reflect the production environment.

This is necessary because it seems that the cookie store behaves very slightly differently to the cache store. When using cookie-backed sessions, journeys which rely on the user's session are able to successfully pass tests but break in production.

Things were breaking in production because we were storing objects in the user's session that didn't serialize/deserialize consistently between the different session stores. The cookie store implicitly transforms objects into a Hash, whereas the cache store rehydrates objects upon retrieval.

The moral of the story is: Don't store objects in the user's session – instead, serialise them into JSON-friendly primitives (i.e. Hash and Array objects) which will be stored and retrieved consistently across the different session stores. And make sure the test environment reflects production as closely as possible.


app/controllers/female_allocations_controller.rb
```diff
@@ -111,4 +114,9 @@ def nomis_offender_id_from_url
   def staff_id
     params.require(:staff_id).to_i
   end
+
+  def save_to_session(key, record)
+    # Store the ActiveRecord object's attributes
+    session[key] = record.attributes
+  end
```
starswan, line 121: I'm not sure about the reason for this method - it doesn't abstract the session concept, so the code still has loading from session and deleting session keys inline.

PR #1709, opened 2021-05-10
MO-648 allocation history bug - dealloc after altering vlo data

starswan:
When viewing case/allocation history, the 'deallocate' view assumed that the previous record was an allocation containing the name of the person involved in the allocation. Since creating case history, this assumption is false. 

This PR fixes the issue by pairing up allocations rather than giving them access to the entire history collection - then the name of the person is in an explicit place rather than being grabbed from the N+1 (N-1 but reversed) record

### DIFF
--- a/app/controllers/allocations_controller.rb
+++ b/app/controllers/allocations_controller.rb
@@ -8,12 +8,12 @@ class AllocationsController < PrisonsApplicationController
   def index
     offender_id = params.require(:prisoner_id)
     @prisoner = offender(offender_id)
-    @previously_allocated_pom_ids =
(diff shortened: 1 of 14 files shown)


app/presenters/allocation_history.rb
```diff
@@ -22,6 +23,30 @@ def created_at
     YAML.load(@version.object_changes).fetch('updated_at').second.getlocal
   end
 
+  def event
+    if @allocation.event == 'reallocate_primary_pom' && @previous_version.prison != @allocation.prison
```
starswan, line 27: carry forward the comment from old view code...

ollietreend (approved):

PR #1710, opened 2021-05-10
MO-671 rework community API URLs to remove dsd.io domains

starswan:


starswan:
Don't merge this until the preprod URL has been ported as well - there's no hurry

config/environments/development.rb
```diff
@@ -65,6 +65,8 @@
   # Use an evented file watcher to asynchronously detect changes in source code,
   # routes, locales, etc. This feature depends on the listen gem.
   config.file_watcher = ActiveSupport::EventedFileUpdateChecker
+
+  config.community_api_host = 'https://community-api.dev.probation.service.justice.gov.uk'
```
ollietreend, line 69: Any reason why you've hardcoded these URLs here?

We haven't done that for any of the other APIs we use. Is there any reason for this to be different?

PR #1715, opened 2021-05-11
MO-679 unify pre allocation summary

starswan:
This PR unifies the 2 summary controllers for male and female prisons, and leaves the SummaryService just for the debugging controller


app/models/prison.rb
```diff
@@ -36,6 +36,48 @@ def unfiltered_offenders
...
+                      else
+                        :unallocated
+                      end
+                    elsif offender.prison_arrival_date.to_date == Time.zone.today
+                      :new_arrival
+                    else
+                      :missing_info
+                    end
+                  end
+                end
```
ollietreend, line 73: Any chance we can DRY this up? It looks like the existence of a Complexity Level is the only difference between these two.

Or are you deliberately keeping these separate?

app/services/allocation_service.rb
```diff
@@ -69,10 +69,12 @@ def self.create_or_update(params)
...
-  def self.active_allocations(nomis_offender_ids, prison)
-    Allocation.active(nomis_offender_ids, prison).index_by { |a|
-      a[:nomis_offender_id]
-    }
+  def self.previously_allocated_poms(nomis_offender_id)
+    allocation = Allocation.find_by(nomis_offender_id: nomis_offender_id)
+
+    return [] if allocation.nil?
+
+    allocation.get_old_versions.map(&:primary_pom_nomis_id)
```
ollietreend, line 77: Looks like this method might duplicate `Allocation#previously_allocated_poms`.

Is it needed? I can't find any uses of this method anywhere, aside from in its own spec file.

spec/controllers/prisoners_controller_spec.rb
```diff
@@ -0,0 +1,718 @@
...
+
+          context "when 'primary_pom_allocated_at' date is nil" do
+            let(:allocated_date) { DateTime.now.utc }
+
+            it "uses 'updated_at' date when 'primary_pom_allocated_at' date is nil" do
+              get :search, params: { prison_id: prison, q: 'Blog' }
+
+              # expect(alloc_offender.allocated_pom_name).to eq('Ward, Alice')
+              # expect(alloc_offender.allocation_date).to be_kind_of(Date)
+            end
```
ollietreend, line 712: There are no assertions in this example

spec/controllers/prisoners_controller_spec.rb
```diff
@@ -0,0 +1,718 @@
...
+
+          context "with a date" do
+            let(:allocated_date) { DateTime.now.utc }
+
+            it 'gets the POM names for allocated offenders' do
+              get :search, params: { prison_id: prison, q: 'Blog' }
+
+              expect(updated_offenders.count).to eq(3)
+              # expect(alloc_offender.allocated_pom_name).to eq('Ward, Alice')
+              # expect(alloc_offender.allocation_date).to be_kind_of(Date)
```
ollietreend, line 694: Unneeded comments

spec/services/allocation_service_spec.rb
```diff
@@ -124,50 +124,33 @@
     end
   end
 
-  describe '#allocations' do
-    it "Can get allocations by prison", vcr: { cassette_name: 'prison_api/allocation_service_get_allocations_by_prison' } do
-      first_offender_id = 'JSHD000NN'
-      second_offender_id = 'SDHH87GD'
-      leeds_prison = 'LEI'
+  describe '#previously_allocated_poms' do
```
ollietreend, line 127: As commented in the implementation: I'm not sure this method is needed or used by anyone? We already have `Allocation#previously_allocated_poms` which is very similar.

ollietreend (commented):
This is looking great 🎉 it's a big improvement over what we had.

app/services/allocation_service.rb
starswan: Thought I had got rid of this - it seems to have come back

spec/controllers/prisoners_controller_spec.rb
starswan: already dealt with

spec/controllers/prisoners_controller_spec.rb
starswan: ditoo

PR #1729, opened 2021-05-20
MO-691 create PrisonTimeline and use it to display CaseHistory records

starswan:
Use new PrisonTimeline model (created from movements)
to create allocation history page, ignoring all 'prison' fields in history records


app/models/hmpps_api/prison_timeline.rb
```diff
@@ -0,0 +1,24 @@
...
+
+    def last_movement
+      @movements.first
+    end
+
+    # obtain a unique visit number for the date specified
+    def prison_visit date
+      @movements.each_with_index.map { |m, index| [m, index] }.detect { |m, _index| m.movement_date <= date }.second
+    end
+
```
Katy600, line 19: My only comment @starswan: I think I might have felt a bit muddled with this name(prison_visit) as it feels very fleeting and more likely feels like someone visiting a prisoner. Wouldn't it be something more substantial than a visit like a prison_stay or prison_resides

app/models/hmpps_api/prison_timeline.rb
```diff
@@ -0,0 +1,23 @@
...
+      @movements = movements.reverse
+    end
+
+    def last_movement
+      @movements.first
+    end
+
+    PrisonEpisode = Struct.new :start_date, :prison_code, keyword_init: true
+
+    # obtain a unique visit number for the date specified
```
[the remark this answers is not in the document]
ollietreend: Doesn't the method do something like this now?

```
# Get the most recent episode which started on or before the given date
```

app/models/hmpps_api/prison_timeline.rb
```diff
@@ -0,0 +1,23 @@
...
+    def prison_episode date
+      relavent_movement = @movements.detect { |m| m.movement_date <= date }
+      PrisonEpisode.new start_date: relavent_movement.movement_date, prison_code: relavent_movement.to_agency
```
ollietreend, line 20: Spelling mistake:

```suggestion
      relevant_movement = @movements.detect { |m| m.movement_date <= date }
      PrisonEpisode.new start_date: relevant_movement.movement_date, prison_code: relevant_movement.to_agency
```

app/views/allocations/history.html.erb
```diff
@@ -2,13 +2,16 @@
 
 <h1 class="govuk-heading-l"><%= @prisoner.full_name %></h1>
 
-<% allocation_list(@history, @early_allocations, @email_histories).each do |prison, case_histories| %>
+<!-- @history is forward-sorted - having converted to a hash key -> list we need to convert back to a list, reverse and also reverse the list itself  -->
+<% @history.group_by { |history| @timeline.prison_episode(history.created_at) }.map { |k, v| [k, v] }.
+    sort_by { |k, v| k.start_date }.
+    reverse.map { |k, v| [k, v.reverse] }.each do |episode, case_histories| %>
```
ollietreend, line 8: I thought we were going to move this into a Presenter to avoid over-complicating the View?

It feels like `@history` should really be a 'case history grouped by prison' Presenter or Enumerator rather than a plain Array – to avoid having to massage the data into the right shape within the View itself.

app/presenters/case_history.rb
```diff
@@ -33,16 +33,24 @@ def event
...
-    @previous_allocation.primary_pom_nomis_id
+  def primary_pom_email
+    HmppsApi::PrisonApi::PrisonOffenderManagerApi.fetch_email_addresses(@allocation.primary_pom_nomis_id).first
   end
 
-  def previous_primary_pom_name
-    @previous_allocation.primary_pom_name
+  def secondary_pom_email
+    HmppsApi::PrisonApi::PrisonOffenderManagerApi.fetch_email_addresses(@allocation.secondary_pom_nomis_id).first
+  end
```
ollietreend, line 42: It might be worth dropping a small comment in to acknowledge that fetching data in the View layer is a 'Bad Thing', but it gets us one step closer to our target data model.

The POM's email address is something which should belong to the data object itself, rather than being injected by the Controller or View. We'll be able to do this once our data model is a bit smarter.

spec/factories/movement.rb
```diff
@@ -26,7 +28,7 @@
 
     # This should be far enough in the past so that the offender isn't considered a 'new arrival' by default
     sequence(:movementDate) do |seq|
-      (Time.zone.today - 1.year - seq.days).to_s
+      (Time.zone.today - 5.years + seq.days).to_s
```
ollietreend, line 31: Additional brackets might make this maths a bit clearer.

Off the top of my head, I have no idea if addition or subtraction is handled first, so seeing a line like this always feels ambiguous to me.

```suggestion
      (Time.zone.today - (5.years + seq.days)).to_s
```

ollietreend (approved):

PR #1730, opened 2021-05-20
MO-680 content change in case history for released and transferred offenders

starswan:


app/views/case_history/allocation/_deallocate_released_offender.html.erb
```diff
@@ -1,4 +1,4 @@
 <%= render 'case_history/case_history',
-           title: 'Prisoner released',
+           title: 'Prisoner unallocated',
            description: '',
```
ollietreend, line 3: Should this say 'Prisoner released' in the description? To be consistent with the transfers.

ollietreend (approved):

PR #1733, opened 2021-05-25
Remove several unused rake tasks (mostly import and backfill of old data)

starswan:


lib/tasks/import_complexity.rake
```diff
@@ -1,23 +0,0 @@
...
-  task :import, [:filename] => [:environment] do |_task, args|
-    Rails.logger = Logger.new(STDOUT)
-
-    if args[:filename].blank?
-      Rails.logger.error('No CSV file specified') if args[:filename].blank?
-    else
-      Rails.logger.info('Beginning import')
-
-      File.open args[:filename] do |f|
-        ComplexityImporter.import f
```
ollietreend, line 17: Let's also lose the associated `ComplexityImporter` – otherwise we'll never remember under what conditions it was used.

ollietreend (approved):

PR #1734, opened 2021-05-26
Mo 685 join up the prison and pom detail models

Katy600:
This PR moves  get_pom_for and get_pom_at methods from the PrisonOffenderManagerService into the Prison model.
Because controllers have @prison in their hand it meant that PrisonOffenderManagerService.get_poms_for could be replaced with @prison.get_poms_for. 

This was not the case in the Service and Job classes but the new database backed Prison model meant that Prison.find could be used here instead - which made the refactor a lot easier as EmailService was quite intwined with AllocationService .

### DIFF
--- a/app/models/prison.rb
+++ b/app/models/prison.rb
@@ -1,8 +1,33 @@
 class Prison < ApplicationRecord
   validates :prison_type, presence: true
   validates :code, :name, presence: true, uniqueness: true
+  has_many :pom_details, dependent: :destroy, foreign_key: :prison_code, inverse_of: :prison
 
   enum prison_type: { womens: 'womens', mens_open: 'mens_open', mens_closed: 'mens_closed' }
+
+  def get_list_of_poms
+    # This API call doesn't do what it says on the tin. It can return duplicate
+    # staff_ids in the situation where someone has more than one role.
+    poms = HmppsApi::PrisonApi::PrisonOffenderManagerApi.list(code).
+      select { |pom| pom.prison_officer? || pom.probation_officer? }.uniq(&:staff_id)
+
+    details = pom_details.where(nomis_staff_id: poms.map(&:staff_id))
+
+    poms.map { |pom| PomWrapper.new(pom, get_pom_detail(details,  pom.staff_id.to_i)) }
+  end
+
+  def get_single_pom(nomis_staff_id)
+    raise ArgumentError, 'Prison#get_single_pom(nil)' if nomis_staff_id.nil?
+
+    poms_list = get_list_of_poms
+    pom = poms_list.find { |p| p.staff_id == nomis_staff_id.to_i }
+    if pom.blank?
+      pom_staff_ids = poms_list.map(&:staff_id)
+      raise StandardError, "Failed to find POM ##{nomis_staff_id} at #{code} - list is #{pom_staff_ids}"
+    end
+
+    pom
+  end
 
   class << self
     def active
@@ -81,6 +106,14 @@ private
     end
   end
 
+  def get_pom_detail(details, nomis_staff_id)
+    details.detect { |pd| pd.nomis_staff_id == nomis_staff_id } ||
+      PomDetail.find_or_create_by!(prison_code: code, nomis_staff_id: nomis_staff_id) do |pom|
+        pom.working_pattern = 0.0
+        pom.status = 'active'
+      end
+  end
+
   class OffenderEnumerator
     include Enumerable
     FETCH_SIZE = 200 # How many records to fetch from nomis at a time
@@ -111,8 +144,6 @@ private
         offenders.each { |offender| yield offender }
       end
     end
-
-  private
 
     def enrich_offenders(offender_list)
       nomis_ids = offender_list.map(&:offender_no)
--- a/spec/services/prison_offender_manager_service_spec.rb
+++ b/spec/services/prison_offender_manager_service_spec.rb
@@ -11,9 +11,11 @@ describe PrisonOffenderManagerService do
   }
 
   context 'when using T3 and VCR' do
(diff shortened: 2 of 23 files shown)


app/models/prison.rb
```diff
@@ -4,6 +4,32 @@ class Prison < ApplicationRecord
 
   enum prison_type: { womens: 'womens', mens_open: 'mens_open', mens_closed: 'mens_closed' }
 
+  # Note - get_poms_for and get_pom_at return different data...
+  def get_poms_for
+    # This API call doesn't do what it says on the tin. It can return duplicate
```
starswan, line 9: Think this comment is no longer true...?

app/models/prison.rb
```diff
@@ -4,6 +4,32 @@ class Prison < ApplicationRecord
 
   enum prison_type: { womens: 'womens', mens_open: 'mens_open', mens_closed: 'mens_closed' }
 
+  # Note - get_poms_for and get_pom_at return different data...
```
starswan, line 7: Think this comment is no longer true about it returning different things

app/models/prison.rb
```diff
@@ -4,6 +4,32 @@ class Prison < ApplicationRecord
 
   enum prison_type: { womens: 'womens', mens_open: 'mens_open', mens_closed: 'mens_closed' }
```
starswan, line 5: was hoping that this ticket would create Prison.has_many :pom_details ?

app/models/prison.rb
```diff
@@ -56,6 +82,18 @@ def summary
...
 private
 
+  def get_pom_detail(pom_details, nomis_staff_id)
+    pom_details.detect { |pd| pd.nomis_staff_id == nomis_staff_id } ||
+      PomDetail.find_or_create_by!(prison_code: code, nomis_staff_id: nomis_staff_id) do |pom|
+        pom.working_pattern = 0.0
+        pom.status = 'active'
+      end
+  end
+
```
starswan, line 92: This method is a bit redundant - we throw (into Sentry) so there's no point logging it. No-one was looking anyway..

spec/services/prison_offender_manager_service_spec.rb
```diff
@@ -11,9 +11,11 @@
   }
 
   context 'when using T3 and VCR' do
+    let(:prison)  { Prison.find('LEI') }
+
     describe '#get_poms_for' do
       subject {
-        described_class.get_poms_for('LEI')
+        prison.get_poms_for
```
starswan, line 18: Think this method should now be called get_poms...?

spec/services/prison_offender_manager_service_spec.rb
```diff
@@ -28,14 +30,14 @@
...
-        pom = described_class.get_pom_at('LEI', staff_id)
+        pom = prison.get_pom_at(staff_id)
         expect(pom).not_to be nil
       end
 
       it "raises an exception when fetching a pom if they are not a POM",
          vcr: { cassette_name: 'prison_api/pom_service_get_pom_none' } do
         expect {
-          described_class.get_pom_at(prison.code, 1234)
+          prison.get_pom_at(1234)
```
starswan, line 40: This behaviour doesn't feel very nice. It's uncatchable and slightly awkward. Maybe we should mirror ActiveRecords interface so that we can decide whether to throw or not?

app/models/prison.rb
```diff
@@ -1,9 +1,34 @@
 class Prison < ApplicationRecord
-  validates :prison_type, presence: true
-  validates :code, :name, presence: true, uniqueness: true
+  validates :prison_type, :code, presence: true
```
starswan, line 2: this looks like a typo? This line should be unaltered - code is already the primary key

app/models/prison.rb
```diff
@@ -87,8 +120,6 @@ def each
       end
     end
 
-  private
-
```
starswan, line 91: STET - we shouldn't be removing private keywords

lib/tasks/backfill_pom_details.rake
```diff
@@ -8,7 +8,7 @@ namespace :backfill do
     Rails.logger = Logger.new(STDOUT)
 
     Prison.all.each do |prison|
-      PrisonOffenderManagerService.get_poms_for(prison.code).map(&:staff_id).each { |staff_id|
+      Prison.find(prison.code).get_list_of_poms.map(&:staff_id).each { |staff_id|
```
starswan, line 11: This is just prison not Prison.find(prison.code)

spec/services/prison_offender_manager_service_spec.rb
```diff
@@ -50,9 +52,12 @@
   context 'without T3 fixtures' do
     before do
       stub_auth_token
+      create(:prison, code: 'WSI')
     end
 
-    describe '#get_poms_for' do
+    describe '#get_list_of_poms' do
+      let(:prison)  { Prison.find('WSI') }
```
starswan, line 59: This would be better as just let!(:prison) { create(:prison) }

spec/services/prison_offender_manager_service_spec.rb
```diff
@@ -134,7 +139,7 @@
 
         it "raises" do
           expect {
-            described_class.get_pom_at('LEI', 1234)
+            described_class.get_single_pom('LEI', 1234)
```
starswan, line 142: this call is wrong (get_single_pom only takes one argument) so yet again the raise check needs to be more explicit.

app/models/prison.rb
Katy600: Hi @starswan I removed it because there was already a private key word. Is it ok to have 2?

starswan (approved):

PR #1737, opened 2021-05-27
MO-628 different offender states

starswan:
This spike has shown that its goal is quite a challenge. This PR is a small piece of work which came out of the spike - but it needs re-thinking to come up with some smaller tasks


spec/models/hmpps_api/offender_spec.rb
```diff
@@ -219,24 +219,28 @@
   end
 
   describe '#over_18?' do
+    subject {
+      build(:offender, dateOfBirth: dob.to_s)
+    }
+
     context 'with a date of birth 50 years ago' do
```
ollietreend, line 226: This context block is wrong – it seems worth fixing whilst we're here:

**~50 years ago~ 18 years ago**

spec/models/hmpps_api/offender_spec.rb
```diff
@@ -219,24 +219,28 @@
...
-      before { subject.date_of_birth = 18.years.ago }
+      let(:dob) { 18.years.ago }
 
       it 'returns true' do
         expect(subject.over_18?).to eq(true)
       end
     end
 
     context 'with a date of birth just under 50 years ago' do
```
ollietreend, line 234: This context block is wrong – it seems worth fixing whilst we're here:

**~just under 50 years ago~ just under 18 years ago**

app/models/staff_member.rb
```diff
@@ -47,7 +47,7 @@ def position
   end
 
   def allocations
-    @allocations ||= fetch_allocations
+    @allocations ||= @prison.allocated.select { |a| a.for_staff_id? @staff_id }
```
ollietreend, line 50: This is lovely

ollietreend (approved):

PR #1744, opened 2021-06-01
Bugfix: don't calculate dates if missing case info

ollietreend:
This fixes a bug in the `RecalculateHandoverDateJob` which was attempting to calculate handover dates for offenders who were missing nDelius case information. This was causing jobs to raise an error, resulting in our job runner (Sidekiq) continually retrying them.

This was affecting offenders who are in prison, but are missing probation data, so fall into the 'Add missing information' category.

In this case, we should skip them as we cannot calculate an accurate handover date for them.

### DIFF
--- a/app/jobs/recalculate_handover_date_job.rb
+++ b/app/jobs/recalculate_handover_date_job.rb
@@ -5,7 +5,7 @@ class RecalculateHandoverDateJob < ApplicationJob
 
   def perform(nomis_offender_id)
     offender = OffenderService.get_offender(nomis_offender_id)
-    if offender&.inside_omic_policy?
+    if offender&.inside_omic_policy? && offender&.has_case_information?
       # Recalculate handover dates, which will trigger a push to the Community API after_save
       recalculate_dates_for(offender)
     end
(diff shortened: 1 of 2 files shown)


app/jobs/recalculate_handover_date_job.rb
```diff
@@ -5,7 +5,7 @@ class RecalculateHandoverDateJob < ApplicationJob
 
   def perform(nomis_offender_id)
     offender = OffenderService.get_offender(nomis_offender_id)
-    if offender&.inside_omic_policy?
+    if offender&.inside_omic_policy? && offender&.has_case_information?
```
starswan, line 8: This should become redundant after MO-628 is merged - prison.offenders changes to only return people 'actively' in the service, so this job can't be executed with an invalid offender

ollietreend: @starswan Does something in MO-628 also check for `has_case_information?` – or is that still needed here?

starswan: This whole ticket becomes redundant. Prison#offenders now returns just the offenders who are 'officially' in the service

ollietreend:
Closing as superseded by MO-628, MO-724, MO-725 and MO-726

PR #1747, opened 2021-06-02
MO-700: Introduce new Offender Event model

ollietreend:
MO-700 calls for a new 'Offender Event' model, which can be used to form a log of events which have happened to offenders. This model uses Rails' Single Table Inheritance to allow for different types of events.

The intention is to use this in the upcoming allocations refactor to record allocation history.

In order to avoid intertwined ticket dependencies, I'm introducing this as an independent thing which can then be used in the allocations refactor.

The spec file `models/events/example_event_spec.rb` provides an example implementation of an Offender Event. Every type of event should extend the main `Events::OffenderEvent` model, as shown in this example spec.


app/models/events/offender_event.rb
```diff
@@ -0,0 +1,70 @@
...
+    # This will enable validation and type casting for metadata fields
+    def self.set_metadata_model(model_class)
+      attribute :metadata, MetadataType.new(model_class)
+      validates :metadata, metadata: true
+    end
+
+    # Return metadata as an OpenStruct (frozen, because these records are read-only)
+    # def metadata
+    #   @metadata ||= OpenStruct.new(self[:metadata]).freeze
+    # end
```
ollietreend, line 63: Remove comments

app/models/events/offender_event.rb
```diff
@@ -0,0 +1,70 @@
+module Events
+  class OffenderEvent < ApplicationRecord
+    # A custom attribute type to represent the metadata field as its own model object (to support validation)
```
starswan, line 3: self.abstract_class = true

spec/models/events/example_event_spec.rb
```diff
@@ -0,0 +1,105 @@
...
+  # Valid metadata
+  let(:metadata) {
+    {
+      name: Faker::Name.name,
+      date_of_birth: Faker::Date.birthday(min_age: 18, max_age: 99),
+      language: 'ruby',
+    }
+  }
+
+  # include_context "core OffenderEvent behaviour"
```
ollietreend, line 53: Remove comment

app/models/events/offender_event.rb
```diff
@@ -0,0 +1,70 @@
+module Events
+  class OffenderEvent < ApplicationRecord
```
ollietreend, line 2: Make this an abstract class

app/models/events/offender_event.rb
```diff
@@ -0,0 +1,65 @@
...
+    # A custom attribute type to represent the metadata field as its own model object (to support validation)
+    class MetadataType < ActiveRecord::Type::Value
```
ollietreend, line 4: Try using gem [jsonb_accessor](https://github.com/madeintandem/jsonb_accessor) instead of all this hand-rolled stuff

app/models/events/offender_event.rb
```diff
@@ -0,0 +1,13 @@
+module Events
+  class OffenderEvent < ApplicationRecord
+    validates :nomis_offender_id, :happened_at, :type, presence: true
+    enum triggered_by: { user: 'user', system: 'system' }, _prefix: true
+    validates :triggered_by, presence: true, inclusion: { in: triggered_bies.keys }
```
starswan, line 5: Don't think you need line 5 - it is implemented by the enum

starswan (approved):

PR #1748, opened 2021-06-03
MO-628 sort by pom_responsibility

starswan:
As a result of the previous PR for MO-628, it appears that sorting by pom_responsibility was broken. This PR puts that feature back, adding a test as well


spec/features/pom_caseload_feature_spec.rb
```diff
@@ -167,6 +167,12 @@
       end
     end
 
+    it 'can be sorted by responsibility', :js do
+      sleep 10
+      click_link 'Role'
+      sleep 50
+    end
```
ollietreend, line 174: These long sleeps are going to have a real impact on the length of our test runs. This alone will add a whole minute ☹️

Do these sleeps really need to be this long?

ollietreend (approved):

PR #1750, opened 2021-06-04
MO-690 Part III - move responsibility

starswan:


app/models/hmpps_api/offender.rb
```diff
@@ -66,7 +66,7 @@ def load_case_information(record)
       @case_information = record
 
       # This is separate from @case_information so the rest of this class doesn't need to know about the Active Record association
-      @responsibility_override = record.responsibility
+      @responsibility_override = record.offender.responsibility
```
ollietreend, line 69: As with previous PRs: this might be nicer if we `load_offender_record` rather than `load_case_information`, since that's the new 'top level' record representing an offender within the service.

Specifically loading CaseInformation is almost arbitrary at this point...

ollietreend (approved):

PR #1751, opened 2021-06-04
MO-690 Part II - move email histories

starswan:


app/jobs/recalculate_handover_date_job.rb
```diff
@@ -29,11 +29,11 @@ def recalculate_dates_for(offender)
       offender.ldu_email_address.present? &&
       offender.allocated_com_name.blank?
       # need to chase if we haven't chased recently
-      last_chaser = case_info.email_histories.where(event: EmailHistory::IMMEDIATE_COMMUNITY_ALLOCATION).last
+      last_chaser = case_info.offender.email_histories.where(event: EmailHistory::IMMEDIATE_COMMUNITY_ALLOCATION).last
```
ollietreend, line 32: Same thing in this PR – it would be nice if we had a merged offender object (eventually) where we could do something like:

```ruby
if offender.has_case_information?
  last_chaser = offender.email_histories.where(event: EmailHistory::IMMEDIATE_COMMUNITY_ALLOCATION).last
end
```

- so we don't have to go 'through' the Case Information record, which isn't at the 'root' of our data model any more, and
- so we have that ActiveRecord-style object which combines all data sources related to an offender

ollietreend (approved):

PR #1753, opened 2021-06-04
MO-690 Part V - move vlo data Part I - add offender id and allow back-fill

starswan:


db/migrate/20210604083336_add_offender_id_to_vlo.rb
```diff
@@ -0,0 +1,7 @@
+class AddOffenderIdToVlo < ActiveRecord::Migration[6.0]
+  def change
+    change_table :victim_liaison_officers do |vlo_table|
+      vlo_table.string :nomis_offender_id, limit: 7, null: false, index: true
```
ollietreend, line 4: `null: false` assumes that every VLO record has an existing Offender record – which I don't think we can assume to be true, given the current state of our database.

lib/tasks/backfill_vlo_offenders.rake
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+require 'rake'
+
+namespace :backfill do
+  desc 'Back-fill offender ids for VLO records'
+  task vlo: :environment do
+    VictimLiaisonOfficer.includes(case_information: :offender).where(nomis_offender_id: nil).find_each do |vlo|
+      vlo.update!(offender: vlo.case_information.offender)
```
ollietreend, line 9: This `update!` will fail if `vlo.case_information.offender` is `nil`.

ollietreend (commented):
This PR assumes that every `CaseInformation` record has an `Offender` record.

However in reality, we know this isn't the case. There are numerous 'orphaned' CaseInformation records waiting to be cleaned up which will cause this to fail.

db/migrate/20210604083336_add_offender_id_to_vlo.rb
ollietreend: @starswan can I just double check this with you? I still think it's a problem...

Surely this migration will fail in production, since `null: false` is specified but without a `default` value.

We can't run the migration rake task until the columns has been added... a bit of a chicken & egg situation?

ollietreend (approved):

PR #1760, opened 2021-06-07
MO-702: Rename Allocation to AllocationHistory

ollietreend:
Ticket: MO-702

This pull request renames the Allocation model to AllocationHistory, and its associated database table from "allocations" to "allocation_history". It will enable us to begin implementing our new Allocation data model by avoiding a name clash.

This refactor was performed by a bash script – making it repeatable and feel 'safer'. This approach was inspired by the following blog post:
https://corgibytes.com/blog/2016/07/21/rename_rails_models/

You can see the bash script I used for this refactor here:
https://gist.github.com/ollietreend/769fb8b43bbfa77aea94fbb871f8b9a6


db/migrate/20210603122414_rename_allocation_to_allocation_history.rb
```diff
@@ -0,0 +1,16 @@
+class RenameAllocationToAllocationHistory < ActiveRecord::Migration[6.0]
+  def change
+    rename_table :allocations, :allocation_history
```
starswan, line 3: what was wrong with the default allocation_histories table name?

ollietreend: No strong reason: I didn't think 'histories' made sense in this context, so thought an "allocation_history" table would be more readable.

Although I've just seen that we have an "email_histories" table, so maybe it's better to be consistent and stick to the Rails convention.

I'm happy to change this back if you think it's worth doing?

starswan (approved):

PR #1761, opened 2021-06-07
MO-690 Part VI - create parole record,  remove field from CaseInformation

starswan:


spec/models/parole_record_spec.rb
```diff
@@ -0,0 +1,7 @@
+require 'rails_helper'
+
+RSpec.describe ParoleRecord, type: :model do
+  it 'always elongs to an offender' do
```
ollietreend, line 4:
```suggestion
  it 'always belongs to an offender' do
```

ollietreend (approved):
Approved – although we may end up renaming the `ParoleRecord` model if it turns out we get _two different types_ of record from PPUD...

I don't think we should let that hold this PR up, though.

PR #1764, opened 2021-06-09
MO-731 fix editing of POM status from active -> inactive and back again

starswan:


spec/features/edit_a_pom_spec.rb
```diff
@@ -44,22 +45,36 @@
...
+      click_link 'Moic Integration-Tests'
+
+      click_link "Edit profile"
 
-    find('label[for=working_pattern-5]').click
-    find('label[for=status-1]').click
+      expect(page).to have_css('h1', text: 'Edit profile')
 
-    click_button('Save')
+      find('label[for=status-1]').click
```
Katy600, line 67: Is this label a radio button for 'active status'? I t is quite easy to guess but it could be more explicit too.

ollietreend: What Katy said ^^ +1

ollietreend (approved):

spec/features/edit_a_pom_spec.rb
```diff
@@ -44,22 +45,36 @@
...
+  describe "making an inactive POM active", vcr: { cassette_name: 'prison_api/edit_poms_activate_pom_feature' } do
+    let(:prison) { Prison.find('LEI') }
+    let(:other_prison) { create(:prison) }
 
-    click_link "Edit profile"
+    before do
+      # This is the ID for MOIC Integration-tests
+      create(:pom_detail, :inactive, prison: other_prison, nomis_staff_id: 485_758)
+      create(:pom_detail, :inactive, prison: prison, nomis_staff_id: 485_758)
+    end
```
Katy600, line 56: I can't see where other_prison is being used @starswan. What is its purpose? I think that might be just a question for me. :)

Katy600 (approved):
I have just added a couple but I don't think they are very significant comments so I have approved it. :)

spec/features/edit_a_pom_spec.rb
```diff
@@ -44,22 +45,36 @@
...
-    click_button('Save')
+      find('label[for=status-1]').click
+      find('label[for=part-time-conditional-1]').click
+      find('label[for=working_pattern-5]').click
 
-    expect(page).to have_content('0.5')
-    expect(page).to have_content('Active')
+      click_button('Save')
+
+      expect(prison.pom_details.find_by(nomis_staff_id: 485_758).status).to eq('active')
```
ollietreend, line 73: It might be nice to check that the same POM in `other_prison` is still marked as "inactive" – to prove that we haven't just activated them globally.

PR #1767, opened 2021-06-11
MO-690 Prequel - move case_information outside of API offender

starswan:
This PR is in response to conversations we have been having, and the fact that we now have 2 offender objects (well we always did, its just that we called it case_information) which is really confusing. This unifies the 2 objects in a an ActiveRecord first pattern, so that data is accessible for entities that need it.

### DIFF
--- a/app/helpers/offender_helper.rb
+++ b/app/helpers/offender_helper.rb
@@ -88,4 +88,8 @@ module OffenderHelper
     offenders.count { |a| a.tier == 'N/A' }
   end
   # :nocov:
+
+  def probation_field offender, field
+    offender.public_send field if offender.probation_record.present?
+  end
 end
--- a/app/models/case_information.rb
+++ b/app/models/case_information.rb
@@ -6,9 +6,10 @@ class CaseInformation < ApplicationRecord
   NPS = 'NPS'
   CRC = 'CRC'
 
-  belongs_to :offender, primary_key: :nomis_offender_id, foreign_key: :nomis_offender_id, inverse_of: :case_information
+  belongs_to :offender, foreign_key: :nomis_offender_id, inverse_of: :case_information
 
-  belongs_to :local_delivery_unit, optional: true
+  belongs_to :local_delivery_unit, -> { enabled }, optional: true, inverse_of: :case_information
+  delegate :name, :email_address, to: :local_delivery_unit, prefix: :ldu, allow_nil: true
 
   has_many :early_allocations,
            -> { order(created_at: :asc) },
@@ -43,15 +44,16 @@ class CaseInformation < ApplicationRecord
            inverse_of: :case_information,
            dependent: :destroy
 
-  def nps?
+  def nps_case?
     case_allocation == NPS
   end
 
-  # Only return the LDU if it's enabled
-  def ldu
-    if local_delivery_unit&.enabled?
-      local_delivery_unit
-    end
+  def welsh_offender
+    probation_service == 'Wales'
+  end
+
+  def delius_matched?
+    manual_entry == false
   end
 
   validates :manual_entry, inclusion: { in: [true, false], allow_nil: false }
--- a/app/models/mpc_offender.rb
+++ b/app/models/mpc_offender.rb
@@ -1,0 +1,162 @@
+# frozen_string_literal: true
+
+class MpcOffender
+  delegate :get_image,
+           :convicted?, :recalled?, :immigration_case?, :indeterminate_sentence?,
+           :sentenced?, :over_18?, :describe_sentence, :civil_sentence?,
+           :sentence_start_date, :conditional_release_date, :automatic_release_date, :parole_eligibility_date,
+           :tariff_date, :post_recall_release_date, :licence_expiry_date,
+           :home_detention_curfew_actual_date, :home_detention_curfew_eligibility_date,
+           :prison_arrival_date, :earliest_release_date, :latest_temp_movement_date, :release_date,
+           :date_of_birth, :main_offence, :awaiting_allocation_for, :cell_location,
+           :category_label, :complexity_level, :category_code, :category_active_since,
+           :first_name, :last_name, :full_name_ordered, :full_name,
+           :inside_omic_policy?, :offender_no, :prison_id, to: :@prison_record
+
+  delegate :crn, :case_allocation, :parole_review_date, :manual_entry?, :nps_case?,
+           :tier, :victim_liaison_officers, :early_allocations, :delius_matched?,
+           :mappa_level, :welsh_offender, to: :probation_record
+
+  # These fields make sense to be nil when the probation record is nil - the others dont
+  delegate :ldu_email_address, :team_name, :ldu_name, to: :probation_record, allow_nil: true
+
+  attr_reader :probation_record, :prison
+
+  def initialize(prison:, offender:, prison_record:)
+    @prison = prison
+    @offender = offender
+    @prison_record = prison_record
+    @probation_record = offender.case_information
+  end
+
+  # TODO - view method in model needs to be removed
+  def case_owner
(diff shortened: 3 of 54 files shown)


app/models/mpc_offender.rb
```diff
@@ -0,0 +1,156 @@
...
+           :victim_liaison_officers, :early_allocations,
+           :mappa_level, :welsh_offender, to: :probation_record
+
+  attr_reader :probation_record
+
+  def initialize(offender, api_offender)
+    @offender = offender
+    @prison_record = api_offender
+    @probation_record = offender.case_information
+  end
```
ollietreend, line 29: We also talked about passing the offender's current `Prison` into the initialiser, and making it available at `attr_reader @prison`.

Would be lovely to have that...

app/views/prisoners/show.html.erb
```diff
@@ -71,4 +71,4 @@
 <%= render partial: 'shared/offence_info', locals: { editable_prd: true}  %>
 <%= render 'prison_allocation' %>
 <%= render 'community_information' %>
-<%= render 'shared/vlo_information' %>
+<%= render 'shared/vlo_information' if @prisoner.probation_record.present? %>
```
ollietreend, line 74: Why can't VLO information be shown if the Probation Record is missing?

I thought VLOs were standalone.

app/models/mpc_offender.rb
ollietreend: Also:

This could just be personal preference, but I quite like the idea of accepting `probation_record` as its own initialiser argument. So the OffenderService knows how to find the probation record, rather than the initialiser knowing that it exists 'inside' the Offender record.

One subtle advantage of this is that it should simplify the way we structure tests.

### Example

For example: `spec/helpers/offender_helper_spec.rb`

Where we currently have:

```ruby
let(:offender) {
  build(:mpc_offender, prison_record: api_offender, offender: build(:case_information).offender)
}
```

We could change this to:

```ruby
let(:offender) {
    build(:mpc_offender,
        prison_record: api_offender,
        probation_record: build(:case_information),
        offender: build(:offender))
}
```

So we don't have to build a `CaseInformation` just to extract the `Offender` record (only for the initialiser to extract the `CaseInformation` from the `Offender` again...)

The `CaseInformation` records could almost be cut loose from the main `Offender` data model... and exist 'on the side' of our internal data model, only being associated with the `LocalDeliveryUnit`.

app/helpers/offender_helper.rb
```diff
@@ -86,4 +86,8 @@ def tier_d_case_count offenders
   def no_tier_case_count offenders
     offenders.count { |a| a.tier == 'N/A' }
   end
+
+  def probation_field offender, field
+    offender.public_send field if offender.probation_record.present?
+  end
```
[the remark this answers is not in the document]
starswan: Yes - but that is mostly the wrong question. This is called probation_field for a reason - its a field from the probation record. Nothing to do with whether someone is allocatable or not - that's probably just used in splitting the buckets in the summary section

app/models/case_information.rb
```diff
@@ -59,6 +59,18 @@ def latest_early_allocation
...
+    probation_service == 'Wales'
+  end
+
+  def ldu_name
+    local_delivery_unit&.name
+  end
+
+  def ldu_email_address
+    local_delivery_unit&.email_address
+  end
```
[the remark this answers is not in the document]
starswan: like the latter - included

app/models/mpc_offender.rb
```diff
@@ -0,0 +1,156 @@
+# frozen_string_literal: true
+
+class MpcOffender
+  delegate :tier, to: :@probation_record
```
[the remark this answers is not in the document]
starswan: oops - typo. Thanks

app/models/offender.rb
```diff
@@ -4,5 +4,5 @@ class Offender < ApplicationRecord
   # NOMIS offender IDs must be of the form <letter><4 numbers><2 letters> (all uppercase)
   validates :nomis_offender_id, format: { with: /\A[A-Z][0-9]{4}[A-Z]{2}\z/ }
 
-  has_one :case_information, foreign_key: :nomis_offender_id, primary_key: :nomis_offender_id, inverse_of: :offender, dependent: :destroy
+  has_one :case_information, foreign_key: :nomis_offender_id, inverse_of: :offender, dependent: :destroy
```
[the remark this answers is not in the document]
starswan: has_one is optional by default - it's only belongs_to where it matters

app/services/offender_service.rb
```diff
@@ -1,27 +1,77 @@
...
+    def get_offender(offender_no)
+      # use find_or_create_by here for performance, but might still have be a small race condition
+      offender = Offender.find_or_create_by(nomis_offender_id: offender_no)
+      if offender
+        api_offender = HmppsApi::PrisonApi::OffenderApi.get_offender(offender_no)
+        if api_offender
+          api_offender.load_main_offence
+          MpcOffender.new offender, api_offender
+        end
+      end
```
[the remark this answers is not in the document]
starswan: It's because we don't have a valid offender number - so find_or_create_by can fail if looking for an invalid id which isn't there.

app/views/prisoners/show.html.erb
starswan: They are attached (for now) - this will hopefully go once the VLOs get moved

app/models/prison.rb
```diff
@@ -79,9 +79,9 @@ def summary
       summary = unfiltered_offenders.group_by do |offender|
         if offender.inside_omic_policy?
           allocatable = if womens?
-                          offender.has_case_information? && offender.complexity_level.present?
+                          offender.probation_record.present? && offender.complexity_level.present?
                         else
-                          offender.has_case_information?
+                          offender.probation_record.present?
```
ollietreend, line 84: At one point I think we talked about the Offender object itself _telling us_ what it is...

e.g. `offender.allocatable?`

Rather than making it the responsibility of the Prison model to decide when an Offender is in an 'allocatable' state.

app/services/offender_service.rb
```diff
@@ -1,27 +1,80 @@
...
+
+          enrich_offenders(offenders).each { |offender| yield offender }
+        end
+      end
+
+      def enrich_offenders(offender_list)
+        nomis_ids = offender_list.map(&:offender_no)
+        offenders = Offender.
+          includes(case_information: [:responsibility, :early_allocations, :local_delivery_unit]).
+          where(nomis_offender_id: nomis_ids)
```
ollietreend, line 52: ### Performance optimisation (todo?)

Presumably we'll also want to be eager loading things like `ParoleRecord` (for PRDs) once that PR is in...

For example, PRD is likely to be needed (I think?) when calculating handover dates/responsibility. And that happens when filtering offenders (e.g. approaching handover).

app/views/prisoners/show.html.erb
ollietreend: ### Todo

Once the VLO PR is merged in, this case change...

app/models/mpc_offender.rb
```diff
@@ -0,0 +1,161 @@
+# frozen_string_literal: true
+
+class MpcOffender
```
ollietreend, line 3: ### Naming

It's a shame the `Offender` model name is already taken by our ActiveRecord model – since it seems like it would be a more suitable name for this model.

But I don't have a strong opinion about how better to name either of these models – so it doesn't seem worth changing!

ollietreend (commented):
This PR looks glorious 🤩 really nice work @starswan.

I've not 'approved' yet, simply because I think this PR will need to change after the other MO-690 PRs have been merged in...

ollietreend (approved):

PR #1769, opened 2021-06-11
Refactor how factories generate NOMIS Offender IDs

ollietreend:
Several FactoryBot factories need to generate NOMIS Offender IDs.

Rather than generating them all as independent sequences, each with slightly different letters to avoid clashes, this commit refactors them to use one single global sequence which can be used in any factory.

That way, the logic for generating a realistic NOMIS Offender ID only exists in one place.

### Using it in factories

Most factories shouldn't need to generate a NOMIS Offender ID now that we have Offender records.

However, if you do need one, you have a couple of options.

1. If the field name is `nomis_offender_id`, then you can do something like this:
    
    ```ruby
    FactoryBot.define do
      factory :example do
        nomis_offender_id
      end
    end
    ```
    
    This will create an object where the `nomis_offender_id` attribute will be set.

2. If the field name is different (e.g. `offenderNo`), then you can use the `generate` method:
    
    ```ruby
    FactoryBot.define do
      factory :example do
        offenderNo { generate :nomis_offender_id }
      end
    end
    ```

You can [learn more about global sequences in the FactoryBot documentation](https://github.com/thoughtbot/factory_bot/blob/master/GETTING_STARTED.md#global-sequences).

### DIFF
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -123,3 +123,17 @@ end
 # The en-GB locale gives us UK counties (used by the LocalDeliveryUnit factory)
 # See all the things we gain here: https://github.com/faker-ruby/faker/blob/master/lib/locales/en-GB.yml
 Faker::Config.locale = 'en-GB'
+
+# Define a global sequence for generating NOMIS Offender IDs
+# NOMIS offender IDs follow the format: <letter><4 numbers><2 letters> (all uppercase)
+FactoryBot.define do
+  sequence :nomis_offender_id do |seq|
+    index = seq - 1 # because seq starts at 1, not 0
+    number = "%04d" % (index / 26) # zero-pad to 4 digits, e.g. 0001
+    letter = ('A'..'Z').to_a[index % 26]
+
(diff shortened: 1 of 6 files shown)


spec/rails_helper.rb
```diff
@@ -123,3 +123,14 @@
...
 Faker::Config.locale = 'en-GB'
+
+# Define a global sequence for generating NOMIS Offender IDs
+# NOMIS offender IDs follow the format: <letter><4 numbers><2 letters> (all uppercase)
+FactoryBot.define do
+  sequence :nomis_offender_id do |seq|
+    index = seq - 1 # because seq starts at 1, not 0
+    number = "%04d" % (index / 26) # zero-pad to 4 digits, e.g. 0001
+    letter = ('A'..'Z').to_a[index % 26]
+    "A#{number}A#{letter}"
```
starswan, line 134: I deliberately used 'T' here tp try to indicate 'test' - IIRC A is used by real NOMIS (I think production) so maybe another letter might be prudent?

starswan (approved):

PR #1771, opened 2021-06-16
MO-690 Test spec changes

starswan:
As a result of MO-690, a number of spec changes were identified. This PR pulls those changes out independently


Katy600 (approved):
No comment :)

PR #1773, opened 2021-06-17
Ignore the early allocations that are before the current sentence start date

Katy600:
Early allocations are not held onto by the offender but are attached to a specific sentence, so an offender returning on a new sentence should not have their previous early allocation surfaced. 

This PR fixes that.

### DIFF
--- a/app/helpers/badge_helper.rb
+++ b/app/helpers/badge_helper.rb
@@ -23,17 +23,13 @@ module BadgeHelper
     offender.indeterminate_sentence? && offender.parole_review_date.present?
   end
 
-  def early_allocation_notes?(offender, early_allocations)
-    if early_allocations.any?
-      !offender.within_early_allocation_window? || early_allocations.last.community_decision == false
+  def early_allocation_notes?(early_allocations)
+    if early_allocations.present?
+      !early_allocations.last.created_within_referral_window? || early_allocations.last.community_decision == false
     end
   end
 
   def early_allocation_active?(early_allocations)
-    early_allocations.any? && early_allocations.last.awaiting_community_decision?
-  end
-
-  def early_allocation_approved?(early_allocations)
-    early_allocations.any? && early_allocations.last.created_within_referral_window? && early_allocations.last.community_decision_eligible_or_automatically_eligible?
+    early_allocations.present? && early_allocations.last.awaiting_community_decision?
   end
 end
--- a/app/models/hmpps_api/offender.rb
+++ b/app/models/hmpps_api/offender.rb
@@ -108,7 +108,7 @@ module HmppsApi
     def early_allocation?
       return false if @case_information.blank?
 
-      early_allocation.present?
+      latest_early_allocation.present?
     end
 
     def needs_early_allocation_notify?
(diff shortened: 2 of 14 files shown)


app/helpers/badge_helper.rb
```diff
@@ -23,17 +23,21 @@ def parole_review_date?(offender)
...
+    allocations = current_sentence_allocations(sentence, early_allocations)
+    allocations.any? && allocations.last.awaiting_community_decision?
   end
 
-  def early_allocation_approved?(early_allocations)
-    early_allocations.any? && early_allocations.last.created_within_referral_window? && early_allocations.last.community_decision_eligible_or_automatically_eligible?
+private
+
+  def current_sentence_allocations(sentence, early_allocations)
+    early_allocations.select { |early_allocation| early_allocation.assessment_date.after?(sentence.sentence_start_date) }
```
starswan, line 41: This is duplicated from the model - this one needs to be removed

starswan, line 41: This filtering is handled by the model, so we should be using the model method and deleting this

app/models/hmpps_api/offender.rb
```diff
@@ -257,6 +257,15 @@ def inside_omic_policy?
...
     end
 
+    def latest_early_allocation
+      allocation = current_sentence_allocations(sentence, @case_information&.early_allocations || [])&.last
+      allocation if allocation.present? && allocation.created_within_referral_window? && (allocation.eligible? || allocation.community_decision?)
+    end
+
+    def current_sentence_allocations(sentence, early_allocations)
+      early_allocations.select { |early_allocation| early_allocation.assessment_date.after?(sentence.sentence_start_date) }
+    end
```
starswan, line 267: Think we can be smarter than this: something like

def early_allocations
   @case.informatiom.early_allocations.where(created_at >= ?, sentence_start_date)
end

then the controller should feed this rather than the raw data to the view - all the offender parameters go away again

starswan: It would also be good to add a check to the early_allocations_controller :index action to ensure that it picks up all the allocations not just the pre-sentence ones - and maybe same for allocation history

app/views/shared/_badges.html.erb
```diff
@@ -1,11 +1,13 @@
 <span id="prisoner-case-type" class="moj-badge moj-badge--<%=badge_colour(offender) %>"><%=badge_text(offender)%></span>
 <%= render('shared/badges/recall_badge') if offender.recalled? %>
 <%= render('shared/badges/parole_badge') if badge_for_parole?(offender) %>
-<% if early_allocation_approved?(early_allocations) %>
+
+<% if offender.latest_early_allocation.present? %>
```
Katy600, line 5: This should be early_allocation?

app/views/shared/_badges.html.erb
```diff
@@ -1,11 +1,13 @@
...
   <%=  render('shared/badges/early_allocation_approved_badge') %>
```
starswan, line 6: Should just be offender.early_allocation?

starswan (approved):

PR #1778, opened 2021-06-21
MO-647 brand new api

starswan:
Create brand-new API endpoint to expose boolean early allocation status. Add an SNS event so that clients don't have to poll the endpoint. This PR also includes a significant refactor of the early allocation calculation and display code (and fixes one bug in the badge display handling around auto-eligibility) so that the state is now calculated once in the MpcOffender model and displayed by the view code, rather than the view code trying to calculate the state in a myriad of different ways


deploy/preprod/cron-early-allocation-events.yaml
```diff
@@ -0,0 +1,53 @@
+apiVersion: batch/v1beta1
+kind: CronJob
+metadata:
+  name: early-allocation-events
+spec:
+  schedule: "00 23 * * *"
```
ollietreend, line 6: It would be preferable to run this once a day _in the early morning_ rather than _at the end of the day_ – e.g. `00 02 * * *` rather than `00 23 * * *`.

This will help us catch when someone stops being eligible due to their sentence date having passed. Otherwise we'll effectively be > 23 hours late to emit the event.

deploy/production/cron-early-allocation-events.yaml
```diff
@@ -0,0 +1,53 @@
+apiVersion: batch/v1beta1
+kind: CronJob
+metadata:
+  name: early-allocation-events
+spec:
+  schedule: "00 23 * * *"
```
ollietreend, line 6: Ditto

deploy/production/cron-early-allocation-events.yaml
```diff
@@ -0,0 +1,53 @@
...
+  concurrencyPolicy: Forbid
+  successfulJobsHistoryLimit: 3
+  failedJobsHistoryLimit: 2
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          containers:
+            - name: community-api-import
```
ollietreend, line 15: Ditto

deploy/staging/cron-early-allocation-events.yaml
```diff
@@ -0,0 +1,53 @@
+apiVersion: batch/v1beta1
+kind: CronJob
+metadata:
+  name: early-allocation-events
+spec:
+  schedule: "00 23 * * *"
```
ollietreend, line 6: Ditto

deploy/staging/cron-early-allocation-events.yaml
```diff
@@ -0,0 +1,53 @@
...
+  concurrencyPolicy: Forbid
+  successfulJobsHistoryLimit: 3
+  failedJobsHistoryLimit: 2
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          containers:
+            - name: community-api-import
```
ollietreend, line 15: Ditto

lib/tasks/trigger_ea_events.rake
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+require 'rake'
+
+namespace :trigger do
+  desc 'Send follow-up emails LDUs two weeks before handover if COM still not assigned'
```
ollietreend, line 6: This description is wrong (although I can tell which rake task you forked this from 😆)

ollietreend (approved):

PR #1780, opened 2021-06-21
MO-495 change date format to RFC 822 format (01 Sep 2034)

starswan:


app/helpers/application_helper.rb
```diff
@@ -2,15 +2,11 @@
...
-
-  def format_date_string(date_string)
-    return '' if date_string.empty?
-
-    Date.parse(date_string).strftime('%d/%m/%Y')
+    if date_obj
+      date_obj.to_s(:rfc822)
+    else
+      replacement
+    end
```
Katy600, line 9: What is replacement for? Is that if a string date is passed?

Katy600 (approved):
I have only left a question about replacement as it is not clear to me when that is used. :)

app/helpers/application_helper.rb
starswan: It's the replacement string for nil (as shown in the if statement)

PR #1781, opened 2021-06-21
MO-736 new allocation email is a link to the case not the caseload

starswan:


Katy600 (approved):
I have been trying to understand your initial refactor. Is the main purpose of replacing instance with send_email and send_secondary_email to encapsulates things? So you are no longer having to have instance variables everywhere, as these are now passed to the send_email & secondary_email methods parameters? And then you have just made a few extra checks in send_deallocation_email. This is also the case for deliver_new_allocation_email? So is it mostly about encapsulation?

PR #1782, opened 2021-06-21
Add retries on 500 errors, and remove POST retries apart from Prison API and Prison Search

starswan:


Katy600 (approved):
I have no comment. I just tried to make sure I understood it!

PR #1794, opened 2021-06-30
MO-765 crash on encountering an old allocation due to missing a release event

starswan:
prevent crash when encountering an allocation from a previous prison or one from a POM who is no longer in the POMs list for the prison

### DIFF
--- a/app/controllers/allocation_staff_controller.rb
+++ b/app/controllers/allocation_staff_controller.rb
@@ -7,7 +7,7 @@ class AllocationStaffController < PrisonsApplicationController
 
   def index
     @case_info = Offender.find_by!(nomis_offender_id: prisoner_id_from_url).case_information
-    @allocation = AllocationHistory.find_by nomis_offender_id: prisoner_id_from_url
+    @allocation = AllocationHistory.find_by prison: @prison.code, nomis_offender_id: prisoner_id_from_url
     previous_pom_ids = if @allocation
                          @allocation.previously_allocated_poms
                        else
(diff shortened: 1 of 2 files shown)


app/controllers/allocation_staff_controller.rb
```diff
@@ -7,15 +7,15 @@ class AllocationStaffController < PrisonsApplicationController
...
+    @allocation = AllocationHistory.find_by prison: @prison.code, nomis_offender_id: prisoner_id_from_url
     previous_pom_ids = if @allocation
                          @allocation.previously_allocated_poms
                        else
                          []
                        end
     poms = @prison.get_list_of_poms.index_by(&:staff_id)
     @previous_poms = previous_pom_ids.map { |staff_id| poms[staff_id] }.compact
-    @current_pom = @prison.get_single_pom(@allocation.primary_pom_nomis_id) if @allocation&.primary_pom_nomis_id
+    @current_pom = poms[@allocation.primary_pom_nomis_id] if @allocation&.primary_pom_nomis_id
```
Katy600, line 18: I am just trying to make sure I understand this. `@allocation&.primary_pom_nomis_id` checks if the allocation has a primary pom and if it does `poms[@allocation.primary_pom_nomis]` returns the current pom. But if it doesn't have a primary_pom_nomis_id when current_pom get assigned nil?

spec/controllers/allocation_staff_controller_spec.rb
```diff
@@ -28,9 +28,22 @@
...
     describe '#index' do
-      context 'with previous allocations for this POM' do
-        let(:alice) { poms.first }
+      let(:alice) { poms.first }
+
+      context 'with an allocation from a previous prison' do
+        before do
+          create(:case_information, offender: build(:offender, nomis_offender_id: offender_no))
+          create(:allocation_history, prison: build(:prison).code, nomis_offender_id: offender_no)
+        end
```
Katy600, line 37: This `prison: build(:prison).code` inside allocation history, is a different prison to the `prison_id: prison_code` passed to the index action? Which is how this test is able to reproduce the bug?

Katy600 (approved):
I just have a few questions for myself but otherwise I have approved it :)

spec/controllers/allocation_staff_controller_spec.rb
starswan: Yes

PR #1801, opened 2021-07-01
MO-774 Add CaseInformation to active admin in HEROKU-land only

starswan:


Katy600 (approved):
I don't really know what this is doing exactly to be honest

PR #1802, opened 2021-07-01
MO-572 Add copy with counts and order of the first 4 cards

Katy600:
- Change the card headings that say ‘make new allocations’ and ‘update case information’ (this was already done so no change needed)

- Add copy inside the card with the counts in there. 

- Order of the first 4 cards to better reflect the order in the HOMd allocations sub nav. 

Screen Shot 2021-03-31 at 09 45 30


app/views/dashboard/index.html.erb
```diff
@@ -26,25 +26,25 @@
...
     </div>
   </div>
   <div class="govuk-grid-column-one-third">
     <div class="wrapper">
       <h1 class="govuk-heading-s">
-        <%= link_to "Make new allocations", unallocated_prison_prisoners_path(@prison.code), class:"govuk-link" %>
+        <%= link_to "Update case information", missing_information_prison_prisoners_path(@prison.code), class:"govuk-link" %>
       </h1>
-      <p class="govuk-body">Prisoners who need to be allocated to a Prison Offender Manager.</p>
+      <p class="govuk-body">You need to add missing detail to <%= @caseloads.count %> cases so that they can be allocated to a POM .</p>
```
starswan, line 39: I think we need a test here. You're using the same count for both these sections which doesn't look right, and I'm really not sure what is in the variable 'caseloads' - seems a strange name

starswan (dismissed):

starswan (approved):
Only comment would be that content_options isn't a great method name - but I can't think of another one to replace it...?

PR #1806, opened 2021-07-06
Refactor the dashboard controller to separate out the edit view

Katy600:
This PR originally intended to remove the dashboard controller altogether but unfortunately the new rails routes generated prison_path which required an :id as apposed to a :prison_id. This meant that inheriting from PrisonsApplicationController was difficult as it expected `:prison_id`

``` 
def active_prison_id
    params[:prison_id]
end
```

As a result dashboard controller has remained. I had hoped to change the index action in dashboard controller to a show action to represent more accurately what it was doing however, I discovered the show action also generated just ':id' as opposed to :prison_id which would have created the same challenges. 

So it now just separates out the index view into two separate views, to show more clearly whether it is pom or spo and remove the complicated if statements.


ollietreend (commented):
The implication of this PR is that users now only see either the HOMD or POM dashboard.

Currently MPC team members have both HOMD and POM roles, and we rely on the dashboard showing a "hybrid" of both HOMD and POM dashboard links.

So this will be a change in behaviour from what we currently use.

PR #1812, opened 2021-07-14
MO-733: Make the back button return to the previous page it has visited

Katy600:
This PR makes sure the 'Change POM availability' screen (the edit page) returns the user to the previous POM screen they visited, as opposed to hard coding it to return to the list of Poms. If for some reason, they landed directly on the 'Change POM availability' screen, so there is no previous page, then pressing the back button will return them to the home page. 

**The POM edit page**
Screenshot 2021-07-14 at 09 12 35

**The POM show page**
Screenshot 2021-07-14 at 09 12 16

### DIFF
--- a/app/views/poms/edit.html.erb
+++ b/app/views/poms/edit.html.erb
@@ -85,7 +85,7 @@
 
     <div class="govuk-form-group">
       <button type="submit" class="govuk-button">Save</button>
-      <a href="<%= prison_poms_path(@prison.code, nomis_staff_id: @pom.staff_id) %>" class="govuk-link">Cancel</a>
+     <%=link_to 'Cancel', @referrer,  class: "govuk-link" %>
     </div>
 
     </form>
--- a/spec/features/edit_a_pom_spec.rb
+++ b/spec/features/edit_a_pom_spec.rb
@@ -3,119 +3,144 @@
 require "rails_helper"
 
 feature "edit a POM's details" do
-  let(:nomis_staff_id) { 485_637 }
-  let(:fulltime_pom_id) { 485_833 }
-  let(:nomis_offender_id) { 'G4273GI' }
-  let(:pom) { build(:pom) }
+  let!(:prison) { create(:prison) }
 
-  before do
-    create(:case_information, offender: build(:offender, nomis_offender_id: nomis_offender_id))
-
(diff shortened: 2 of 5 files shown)


app/views/poms/edit.html.erb
```diff
@@ -85,7 +85,7 @@
 
     <div class="govuk-form-group">
       <button type="submit" class="govuk-button">Save</button>
-      <a href="<%= prison_poms_path(@prison.code, nomis_staff_id: @pom.staff_id) %>" class="govuk-link">Cancel</a>
+     <%=link_to 'Cancel', @referrer,  class: "govuk-link" %>
```
ollietreend, line 88: Please add button styling to this link.

spec/features/edit_a_pom_spec.rb
```diff
@@ -3,119 +3,144 @@
...
-      create(:pom_detail, :inactive, prison: prison, nomis_staff_id: moic_integration_tests_staff_id)
+      expect(page).to have_css('h1', text: 'Edit profile')
+      expect(page).to have_content('Select number of days worked')
     end
 
-    it 'makes the pom have a status of active' do
-      visit "/prisons/LEI/poms"
-      click_link "Inactive staff (1)"
-      click_link 'Moic Integration-Tests'
+    describe "making an inactive POM active", :js, vcr: { cassette_name: 'prison_api/edit_poms_activate_pom_feature' } do
```
ollietreend, line 51: Does this need to be a `:js` test, or is it just a remnant from when you were working on the test?

Feature tests without `:js` run faster because they don't need to run in Firefox.

spec/support/helpers/features_helper.rb
```diff
@@ -1,10 +1,20 @@
...
+      mock_sso_response('MOIC_POM', [SsoIdentity::SPO_ROLE], prisons)
+    end
+    stub_user(staff_id: 1)
   end
 
+  # def signin_spo_user(prisons = %w[LEI RSI], name = 'MOIC_POM')
+  #   mock_sso_response(name, [SsoIdentity::SPO_ROLE], prisons)
+  #   stub_user(staff_id: 1)
+  # end
+
```
ollietreend, line 17: Please remove this commented out code – looks like it's not needed any more.

ollietreend:
Closing this as it was duplicated by MO-789, which has been deployed.

PR #1814, opened 2021-07-15
MO-735-The case information edit journey now returns to the previous page 

Katy600:
This PR fixes the update button on the case information edit page, which used to return the user to the `Reallocate a POM` page. It now uses the referrer to return the user to the last page they were on. In this feature test it is checking it can return to the `Allocation information` page. 

Screenshot 2021-07-15 at 12 40 22
Screenshot 2021-07-15 at 12 40 38

### DIFF
--- a/app/views/allocations/show.html.erb
+++ b/app/views/allocations/show.html.erb
@@ -33,7 +33,7 @@
     </tr>
     <tr class="govuk-table__row">
       <td class="govuk-table__cell">Tiering calculation</td>
-      <td class="govuk-table__cell table_cell__left_align">
+      <td id="tier" class="govuk-table__cell table_cell__left_align">
         Tier <%= @prisoner.tier %>
         <% unless auto_delius_import_enabled?(@prison.code) %>
           <a class="govuk-link pull-right" href="<%= edit_prison_case_information_path(@prison.code, @prisoner.offender_no) %>">Change</a>
--- a/app/controllers/case_information_controller.rb
+++ b/app/controllers/case_information_controller.rb
@@ -5,7 +5,7 @@ class CaseInformationController < PrisonsApplicationController
   before_action :set_case_info, only: [:edit, :show]
 
   before_action :set_referrer
-  before_action :store_referrer_in_session, only: [:edit_prd]
+  before_action :store_referrer_in_session, only: [:edit_prd, :edit]
 
   def new
     @case_info = Offender.find_by!(nomis_offender_id: nomis_offender_id_from_url).build_case_information
@@ -64,7 +64,7 @@ class CaseInformationController < PrisonsApplicationController
     @prisoner = prisoner(case_information_params[:nomis_offender_id])
     # Nothing here can fail due to radio buttons being unselectable
     @case_info.update!(case_information_params.merge(manual_entry: true))
-    redirect_to prison_prisoner_staff_index_path(active_prison_id, @case_info.nomis_offender_id)
+    redirect_to referrer
   end
 
 private
--- a/app/views/allocations/_case_information.html.erb
+++ b/app/views/allocations/_case_information.html.erb
@@ -55,7 +55,7 @@
   </tr>
   <tr class="govuk-table__row" id="tier-row">
     <td class="govuk-table__cell">Tiering calculation</td>
-    <td class="govuk-table__cell table_cell__left_align">
+    <td id="tier" class="govuk-table__cell table_cell__left_align">
     Tier <%= @prisoner.tier %>
       <%= link_to 'Change', edit_prison_case_information_path(@prison.code, @prisoner.offender_no), class: 'govuk-link pull-right' if @prisoner.manual_entry? %>
     </td>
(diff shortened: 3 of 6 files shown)


app/views/allocation_staff/_pom_table_row.html.erb
```diff
@@ -1,4 +1,4 @@
-<tr class="govuk-table__row">
+<tr id=pom-list class="govuk-table__row">
```
ollietreend, line 1: ### ⚠️ Invalid HTML

This view partial represents a single table row, which is used multiple times on the same page – i.e. it's rendered in a loop, once for each POM that needs to be displayed in the table. We should avoid giving elements in this partial a static/fixed ID attribute – because otherwise we'll end up with multiple HTML elements on the page that have the same ID.

For example, the HTML structure will look something like this:

```html
<table>
  <tr id="pom-list" class="govuk-table__row">
    <!-- First POM's table cells -->
  </tr>
  <tr id="pom-list" class="govuk-table__row">
    <!-- Second POM's table cells -->
  </tr>
  <tr id="pom-list" class="govuk-table__row">
    <!-- Third POM's table cells -->
  </tr>
</table>
```

HTML IDs must be unique on the page – so you shouldn't have multiple elements with the same `id` attribute. It's fine for multiple elements to use the same `class` though – so that might be a better thing to use here.

Also, HTML attribute values should have `"` double quotes around them – e.g. `id="pom-list"` rather than `id=pom-list`.

Browsers are very forgiving when it comes to parsing invalid HTML – they usually just figure out what we meant and don't complain about it. But nonetheless, we shouldn't knowingly write invalid HTML.

app/views/allocations/show.html.erb
```diff
@@ -33,7 +33,7 @@
     </tr>
     <tr class="govuk-table__row">
       <td class="govuk-table__cell">Tiering calculation</td>
-      <td class="govuk-table__cell table_cell__left_align">
+      <td id="tier", class="govuk-table__cell table_cell__left_align">
```
ollietreend, line 36: ### ⚠️ Invalid HTML

You don't need a comma after HTML attributes.

```suggestion
      <td id="tier" class="govuk-table__cell table_cell__left_align">
```

spec/features/update_case_information_from_allocation_page_spec.rb
```diff
@@ -0,0 +1,36 @@
+require 'rails_helper'
+
+RSpec.feature "Update case information from allocation page", type: :feature do
```
ollietreend, line 3: Now I'm looking at it – it feels like this test might be covering the same thing as the other feature spec?

This feature spec is nicer because it doesn't depend on VCR. Maybe we could remove the test from the Case Information feature spec if it's already covered here.

I can only see one context block in this file, but I think the ticket MO-735 wanted us to support this behaviour for two user journeys:

1. **For an existing allocation** Allocation information => change Case Info => allocation information
2. **For a new allocation** Allocate a POM => change Case Info => Allocate a POM

I think there might actually be a third journey, too:
3. **Reallocating a POM on an existing allocation**: Allocation information => Reallocate POM => change Case Info => Reallocate POM

So I was expecting to see more context blocks in this spec.

Katy600: Thanks @ollietreend, yes you are right. I think I really struggled understanding those flow diagrams but now you have written it out that makes lots of sense. Thanks you.

app/views/allocation_staff/_pom_table_row.html.erb
Katy600: Thanks @ollietreend, yes that makes sense.

spec/features/update_case_information_from_allocation_page_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+  let(:offender_no) { offender.fetch(:offenderNo) }
+
+  before do
+    stub_offenders_for_prison(prison.code, offenders)
+    stub_signin_spo(spo, [prison.code])
+    stub_poms(prison.code, [pom, spo])
+    stub_keyworker(prison.code, offender.fetch(:offenderNo), build(:keyworker))
+  end
+
+  context 'when updating the case information' do
```
ollietreend, line 18: It looks like this top level context block isn't really needed, since it wraps all the tests in this spec.

ollietreend (commented):
@Katy600 this is looking great 🎉 

One final thing before I hit 'approve':

Could you please squash the commit history of this branch into one commit? I don't think we'll need the details about the individual changes you've made in this branch once it's merged into `main`.

Please include the ticket number somewhere in the commit message – that way if we're ever need to refer back to this commit in the future, we'll know which Jira ticket it relates to.

For example, something like this would be good:

```
Return to the correct page after editing case info

The case information edit journey now returns to the previous page the user was on.

Jira ticket: MO-735
```

I realise we don't have strict rules around what we put into commit messages, but I think this is probably a good habit for us to get into.

ollietreend (approved):

PR #1815, opened 2021-07-16
MO-690 VLO Part II

starswan:
Victim Liaison Officers model now completely seperated from case_information model - only links to Offender

### DIFF
--- a/spec/models/victim_liaison_officer_spec.rb
+++ b/spec/models/victim_liaison_officer_spec.rb
@@ -7,12 +7,12 @@ RSpec.describe VictimLiaisonOfficer, type: :model do
     expect(subject).to validate_presence_of :email
   end
 
-  context 'with case information record' do
+  context 'with an offender record' do
     before do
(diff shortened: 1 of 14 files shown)


spec/models/victim_liaison_officer_spec.rb
```diff
@@ -7,20 +7,20 @@
...
 
-  context 'with case information record' do
+  context 'with an offender record' do
     before do
-      create(:case_information, victim_liaison_officers: [build(:victim_liaison_officer)])
+      create(:offender, victim_liaison_officers: [build(:victim_liaison_officer)])
     end
 
-    let(:ci) { CaseInformation.last }
+    let(:ci) { Offender.last }
```
ollietreend, line 15: This variable could probably have a better name now.

```suggestion
    let(:offender) { Offender.last }
```

ollietreend (approved):
Looks good, although one comment about using an `MpcOffender` instead of going directly to the `Offender`

PR #1818, opened 2021-07-21
MO-647 Take 2: Early Allocations API and Domain Events

ollietreend:
This PR represents take 2 of MO-647 – the original PR being #1778

From the original PR description:

> Create brand-new API endpoint to expose boolean early allocation status. Add an SNS event so that clients don't have to poll the endpoint. This PR also includes a significant refactor of the early allocation calculation and display code (and fixes one bug in the badge display handling around auto-eligibility) so that the state is now calculated once in the MpcOffender model and displayed by the view code, rather than the view code trying to calculate the state in a myriad of different ways

### Additions to the original PR

This PR configures the MOJ Cloud Platform to inject environment variables needed by the AWS client to publish events to the HMPPS Domain Events SNS topic:
- `AWS_ACCESS_KEY_ID`
- `AWS_SECRET_ACCESS_KEY`
- `DOMAIN_EVENTS_TOPIC_ARN`

If these environment variables are not present, then the application won't attempt to publish events when an offender's Early Allocation status changes. This allows us to run MPC in environments which don't have access to the HMPPS Domain Events SNS topic – e.g. local development and Heroku preview apps


Katy600 (approved):
I don't know a great deal about ASW topic and sns topic but based on my loose reading up about it , it seems to make some sense to me.

PR #1823, opened 2021-07-23
MO-785 - Bugfix for missing allocation history

ollietreend:
Fix a bug which was causing the 'new allocation' screen to incorrectly show offenders as having 'no allocation history' under certain conditions.

If the offender's most recent allocation was in a different prison, their allocation history was incorrectly ignored and the page would show "no history". It would only show a link to the case history page if the most recent allocation had been in the current prison.

This commit adds a test to cover this specific scenario to avoid the chance of future regressions.

### DIFF
--- a/spec/controllers/allocation_staff_controller_spec.rb
+++ b/spec/controllers/allocation_staff_controller_spec.rb
@@ -30,19 +30,6 @@ RSpec.describe AllocationStaffController, type: :controller do
     describe '#index' do
       let(:alice) { poms.first }
 
-      context 'with an allocation from a previous prison' do
-        before do
-          create(:case_information, offender: build(:offender, nomis_offender_id: offender_no))
-          create(:allocation_history, prison: build(:prison).code, nomis_offender_id: offender_no)
-        end
-
-        it 'makes a nil current POM' do
-          get :index, params: { prison_id: prison_code, prisoner_id: offender_no }
-          expect(response).to be_successful
-          expect(assigns(:current_pom)).to be_nil
-        end
-      end
-
       context 'with previous allocations for this POM' do
         before do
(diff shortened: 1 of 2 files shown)


spec/controllers/allocation_staff_controller_spec.rb
```diff
@@ -98,17 +85,20 @@
         end
       end
 
-      context 'with an allocation' do
+      context 'when the offender has been allocated before' do
+        let!(:allocation) {
+          create(:allocation_history, prison: prison_code, nomis_offender_id: offender_no, primary_pom_nomis_id: poms.last.staff_id)
+        }
```
thomasleese, line 91:
```suggestion
        let!(:allocation) do
          create(:allocation_history, prison: prison_code, nomis_offender_id: offender_no, primary_pom_nomis_id: poms.last.staff_id)
        end
```

I tend to prefer `do`/`end` with multi-line `let` since this statement is too long to fit on one line.

spec/controllers/allocation_staff_controller_spec.rb
```diff
@@ -123,6 +113,53 @@
...
+          it 'sets @allocation so that last_event can be displayed' do
+            get :index, params: { prison_id: prison_code, prisoner_id: offender_no }
+            expect(response).to be_successful
+            expect(assigns(:allocation)).to eq(allocation)
+          end
+        end
+
+        context "when they weren't deallocated from the last prison (a known bug)" do
+          it 'sets @allocation so that last_event can be displayed' do
+            get :index, params: { prison_id: prison_code, prisoner_id: offender_no }
```
thomasleese, line 151: This line could go in a `before` block as every test does it at the beginning.

thomasleese (approved):
Left a couple of comments, but no major blockers.

PR #1829, opened 2021-08-04
MO-791: Fix womens handover date bug

ollietreend:
This PR is formed of two parts:

1. Fix the bug described in MO-791
2. Refactor the `HandoverDateService` code to make it more readable

### Bug fix

In the first commit of this PR, I fixed the bug and added tests to cover it. I think this bug previously went unnoticed because the structure of the code was quite unclear, plus our test assertions weren't tight enough.

### Refactor code

This class is always a bit scary to work with, and I think there's potential for more bugs to crop up due to this. In the second commit of this PR, I've refactored the code in an attempt to make it easier to follow and reason with.

### DIFF
--- a/app/controllers/prisoners_controller.rb
+++ b/app/controllers/prisoners_controller.rb
@@ -33,6 +33,9 @@ class PrisonersController < PrisonsApplicationController
 
   def show
     @prisoner = OffenderService.get_offender(params[:id])
+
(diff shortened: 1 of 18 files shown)


Katy600 (approved):
I have no comments I am afraid! I think all your changes make sense me as a result of you walking through it with me earlier! :)

PR #1830, opened 2021-08-06
MO-811: Fix a bug where PRDs couldn't be updated

ollietreend:
Jira ticket: MO-811

---

This commit fixes a bug where users weren't able to update Parole Review Dates (PRDs) for offenders.

I took the opportunity to re-write the feature test for updating PRDs, since it needed some love: it didn't test the 'update' journey (so didn't cover this bug), and it was dependent on VCR (which we're trying to move away from in favour of stubbed API calls).

The new feature test covers a few journeys:

1. The PRD cannot be updated if the TED is in the future
2. A new PRD can be entered (first time entry)
3. An existing PRD can be updated

This behaviour is accessible to both HOMDs and POMs, so the feature test covers both types of user.

I've also introduced a couple of helper methods for feature tests – they make it easier to find and assert values in the key/value tables on prisoner profile pages.

### DIFF
--- a/spec/features/change_parole_review_date_spec.rb
+++ b/spec/features/change_parole_review_date_spec.rb
@@ -1,46 +1,152 @@
 require 'rails_helper'
 
 RSpec.feature "ChangeParoleReviewDates", type: :feature do
-  # This ID has an indeterminate sentence
-  let(:nomis_offender_id) { 'G0549UO' }
-  let!(:case_info) { create(:case_information, offender: build(:offender, nomis_offender_id: nomis_offender_id)) }
-  let!(:alloc) { create(:allocation_history, prison: prison, nomis_offender_id: nomis_offender_id, primary_pom_nomis_id: 485_926) }
-  let(:year) { Time.zone.today.year + 1 }
-  let(:yesterday) { Time.zone.yesterday }
-  let!(:prison) { 'LEI' }
+  let(:prison) { create(:prison) }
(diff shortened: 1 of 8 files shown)


spec/features/change_parole_review_date_spec.rb
```diff
@@ -1,46 +1,152 @@
...
+  let(:tariff_date) { Time.zone.today + 1.year }
+  let(:existing_prd) { nil }
+  let(:nomis_offender_id) { nomis_offender.fetch(:offenderNo) }
+
+  # Stub API response to represent an offender in NOMIS
+  let(:nomis_offender) {
+    build(:nomis_offender, agencyId: prison.code,
+          sentence: attributes_for(:sentence_detail, :indeterminate, tariffDate: tariff_date)
+    )
+  }
```
thomasleese, line 14:
```suggestion
  let(:nomis_offender) do
    build(:nomis_offender, agencyId: prison.code,
          sentence: attributes_for(:sentence_detail, :indeterminate, tariffDate: tariff_date)
    )
  end
```

I tend to prefer using `do`/`end` for multiline blocks.

thomasleese (approved):
Change looks good.

I'd be tempted to split this up in to two commits though, one which fixes the bug (and then probably does a minor fix on the tests) and then another which refactors the tests to be clearer. This means we're able to see the bug fix working on its own.

PR #1834, opened 2021-08-11
MO-734 user is re-directed after completing allocation 

Katy600:
Screenshot 2021-08-11 at 13 28 47

1. This has already been fixed else where it seems. 
2. This has been revised and now redirects the user to ‘3.2 View a case’, and not ‘1.8 See allocations list’ after they have allocated a previously unallocated offender. Previously it redirected to the 'make new allocations' list - which was a bug. 

As part of the journey, the user will click the 'Complete allocations' button once the allocation has been made. This redirect happens inside the 'update' action in the BuildAllocationsController. This button is shared by 2 journeys. One where a user wants to allocate for the first time and one where a user wants to reallocate. Previously it was hardcoded to always return to the unallocated list used by the former, which is why the bug existed. 

There for a @refferer has been used to send the user back to the relevant page however I found it quite tricky as the journey goes through several controllers: PrisonersController, AllocationStaffController and BuildAllocationsController. I had to put the store_referrer_in_session in the AllocationsStaffController to capture any visited pages prior to the index but it still kept on displaying a page within the BuildAllocatonsController.  I ended up putting an 'except:[:override] inside the AllocationsStaffController, which felt a bit strange as override is not part of that controller. But it worked. However rubocop errored and told me that I had to put an empty `override' action inside AllocationsStaffController for it to be valid. 

I had to fix a few broken tests - which I am still confused by. 

**The two tests that confused me were:**
1. scenario 'accepting the recommended POM type'
2. scenario 'overriding the recommended POM type

This could be because of my gaps in domain knowledge. But when I got visibility over them it looked like they were doing the opposite of what the set out to do. Test 1. was displaying the override POMs, and so `Moic Integration-Tests` was not available in the list and Test 2. was displaying the recommended POMs, so `Moic Pom' was not available. 

So I have edited them. But I am not confident I fully understand them.

### DIFF
--- a/spec/features/allocate_feature_spec.rb
+++ b/spec/features/allocate_feature_spec.rb
@@ -5,8 +5,9 @@ feature 'Allocation' do
   let!(:prison_officer_nomis_staff_id) { 485_926 }
   let!(:nomis_offender_id) { 'G7266VD' }
   let(:offender_name) { 'Omistius Annole' }
-  let!(:never_allocated_offender) { 'G9403UP' }
+  let!(:never_allocated_offender_id) { 'G9403UP' }
   let!(:unallocated_offender_name) { 'Albina, Obinins' }
+  let(:recently_allocated_offender_name) { 'Obinins Albina' }
 
   let!(:probation_officer_pom_detail) {
     PomDetail.create!(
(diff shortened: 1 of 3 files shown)


spec/features/allocate_feature_spec.rb
```diff
@@ -19,55 +20,64 @@
...
+    scenario 'accepting the recommended POM type', :js, vcr: { cassette_name: 'prison_api/create_new_allocation_feature' } do
+      expect(page).to have_content('Determinate')
 
-    # Allocate to the Probation POM called "Moic Integration-Tests"
-    within '#recommended_poms' do
-      within row_containing 'Moic Integration-Tests' do
-        click_link 'Allocate'
+      # Allocate to the Prison POM called "Moic Pom"
+      within '#recommended_poms' do
+        within row_containing 'Moic Pom' do
```
ollietreend, line 44: When we talked it through, we were both confused as to why the recommended POM type had changed from wanting a Probation POM to now wanting a Prison POM...

spec/features/allocate_feature_spec.rb
```diff
@@ -19,55 +20,64 @@
...
 
-  scenario 'accepting the recommended POM type', vcr: { cassette_name: 'prison_api/create_new_allocation_feature' } do
-    visit prison_prisoner_staff_index_path('LEI', nomis_offender_id)
+  context 'when a journey begins on the "Make new allocations" page' do
+    before do
+      visit unallocated_prison_prisoners_path('LEI')
+
+      # Takes you to the Review case and allocate page where users review case details and scroll down to select from a list of POMs
+      click_link unallocated_offender_name
+      expect(page).to have_current_path(prison_prisoner_staff_index_path('LEI', never_allocated_offender_id))
```
ollietreend, line 36: ...but it looks like this change is probably the culprit!

The previous test loaded the page:

```ruby
prison_prisoner_staff_index_path('LEI', nomis_offender_id)
```

whereas now it loads the page:

```ruby
prison_prisoner_staff_index_path('LEI', never_allocated_offender_id)
```

We're actually allocating to a different offender. Previously we allocated the offender with ID `nomis_offender_id`, whereas now we're using `never_allocated_offender_id`. I'm guessing that they've got different tiers, so need a different type of POM.

---

This would explain why the "accepting the recommended POM type" journey ended up looking like it was actually taking the "override" journey. Because we were allocating a different offender than before, so getting a different POM type recommendation.

spec/features/allocate_feature_spec.rb
```diff
@@ -143,25 +153,29 @@
...
+  scenario 're-allocating', :js, vcr: { cassette_name: 'prison_api/re_allocate_feature' } do
     create(
       :allocation_history,
       prison: 'LEI',
       nomis_offender_id: nomis_offender_id,
       primary_pom_nomis_id: 485_735,
       recommended_pom_type: 'probation'
     )
-
+    # 'Goes to allocations' page
```
ollietreend, line 164: Minor niggle: is the `'` quotation mark in the wrong place in this comment?

```suggestion
    # Goes to 'allocations' page
```

spec/features/allocate_feature_spec.rb
```diff
@@ -19,55 +20,64 @@
...
+      expect(page).to have_css('h1', text: 'Confirm allocation')
 
-    click_button 'Complete allocation'
+      click_button 'Complete allocation'
 
-    expect(current_url).to have_content(unallocated_prison_prisoners_path('LEI'))
-    expect(page).to have_css('.notification', text: "#{offender_name} has been allocated to Moic Integration-Tests (Probation POM)")
-  end
+      # Return to the "Make new allocation" page
+      expect(current_url).to have_content(unallocated_prison_prisoners_path('LEI'))
```
ollietreend, line 54: ### Possible readability improvement?

I wonder whether these tests would be more readable if we pulled the 'start page' (first page of the journey) out into a `let()` variable that we then reference at both the beginning and end of the journey.

For example, where we currently have:

```ruby
before do
  visit unallocated_prison_prisoners_path('LEI')
  # do some things...
end

it 'does the journey' do
  # do some things...
  expect(current_url).to have_content(unallocated_prison_prisoners_path('LEI'))
end
```

How about this?

```ruby
let(:start_page) { unallocated_prison_prisoners_path('LEI') }

before do
  visit start_page
  # do some things...
end

it 'does the journey' do
  # do some things...
  expect(current_url).to eq(start_page)
end
```

The main thing I like about this approach is that it should hopefully make it more obvious that we're expecting to end up back at the exact same page URL. You could say it's a bit more self-describing.

This is just my opinion, though! So completely up to you whether you choose to change it or not. (No offence taken if you don't! 🙂)

ollietreend (commented):
This looks great, @Katy600! 👍🏻

Really nice work.

I've put in a couple of comments, but they're only small things.

Plus a couple of comments which I think might explain the domain knowledge gap that you thought you had. I don't think you had one after all!

spec/features/allocate_feature_spec.rb
Katy600: @ollietreend Oh goodness! Thank you Ollie! I can't believe I missed that. Although I am sure I originally copied that url and then changed it because it didn't work. It is too long ago to remember!

spec/features/allocate_feature_spec.rb
Katy600: @ollietreend Yes I like that. Definitely more readable. Thanks Ollie. :)

ollietreend (approved):

PR #1835, opened 2021-08-11
Stop old cronjobs if they haven't finished running

ollietreend:
This commit configures the concurrency policy of the Kubernetes CronJobs so that pods are killed and replaced if they're still running by the time the next iteration of that job is due to run.

Previously the cronjobs were configured to forbid concurrent runs – but this meant that an old zombie job could prevent new jobs from being started as scheduled.

This new behaviour is more appropriate because none of the cronjobs should take longer than 24 hours to complete – which is the typical frequency that they're scheduled to run to.

We've seen instances where some of our 'overnight sweep' (batch processing) cronjobs have hung and failed to stop. This caused important nightly jobs to fail to run until the hanging job was manually killed – which typically could have been the case for several days until the bad job was noticed. This change means that this should no longer happen
 – a hanging job will now only run for up to 24 hours before it gets replaced by the next iteration of that cronjob.

Also: I've deleted a couple of cronjobs which were scheduled in staging but are no longer needed. In both cases, it looks like they were only created as a one-off test of the job's functionality.


Katy600 (approved):
Thank you for such a clear explanation. This makes lots of sense :)

PR #1838, opened 2021-08-17
Make a feature test more resilient to change

ollietreend:
The feature test `spec/features/show_poms_feature_spec.rb` runs against data held in NOMIS dev (T3). It was previously quite brittle due to expecting the POM "Moic Pom" to appear 8th in the list of POMs. It's inevitable that this will change as team members come and go, and the list of available POMs changes accordingly.

To make the test more resilient to such external change, I've introduced a new helper method `row_containing`. Now rather than expecting the "Moic Pom" to always be in 8th position in the table, we can simply find the table row which contains that POM's name.

### DIFF
--- a/spec/features/allocate_feature_spec.rb
+++ b/spec/features/allocate_feature_spec.rb
@@ -26,14 +26,14 @@ feature 'Allocation' do
     signin_spo_user
   end
 
-  scenario 'accepting a recommended allocation', vcr: { cassette_name: 'prison_api/create_new_allocation_feature' } do
+  scenario 'accepting the recommended POM type', vcr: { cassette_name: 'prison_api/create_new_allocation_feature' } do
     visit prison_prisoner_staff_index_path('LEI', nomis_offender_id)
 
     expect(page).to have_content('Determinate')
 
+    # Allocate to the Probation POM called "Moic Integration-Tests"
     within '#recommended_poms' do
-      # Allocate to the POM called "Moic Integration-Tests"
-      within page.find(:css, 'tr', text: 'Moic Integration-Tests') do
+      within row_containing 'Moic Integration-Tests' do
         click_link 'Allocate'
       end
     end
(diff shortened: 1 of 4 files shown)


spec/features/allocate_feature_spec.rb
```diff
@@ -51,9 +51,8 @@
     visit prison_prisoner_staff_index_path('LEI', nomis_offender_id)
     find('#non-recommended-accordion-section-heading').click
 
-    # Amit is 6th on the non-recommended list
     within '#non-recommended-accordion-section' do
-      within 'tbody > tr:nth-child(6)' do
+      within row_containing 'Amit Muthu' do
         click_link 'Allocate'
       end
     end
```
Katy600, line 58: @ollietreend Does it still rely on Amit Muthu to always be there?

spec/features/case_information_feature_spec.rb
```diff
@@ -84,13 +84,15 @@
     it "clicking back link after viewing prisoner's case information, returns back the same paginated page",
        vcr: { cassette_name: 'prison_api/case_information_back_link' }, js: true do
       visit missing_information_prison_prisoners_path('LEI', page: 3)
+      expect(current_page_number).to eq(3)
```
Katy600, line 87: @ollietreend Why is checking the page number important?

Katy600 (approved):
@ollietreend I have approved it! I just have a couple of questions! :)

spec/features/case_information_feature_spec.rb
ollietreend: @Katy600 the test scenario is:

> clicking back link after viewing prisoner's case information, **returns back the same paginated page**

But the test was only asserting that we end up on a page with the heading 'Add missing details'. It didn't actually prove that we returned to page 3 of the listings, so wasn't testing what it claimed to test. We could have actually returned to page 1 of the results and the test would still have passed.

That's why I've added the check for page numbers. Just to make it clear that it is, in fact, behaving correctly.

spec/features/allocate_feature_spec.rb
ollietreend: @Katy600 Yeah, it does 😕 which isn't ideal.

I've just updated it to use our test POM "Moic Pom" instead. Thanks for the nudge to do this! 🤓 🙌🏻

spec/features/case_information_feature_spec.rb
Katy600: Oh of course! Thanks @ollietreend! That makes so much sense :)

PR #1840, opened 2021-08-19
Mo 573 UI home page cards

Katy600:
This PR modifies the home page cards to use the same card styles as the ones on DPS.

[Link to dps home page](https://digital-dev.prison.service.justice.gov.uk/)


ollietreend (approved):

app/views/dashboard/index.html.erb
```diff
@@ -97,9 +97,9 @@
 <% end %>
 
 <% if @is_spo %>
-  <h2 class="govuk-heading-m no-bottom-margin">Manage staff</h2>
-  <hr class="govuk-section-break govuk-section-break--visible">
-  <div class="govuk-grid-row dashboard-row">
+  <h2 class="govuk-heading-m no-bottom-margin govuk-!-margin-top-3">Manage staff</h2>
+  <hr class="govuk-section-break govuk-section-break--visible govuk-!-margin-bottom-5">
+  <div class="govuk-grid-row card-groups">
```
ollietreend, line 102: You've used `card-groups` here rather than `card-group`. Is this a typo?

```suggestion
  <div class="govuk-grid-row card-group">
```

app/views/dashboard/index.html.erb
```diff
@@ -86,7 +86,7 @@
     %>
   </div>
 
-  <div class="govuk-grid-row dashboard-row">
+  <div class="govuk-grid-row card-group">
```
ollietreend, line 89: If the `dashboard-row` class isn't used any more, it should be safe to remove whatever CSS rules were specified for it in our CSS files.

It's always better to clean things up as we go, otherwise the CSS can become unwieldy with rules for unused classes. 😱 

Before doing so: I'd suggest making sure none of the other views use the `dashboard-row` class for anything.

Katy600: @ollietreend dashboard-row is used in one place sadly. On the contact_us page!

app/views/dashboard/index.html.erb
Katy600: @ollietreend Oh well spotted. It is a typo!

app/views/dashboard/index.html.erb
ollietreend: @Katy600 I've just had a look at the Contact Us page and think it's safe to remove. It adjusts the styling very slightly – but I don't think `dashboard-row` was being particularly helpful on that page anyway.

As part of this PR, could you please remove the use of `dashboard-row` on the Contact Us page (no need to replace it with anything else), and remove the associated CSS styles from `application.scss`?

Thanks! 🙏🏻

PR #1841, opened 2021-08-19
MO-826: Deallocate offenders upon release from prison

ollietreend:
Jira ticket: MO-826

This PR fixes a bug which has become apparent due to offenders not having been automatically deallocated upon release from prison.

### Background

We process offender movements overnight – and if we see that an offender has been released from prison, we automatically deallocate their POM. However due to recent changes in the behaviour of the `OffenderService.get_offender` method, it meant that we could no longer process offenders who aren't in a recognised Prison.

### Old behaviour

Prior to PR #1767, the offender service would happily return a `HmppsApi::Offender` object even if the offender's prison code wasn't recognised within our local list of prisons.

And since in this case we only need the offender's name, we could still call `offender.full_name` without any problems.

### New behaviour

After #1767, the offender service no longer returns a `HmppsApi::Offender` object. Instead, it returns an `MpcOffender` wrapper object which contains a `HmppsApi::Offender`, but also a `Prison` and other records. This new & improved wrapper data model gives us a very clear way to combine offender data from different sources, while still recognising that they come from separate places.

But the knock-on effect of that is: an `MpcOffender` object cannot be returned from the offender service unless they're in a recognised `Prison` (i.e. one which we have a record for in our local database).

Instead, we want this piece of code to continue using a `HmppsApi::Offender` object, since all we need is the offender's name. We don't care which prison they're in, nor do we care about anything else on the offender object – we only need their name.

### DIFF
--- a/spec/jobs/movements_on_date_job_spec.rb
+++ b/spec/jobs/movements_on_date_job_spec.rb
@@ -1,38 +1,76 @@
 require 'rails_helper'
 
 RSpec.describe MovementsOnDateJob, type: :job do
-  let!(:prison) { create(:prison, code: 'MDI') }
-  let(:nomis_offender_id) { 'G3462VT' }
-  let!(:alloc) { create(:allocation_history, nomis_offender_id: nomis_offender_id, secondary_pom_nomis_id: 123_435, prison: 'MDI') }
+  let(:nomis_offender_id) { nomis_offender.fetch(:offenderNo) }
+  let(:nomis_offender) { build(:nomis_offender, agencyId: to_prison_code) }
+  let(:case_info) { create(:case_information) }
+  let!(:offender_record) { create(:offender, nomis_offender_id: nomis_offender_id, case_information: case_info) }
+  let(:today) { Time.zone.today }
+  let(:yesterday) { today - 1.day }
+  let(:pom) { build(:pom) }
+  let!(:allocation) { create(:allocation_history, nomis_offender_id: nomis_offender_id, primary_pom_nomis_id: pom.staff_id, prison: from_prison_code) }
+  let(:mail_params) {
+    {
+      email: pom.email_address,
+      pom_name: pom.first_name.titleize,
+      offender_name: "#{nomis_offender.fetch(:lastName)}, #{nomis_offender.fetch(:firstName)}".titleize,
+      nomis_offender_id: nomis_offender_id,
+      prison_name: Prison.find(from_prison_code).name,
+      url: "http://localhost:3000/prisons/#{from_prison_code}/staff/#{pom.staff_id}/caseload"
+    }
+  }
 
   before do
-    stub_auth_token
+    stub_offender(nomis_offender)
+    stub_poms(from_prison_code, [pom])
+
+    # Stub offender movements from yesterday
+    stub_request(:get, "#{ApiHelper::T3}/movements?fromDateTime=#{yesterday - 1.year}T00:00&movementDate=#{yesterday}").
+      to_return(body: [movement].to_json)
+
+    # Expect POM to be emailed about the deallocation
+    expect(PomMailer).to receive(:offender_deallocated).with(mail_params).and_return(double(deliver_later: nil))
   end
 
-  it 'deallocates' do
-    allow(OffenderService).to receive(:get_offender).and_return(build(:hmpps_api_offender))
+  context 'when the offender has been released from prison' do
+    let(:from_prison_code) { create(:prison).code }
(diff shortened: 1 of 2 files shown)


spec/jobs/movements_on_date_job_spec.rb
```diff
@@ -1,38 +1,76 @@
...
-  end
+    stub_offender(nomis_offender)
+    stub_poms(from_prison_code, [pom])
+
+    # Stub offender movements from yesterday
+    stub_request(:get, "#{ApiHelper::T3}/movements?fromDateTime=#{yesterday - 1.year}T00:00&movementDate=#{yesterday}").
+      to_return(body: [movement].to_json)
 
-  it 'deallocates' do
-    allow(OffenderService).to receive(:get_offender).and_return(build(:hmpps_api_offender))
```
Katy600, line 13: When the new offender wrapper was introduced and therefore the bug too, then this test didn't catch it? Is that why you have deleted it? It is stubbing OffenderService and returning the hmpps_api_offender so therefore hiding the bug?

spec/jobs/movements_on_date_job_spec.rb
```diff
@@ -1,38 +1,76 @@
...
+    # Expect POM to be emailed about the deallocation
+    expect(PomMailer).to receive(:offender_deallocated).with(mail_params).and_return(double(deliver_later: nil))
```
Katy600, line 32: So this is stubbing at a lower level? - so just the movements? And then testing the PomMailer is able to receive a deallocated offender - with no Prison associated with it. That makes sense. 

Does it mater that it is testing PomMailer inside the MovementsOnDataJob spec?

spec/jobs/movements_on_date_job_spec.rb
```diff
@@ -1,38 +1,76 @@
...
 
-    expect(alloc.primary_pom_nomis_id).not_to be_nil
-    expect(alloc.secondary_pom_nomis_id).not_to be_nil
+    it 'deallocates and emails the POM' do
+      expect(allocation).to be_active
+      described_class.perform_now(today.to_s)
+      allocation.reload
+      expect(allocation).not_to be_active
+    end
+  end
```
Katy600, line 54: It does feel like allocationHistory is being tested inside MovementOnDateJob. It feels a bit like an integration tests because you are testing how the actions of one effect the other. Is this just the nature of testing Jobs or is this because they are quite tightly coupled?

Katy600 (approved):
@ollietreend I left lots of comments but they were really to help me make sense of it. So sorry about that Ollie. I think it helps my focus my brain when I write it out. It all make sense to me though. I found the tests really helped me to understand things so the implementation makes lots of sense. And your comments at the start were really clear! So thank you!

spec/jobs/movements_on_date_job_spec.rb
ollietreend: @Katy600 yes – you're spot on. I initially wanted to write this as a feature spec – but then I realised that it doesn't really make sense to test this as a browser based user journey, since it's not! Then I spotted that there was already a job spec which half-covered the existing functionality. So my approach was to build out that spec instead of create a new one.

Either way: it's definitely a high level test. From a business point of view, we want to be sure that POMs are deallocated (and notified by email) when offenders are released or transferred out of that prison – so I think it's useful to have a test which broadly covers that.

The implementation itself is spread across different classes (as you've already called out: AllocationHistory does the 'meat' of the work). However this job is actually the only thing that uses that functionality – so in terms of implementation code, it could just as easily be shuffled over into the job itself. (And it might make more sense to do so, because it'd remove some noise from the AllocationHistory model.)

I'm not sure if that really answers anything – but I'm just agreeing with what you've said, and explaining my thinking a bit!

spec/jobs/movements_on_date_job_spec.rb
ollietreend: Yep – that's correct! The behaviour of the `OffenderService.get_offender` method (and the type of object returned) changed. But since it was stubbed it out here, it didn't reflect the new behaviour.

The method used to return a `HmppsApi::Offender` object – which is why this stub returned:
```ruby
build(:hmpps_api_offender)
```

However it now returns an MpcOffender object, which is a wrapper around that & other records.

The two methods have a similar interface, but very subtly different meanings/behaviour.

spec/jobs/movements_on_date_job_spec.rb
Katy600: @ollietreend Thanks Ollie. Yes that makes so much sense. It does feel it would make more sense to be in the job now that you have mentioned it. As if it is not being used by any thing else then it feels like it shouldn't be the responsibility of the AllocationHistory model to home it.

PR #1846, opened 2021-09-01
MO-812 Add google tag manager and remove google analytics

Katy600:


ollietreend (approved):
@Katy600 please could you squash these commits into one – and then I think we're good to deploy this.

I've saved our changes in Google Tag Manager, so this should now be production ready.

PR #1847, opened 2021-09-01
Remove unnecessary model concern 'Deserialisable'

ollietreend:
This concern converted date strings into Date objects. It was only used in the SentenceDetail model which needed to parse all of the sentence dates received from the Prison API.

Rails' ActiveSupport library includes the helper method `.to_date` which can be called on String objects – which essentially makes this concern unnecessary.

I also found that the Deserialisable concern had been included into lots of classes that didn't even use it. Whilst only a small change, it's one more piece of cruft removed from the codebase.

I also took the opportunity to alphabetically sort the date fields in the SentenceDetail model. This helps with code readability and has the helpful side effect of grouping similar fields together (e.g. the 'override' date fields).


Katy600 (approved):
That makes sense :)

PR #1848, opened 2021-09-03
Enable Google Tag Manager to track Turbolinks page views

ollietreend:
Also remove the unused method `createFunctionWithTimeout()`.


Katy600 (approved):
So sorry Ollie!

PR #1849, opened 2021-09-03
Mo 501 remove the 'beta banner'

Katy600:
This PR:

1. Removes the beta banner 
3. Moves the prison switcher and changes the text to ‘change location’ for the switcher. Changes the font to 19px. This will make the switcher consistent with DPS. 
5. Remove the feedback link
6. It also changes the formatting for the prison name to look like this `Leeds (HMP)`, instead of `HMP Leeds` which is also to be more inline with DPS. I am not sure I approached this in the right way though and wasn't sure what was best practice in this situation!

Screenshot 2021-09-03 at 17 25 12


app/helpers/layouts_helper.rb
```diff
@@ -0,0 +1,5 @@
+module LayoutsHelper
+  def prison_switcher_name_format(prison)
+    prison.name.split[1] + ' (HMP)'
```
ollietreend, line 3: If we do need to keep this helper (i.e. we don't just change the prison names in the database), it looks like there's a bug with the current implementation.

`prison.name.split[1]` splits the name into words, and uses the second word.

This works for prison names which only have two words, e.g. `HMP Leeds`. This is correctly formatted as `Leeds (HMP)`.

However some prison names consist of more than one word, e.g. `HMP Isle of Wight`. This is incorrectly formatted as `Isle (HMP)` and you lose the `of Wight` bit.

If we keep this helper: it'd be good to have a test case to cover this.

ollietreend (commented):
Nice work @Katy600 🎉 

I've added some comments around the formatting of prison names – things for us to think about. Otherwise everything else looks great.

app/helpers/layouts_helper.rb
ollietreend: Also worth noting that this formatting might be slightly odd for `HMP Peterborough (Female)`.

It'll become `Peterborough (Female) (HMP)`.

Is that what DPS does, too?

ollietreend (approved):
Looks good @Katy600! 🙌🏻  Thanks for taking the time to make it a site-wide change.

This should be good to merge after a squash 🙂

PR #1850, opened 2021-09-08
MO-843: default to NPS if resourcing not assessed

ollietreend:
Jira ticket: MO-843

This commit changes the behaviour of `OffenderService.get_community_data` to gracefully handle resourcing responses from the Community API which are missing the `enhancedResourcing` field.

If the `enhancedResourcing` field is missing, we handle it the same as if we received a 404 Not Found response – i.e. we default the service provider to NPS.

Only if the `enhancedResourcing` field is present and `false` will we designate an offender as being under CRC rules.

### DIFF
--- a/app/services/offender_service.rb
+++ b/app/services/offender_service.rb
@@ -73,19 +73,20 @@ class OffenderService
                               .select { |r|
                                 r.fetch(:active) && r.key?(:registerLevel) && r.dig(:registerLevel, :code).starts_with?('M')
                               }
-      resourcing = begin
-                   HmppsApi::CommunityApi.get_latest_resourcing(nomis_offender_id).deep_symbolize_keys
-                   rescue Faraday::ResourceNotFound
-                     { enhancedResourcing: true } # assume NPS when not found
-                 end
+
+      is_nps = begin
+                 HmppsApi::CommunityApi.get_latest_resourcing(nomis_offender_id)
+                                       .fetch('enhancedResourcing')
+               rescue Faraday::ResourceNotFound, KeyError
+                 true # default to NPS if 404 Not Found or no enhancedResourcing field present in the response
+               end
 
       com = community_info.fetch(:offenderManagers).detect { |om| om.fetch(:active) }
       {
           noms_no: nomis_offender_id,
           tier: community_info.fetch(:currentTier),
           crn: community_info.dig(:otherIds, :crn),
-
-          service_provider: resourcing.fetch(:enhancedResourcing) ? 'NPS' : 'CRC',
+          service_provider: is_nps ? 'NPS' : 'CRC',
           offender_manager: com.dig(:staff, :unallocated) ? nil : "#{com.dig(:staff, :surname)}, #{com.dig(:staff, :forenames)}",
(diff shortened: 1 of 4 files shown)


app/services/offender_service.rb
```diff
@@ -73,19 +73,20 @@ def get_community_data(nomis_offender_id)
...
-                   rescue Faraday::ResourceNotFound
-                     { enhancedResourcing: true } # assume NPS when not found
-                 end
+
+      is_nps = begin
+                 HmppsApi::CommunityApi.get_latest_resourcing(nomis_offender_id)
+                                       .fetch('enhancedResourcing')
+               rescue Faraday::ResourceNotFound, KeyError
+                 true # default to NPS if 404 Not Found or no enhancedResourcing field present in the response
+               end
```
Katy600, line 83: @ollietreend I am just trying to understand this code. If 'enhancedResourcing' exists then `is_nps` is assigned true? But if it doesn't then the .fetch method errors, so it does reach that line of code so can't be assigned true. later in the code is_nps is only checking for true though, so it doesn't really need to be explicitly assigned false. But .fetch is very shouty so, will this error be displayed somewhere?

Katy600: Ok I think partially understand this now as your comment next to true is very helpful. And I have not seen begin and rescue used before. So they prevent the code from exiting when it hits an exception. Instead it allows the code to continue, and then in this case it can assign the final return value of the block to is_nps - which is true. But what happens if their are no errors raised, won't it still be assigned true?

Katy600 (approved):
Sorry @ollietreend I have approved this, I just don't fully understand how is_nps works!!

app/services/offender_service.rb
ollietreend: That's correct @Katy600.

The `enhancedResourcing` field returned from the Community API will be in one of three states:
- `true`
- `false`
- omitted from the API response (i.e. it's not `nil` or any other value – the field simply doesn't exist in the API response)

There is a fourth potential API response that we can expect, which is a 404 Not Found error. This indicates that the offender hasn't had an assessment yet.

So the `begin` part of the block (think of this as the 'happy path') calls the API and returns the value of the `enhancedResourcing` field. So long as the API call succeeds and the field is present in the response, it should be either `true` or `false` – which will be assigned as the value of `is_nps`.

The `rescue` block ('unhappy path') captures two types of exception that we want to handle with a fallback:
1. a `Faraday::ResourceNotFound` exception is raised if the API returned a 404 Not Found error response
2. a `KeyError` exception is raised if we cannot `.fetch(:enhancedResourcing)` because the field isn't present

In either of those error cases, we want to handle them with the same fallback behaviour: return `true`, which will be assigned as the value of `is_nps`.

In other words:
- so long as we receive `enhancedResourcing` from the API, we use that
- in any other case, we default to `true`

This is because the policy team have asked that offenders who haven't been assessed yet be defaulted to NPS.

---

I think I might be able to make the spec a little clearer by renaming the scenarios. Thanks for pointing this out as confusing – hopefully a clearer set of test examples will help remind us what this all means in the future.

PR #1852, opened 2021-09-15
Mo 836 use different client ids for auth vs. api access

Katy600:
This PR uses the new api client that has been created in order to split the server side client from the api client. This involved updating a lot of environmental variables with new client id and client secret.


README.md
```diff
@@ -93,7 +93,7 @@ Required
 | Env var  | Description  |
 |---|---|
 | KEYWORKER_API_HOST | The host where the keyworker API is hosted |
-| NOMIS_OAUTH_CLIENT_ID | The client ID of the application in OAUTH |
+| HMPPS_OAUTH_CLIENT_ID | The client ID of the application in OAUTH |
 | NOMIS_OAUTH_HOST  |  This is the full URL of the OAUTH host where access is granted to read from the relevant APIs |
```
ollietreend, line 97: Could you also update this README file to include the _new_ environment variables that we need?

I know you've already listed them in the `.env.example` file, but it's also handy to have them documented in the main project README.

ollietreend (approved):
Great work @Katy600 – just one comment from me.

Also, don't forget to update the environment variable configurations in Heroku and CircleCI 🙂

README.md
Katy600: Oh yes of course!

PR #1855, opened 2021-09-17
MO-806: Use Prison Search API as much as possible

ollietreend:
Jira ticket: MO-806


app/models/hmpps_api/offender.rb
```diff
@@ -19,34 +17,26 @@ def awaiting_allocation_for
...
+    attr_accessor :prison_arrival_date
+
+    attr_reader :first_name, :last_name, :prison_id, :offender_no, :cell_location, :complexity_level,
+                :date_of_birth, :sentence, :main_offence
+
+    def initialize(prison_api_payload:, search_api_payload:, category:, latest_temp_movement:, complexity_level:)
+      @first_name = search_api_payload.fetch('firstName')
+      @last_name = search_api_payload.fetch('lastName')
+      @offender_no = search_api_payload.fetch('prisonerNumber')
+      @convicted_status = prison_api_payload['convictedStatus']
```
Katy600, line 29: Just out of curiosity, why in some instances have we not used `.fetch`

ollietreend: @Katy600 I'm not entirely sure the reason for it – but the `[]` accessor is useful when you want to gracefully return `nil` if that key doesn't exist. I guess there's potential for the `convictedStatus` field to be missing from the Prison API response... 🤷🏻

Watch this space, though, because MO-852 removes the `convictedStatus` field so that we're no longer dependent on `prison_api_payload` – so we can make one less API request 🙂

Katy600 (approved):
Hi @ollietreend I am sorry this took so long. I have added a few questions just for me really. But  not pressure to answer them as it is mostly just for my learning and understanding.

PR #1857, opened 2021-09-23
MO 409 Create an Assessment API client to get most recent OASys assessment date

Katy600:
This PR creates an Assessment Api client that returns the most recent OASys assessment date.

### DIFF
--- a/app/services/hmpps_api/assessment_api.rb
+++ b/app/services/hmpps_api/assessment_api.rb
@@ -1,0 +1,22 @@
+module HmppsApi
+  class AssessmentApi
+    def self.client
+      host = Rails.configuration.assessment_api_host
+      HmppsApi::Client.new(host)
+    end
+
+    def self.get_latest_oasys_date(offender_no)
+      safe_offender_no = URI.encode_www_form_component(offender_no)
+      route = "/offenders/nomisId/#{safe_offender_no}/assessments/summary?assessmentType=LAYER_3&assessmentStatus=COMPLETE"
+
+      all_assessments = self.client.get(route)
+
+      # it returns the latest assessment date or nil if there are none
+      all_assessments.map { |assessment| assessment.fetch('completed').to_date }.max
+
+      # it returns nil if an offender can't be found
+    rescue Faraday::ResourceNotFound # 404 Not Found error
+      nil
+    end
+  end
+end
--- a/spec/services/hmpps_api/assessment_api_spec.rb
+++ b/spec/services/hmpps_api/assessment_api_spec.rb
@@ -1,0 +1,75 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+describe HmppsApi::AssessmentApi do
+  let(:api_host) { Rails.configuration.assessment_api_host }
+
+  before do
+    stub_auth_token
+  end
+
+  describe '.get_latest_oasys_date' do
+    let(:offender_no) { 'A9346AC' }
+
+    let(:stub_url) { "#{api_host}/offenders/nomisId/#{offender_no}/assessments/summary?assessmentType=LAYER_3&assessmentStatus=COMPLETE" }
+
+    context 'when there is one completed OASys assessment' do
+      # The last assessment date is the 'completed' field.
+      # The test expects it to be a date object for consistency throughout the code base
+      let(:completed_date) { '2012-12-10'.to_date }
+      let(:latest_assessment) { build(:assessment_api_response, completed: completed_date) }
+
+      before do
+        stub_request(:get, stub_url).to_return(status: 200, body: [
+          latest_assessment].to_json)
+      end
+
+      it 'checks that the API has only been called once' do
+        described_class.get_latest_oasys_date(offender_no)
+        expect(a_request(:get, stub_url)).to have_been_made.once
+      end
+
+      it 'returns the date the assessment was completed' do
+        expect(described_class.get_latest_oasys_date(offender_no)).to eq(completed_date)
+      end
+    end
+
+    context 'when there are multiple completed OASys assessments' do
+      let(:old_assessments) { build_list(:assessment_api_response, 9) }
+      let(:latest_assessment) { build(:assessment_api_response, completed: Time.zone.now) }
+      let(:all_assessments) { (old_assessments + [latest_assessment]).shuffle }
+
+      before do
+        stub_request(:get, stub_url).to_return(status: 200, body: all_assessments.to_json)
+      end
+
+      it 'returns the most recently completed assessment' do
+        expect(described_class.get_latest_oasys_date(offender_no)).to eq(Time.zone.today)
+      end
+    end
+
+    context 'when there are no assessments (it returns an empty array)' do
+      before do
+        stub_request(:get, stub_url).to_return(status: 200, body: [].to_json)
+      end
+
+      it 'returns nil' do
+        expect(described_class.get_latest_oasys_date(offender_no)).to eq(nil)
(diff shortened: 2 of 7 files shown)


app/services/hmpps_api/assessment_api.rb
```diff
@@ -0,0 +1,28 @@
...
+
+      Date.parse(returns_most_recent_assessment(get_all_assessments(route)))
+    end
+  end
+end
+
+private
+
+def get_all_assessments(route)
+  self.client.get(route, cache: false)
```
ollietreend, line 22: Remove `cache: false` (so it defaults to on)

ollietreend: Only call this once – save the first GET response as a variable.

If you want to: you can update your tests to enforce that you only perform 1 HTTP GET request.

app/services/hmpps_api/assessment_api.rb
```diff
@@ -0,0 +1,28 @@
...
+      HmppsApi::Client.new(host)
+    end
+
+    def self.get_latest_oasys_date(offender_no)
+      safe_offender_no = URI.encode_www_form_component(offender_no)
+      route = "/offenders/nomisId/#{safe_offender_no}/assessments/summary?assessmentType=LAYER_3&assessmentStatus=COMPLETE"
+
+      return nil if get_all_assessments(route).empty? || get_all_assessments(route).is_a?(Hash) # returns a 404 status offender not found
+
+      Date.parse(returns_most_recent_assessment(get_all_assessments(route)))
```
ollietreend, line 14: This could be refactored as a straightforward `if` statement.

ollietreend: @ollietreend to check what the 404 Not Found response is – because it seems a bit odd right now!

app/services/hmpps_api/assessment_api.rb
```diff
@@ -0,0 +1,28 @@
...
+private
+
+def get_all_assessments(route)
+  self.client.get(route, cache: false)
+end
+
+def returns_most_recent_assessment(assessments)
+  assessments.sort! { |a, b| b['completed'] <=> a['completed'] }
+  assessments[0]['completed']
+end
```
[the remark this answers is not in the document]
ollietreend: This is the refactor we worked on:

```ruby
response.map { |assessment|
  assessment.fetch('completed').to_date
}.max
```

and then this probably doesn't need to be in a separate private method.

spec/services/hmpps_api/assessment_api_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+           "historicStatus": "CURRENT",
+           "refAssessmentOasysScoringAlgorithmVersion": 3,
+           "assessorName": "Oliver Orange",
+           "created": "2012-11-18T13:19:59",
+           "completed": Faker::Date.backward
+            )
+        end
+
+        # Assigns one of the completed field to today's date and shuffles them
+        list_of_all_assessments[5]["completed"] = Time.zone.today
```
ollietreend, line 67: I think this could be a factory – and then it would reduce noise in this file.

`:assessment_api_response`

Thinking out loud:

```ruby
let(:old_assessments) { build_list(:assessment_api_response, 3) }
let(:latest_assessment) { build(:assessment_api_response, completed: Time.zone.now) }
let(:all_assessments) { (old_assessments + [latest_assessment]).shuffle }
# => [ {}, {}, {} ]
```

ollietreend: Useful factory which returns a `Hash` object:

`spec/factories/offenders.rb`

spec/services/hmpps_api/assessment_api_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+        expect(described_class.get_latest_oasys_date(offender_no)).to eq(nil)
+      end
+    end
+
+    context 'when the offender is not found (status 404)' do
+      before do
+        stub_request(:get, stub_url).to_return(status: 200, body: {
+          "status": 404,
+          "developerMessage": "Offender not found for NOMIS, A1857ER"
+        }.to_json)
```
ollietreend, line 67: It looks like the server actually response with a `404 Not Found` status code in this scenario, not a `200 OK`.

I know we talked about it on Friday and said that it was weird considering the response body suggests we should be receiving a `404` error. So it's actually a relief to see that it actually does the right thing and returns a sensible error code.

You can see the response status code by adding the option `-i` to your `curl` command. For example, with the non-existent offender ID `A1234ZC`:

```
$ curl -i "https://offender-dev.aks-dev-1.studio-hosting.service.justice.gov.uk/offenders/nomisId/A1234ZC/assessments/summary?assessmentType=LAYER_3&assessmentStatus=COMPLETE" --header "Content-Type: application/json" --header "Authorization: Bearer $TOKEN"
HTTP/2 404
server: nginx/1.19.2
date: Mon, 27 Sep 2021 16:58:40 GMT
content-type: application/json
vary: Accept-Encoding
request-context: appId=0d5d6f09-1d07-47f5-9d3a-f9f473544e78
x-content-type-options: nosniff
x-xss-protection: 1; mode=block
cache-control: no-cache, no-store, max-age=0, must-revalidate
pragma: no-cache
expires: 0
strict-transport-security: max-age=15724800; includeSubDomains
x-frame-options: SAMEORIGIN

{"status":404,"developerMessage":"Offender not found for NOMIS, A1234ZC"}
```

Note the first line to be returned from `curl` is:

```
HTTP/2 404
```

---

We should change the stubbed request accordingly so that it returns a `404` response.

And then we want our API client to handle that error response gracefully. I think the described behaviour of "it returns nil" is what we want.

app/services/hmpps_api/assessment_api.rb
```diff
@@ -0,0 +1,20 @@
...
+    def self.get_latest_oasys_date(offender_no)
+      safe_offender_no = URI.encode_www_form_component(offender_no)
+      route = "/offenders/nomisId/#{safe_offender_no}/assessments/summary?assessmentType=LAYER_3&assessmentStatus=COMPLETE"
+
+      all_assessments = self.client.get(route)
+
+      if all_assessments.is_a?(Array) && all_assessments.present?
+        # return the latest assessment date
+        all_assessments.map { |assessment| assessment.fetch('completed').to_date }.max
+      end
```
ollietreend, line 17: Given my previous comment about the "offender not found" scenario actually returning a 404 rather than 200 – it means that it should be possible to make this a little nicer.

Hint: so long as we get a `200 OK` response, we should always have an array. It's only when a 404 error code is returned that we receive a hash.

ollietreend (commented):
This is looking fantastic Katy.

I've spotted a couple more things which I wanted to comment for you to take a look at.

The new factory looks great. I've given some feedback about how you could use the `Faker` library a little more in that factory. However this is just an opinion and doesn't need to be followed! 🙂 The factory is also perfectly good as-is.

config/environments/production.rb
```diff
@@ -109,5 +109,7 @@
     config.session_store :cache_store, key: 'manage_pom_cases_session'
 
     config.community_api_host = ENV['COMMUNITY_API_HOST']&.strip
+
+    config.assessment_api_host = ENV['ASSESSMENT_API_HOST']&.strip
```
ollietreend, line 113: This isn't needed because it's now in your main `config/application.rb` file (which applies to every environment).

It'd be good to reset/revert all changes that have been made in this branch to the `config/environments/*` files.

PR #1879, opened 2021-12-03
Mo 809 restricted patient movement changes

mgreen-mojhub:
Ensure a Probation office is deallocated when released from Hospital back to the community. Cross reference the fromAgency code with current list of Hospital Agencies on a release movement record.

### DIFF
--- a/app/services/hmpps_api/prison_api/agencies_api.rb
+++ b/app/services/hmpps_api/prison_api/agencies_api.rb
@@ -2,18 +2,17 @@ module HmppsApi
   module PrisonApi
     class AgenciesApi
       extend PrisonApiClient
+      BASE_PATH = '/agencies/type'.freeze
       HOSPITAL_AGENCY_TYPE = 'HSHOSP'.freeze
       def self.agencies_by_type(type)
-        route = "/agencies/type/#{type}"
-        data = client.get(route, cache: true)
+        data = client.get("#{BASE_PATH}/#{type}", cache: true)
         data.map { |agency|
-          { agency_type: agency['agencyId'], description: agency['description'], active: (!!agency['active']) }
+          { agency_type: agency['agencyId'], description: agency['description'], active: agency['active'] }
         }
       end
 
       def self.agency_ids_by_type(type)
-        route = "/agencies/type/#{type}"
-        data = client.get(route, cache: true)
+        data = client.get("#{BASE_PATH}/#{type}", cache: true)
         data.map { |agency| agency['agencyId'] }
       end
     end
--- a/app/services/hmpps_api/prison_api/agencies_api.rb
+++ b/app/services/hmpps_api/prison_api/agencies_api.rb
@@ -1,0 +1,21 @@
+module HmppsApi
+  module PrisonApi
+    class AgenciesApi
+      extend PrisonApiClient
+      HOSPITAL_AGENCY_TYPE = 'HSHOSP'.freeze
+      def self.agencies_by_type(type)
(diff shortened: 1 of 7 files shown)


app/services/hmpps_api/prison_api/agencies_api.rb
```diff
@@ -0,0 +1,21 @@
+module HmppsApi
+  module PrisonApi
+    class AgenciesApi
+      extend PrisonApiClient
+      HOSPITAL_AGENCY_TYPE = 'HSHOSP'.freeze
+      def self.agencies_by_type(type)
+        route = "/agencies/type/#{type}"
```
thomasleese, line 7: I'd suggest pulling this out to a global constant which can be interpolated in to the string as it's used in two places.

app/services/hmpps_api/prison_api/agencies_api.rb
```diff
@@ -0,0 +1,21 @@
...
+        data = client.get(route, cache: true)
+        data.map { |agency|
+          { agency_type: agency['agencyId'], description: agency['description'], active: (!!agency['active']) }
```
thomasleese, line 10: It's generally unnecessary to do `!!` in Ruby - what type would `agency['active']` be?

mgreen-mojhub: This was actually recommended by the IDE, which is looking at the Rubocop params. 
The api is here https://api-dev.prison.service.justice.gov.uk/swagger-ui/#/agencies/getAgencies and the value is a boolean, so it should be fine to change

app/services/hmpps_api/prison_api/agencies_api.rb
mgreen-mojhub: Just to clarify, are you suggesting HOSPITAL_AGENCY_TYPE should be declared as a rails wide constant? Or should it be under a module for more context?

thomasleese: Yes, just inside this class.

app/services/hmpps_api/prison_api/agencies_api.rb
thomasleese: Hmm, okay, well it certainly looks odd to do `!!` here, and I'm surprised it was recommended by the IDE!

app/services/hmpps_api/prison_api/agencies_api.rb
mgreen-mojhub: Apologies, I'm still not 100% on what you are suggesting, probably down to me being dense in this instance. I could specify a true global using $HOSPITAL_AGENCY_TYPE, or even add the value to the config options. Alternatively I could add the Constant at a higher level within the module stack, such as HmppsApi::HOSPITAL_AGENCY_TYPE

If you have an example of how you would reference the constant from within the app, I should be fine to implement it.

thomasleese: Oh sorry, I'm referring to the `route = "/agencies/type/#{type}"` line rather than `HOSPITAL_AGENCY_TYPE`.

```ruby
BASE_PATH = "/agencies/type".freeze

... # and then it's used

route = URI::join(BASE_PATH, type)
```

mgreen-mojhub: Ah understood now :) I'll make the changes and push up. Thanks for the advice.

thomasleese (approved):

PR #1880, opened 2021-12-09
Mo 889 add mhi prison

mgreen-mojhub:
Morton Prison cannot currently login to MPC. The Prison needs adding to the DB and removing from the immigration list when processing transfers, as Morton is no longer immigration specific.


petergphillips (approved):
LGTM

PR #1888, opened 2022-02-01
Mo 960 update build remove setup kube auth

mgreen-mojhub:


rj-adams (approved):
Approved on recommendation - see below Slack message

Poornima Krishnasamy
@Matthew Green the build script looks good in terms of setting Auth. Give it a go and let us know.

PR #1889, opened 2022-02-02
disable cron jobs on live cluster until migration is complete. fix re…

mgreen-mojhub:
…lative path to deploy folder.


rj-adams (approved):
Approved

PR #1890, opened 2022-02-03
refs #MO-960 amend aws annotations to match MOJ policy. remove refere…

mgreen-mojhub:
…nce to live-1 and use proper subdomain for hostname.


rj-adams (approved):
Approved as requested

PR #1964, opened 2022-07-11
Adding job to push all handover dates to nDelius

SebNegahbanMoJ:
The automatic nDelius push after a new handover is calculated occasionally fails, and due to the method that we use to push dates, it is never retried.
As such, the added code is occasionally run to push all handover dates to nDelius, with the intention of ensuring data consistency.

Currently, this code is run manually via the console, but it makes more sense to have a documented job that can be directly invoked rather than manual code execution. This is not a permanent fix, and should not be treated as one, hence the lack of a cronjob. This is a temporary stop-gap until a more suitable solution is created.

### DIFF
--- a/app/jobs/repush_all_handover_dates_to_delius_job.rb
+++ b/app/jobs/repush_all_handover_dates_to_delius_job.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+
+# Sometimes the PushToDeliusDataJob during the ReallocateHandoverDateJob fails.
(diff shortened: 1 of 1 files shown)


ayqazi:
Comment from Seb for future readers:
> The two things that MPC masters is responsibility and handover dates
A lot of the handover date calculation logic is [here](https://github.com/ministryofjustice/offender-management-allocation-manager/blob/cea24c89491c6a632cb9f8a88d844731d2ef5ee0/app/services/handover_date_service.rb#L123) (or at least stems from there), with another tiny little bit on like 42 of the same file

ayqazi (approved):

PR #1966, opened 2022-07-13
Remove test_staging job from CircleCI workflow

doctorpod:
Ticket: n/a

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -374,6 +374,8 @@ workflows:
                 - main
       - deploy_production_approval:
           type: approval
+          requires:
+            - deploy_staging
           filters:
             branches:
               only: main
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -402,16 +372,8 @@ workflows:
...
-      - test_staging:
-          requires:
-            - deploy_staging
-          filters:
-            branches:
-              only: main
       - deploy_production_approval:
           type: approval
-          requires:
-            - test_staging
```
ayqazi, line 414: We still need a requires step or it'll try to get the deploy_production_approval straight away since it doesn't think anything comes first. Order is not respected in CircleCI, only requirement dependencies.

So it should be changed to:

```yaml
          requires:
            - deploy_staging
```

ayqazi (changes requested):
One change required, looks good otherwise!

PR #1968, opened 2022-07-20
Configure simplecov to be purely informational

ayqazi:
* Remove coverage test from CircleCI
* Lower minimum coverage to very low value


SebNegahbanMoJ (approved):
Agreed that - especially with such a close-knit and trustworthy team - coverage is something that we should keep an eye on but ultimately we should use our best judgement of what does and doesn't need testing, and it shouldn't dictate what we can and cannot merge

PR #1970, opened 2022-07-21
Add new help section and method of contacting support

ayqazi:
Ref: MO-1030


app/views/help/missing_cases.html.erb
```diff
@@ -0,0 +1,51 @@
...
+    <h2 class="govuk-heading-m govuk-!-margin-bottom-2">
+      Information missing on NOMIS or NDelius
+    </h2>
+    <p>
+      Make sure information about the case has been entered correctly on NOMIS.
+      For example, this might include sentence calculation dates.
+    </p>
+    <p>
+      Also, if you can use the service to allocate, check if the case is listed on
+      <a class="govuk-link govuk-link--no-visited-state" href="<%= missing_information_prison_prisoners_path(@prison_code) %>">Add missing details</a>.
```
doctorpod, line 20: Could this be a _link_to_?

app/views/errors/unauthorized.html.erb
```diff
@@ -18,7 +18,7 @@
       </li>
     </ul>
     <div class="govuk-body">
-      <%= link_to 'Contact the POM caseload service team', contact_us_path %> if you think another problem is causing this error or <%= link_to 'go to the POM caseload service’s dashboard', root_path %>.
+      <%= link_to 'Contact the POM caseload service team', '#' %> if you think another problem is causing this error or <%= link_to 'go to the POM caseload service’s dashboard', root_path %>.
```
doctorpod, line 21: What happens when user clicks _Contact the POM caseload service team_ link? Is it controlled by JS?

app/views/help/help_step2.html.erb
```diff
@@ -49,7 +49,7 @@
       <li>
         Select the Allocations – POM role for POMs. HOMDs and case admins need the Allocations – HOMD role
       </li>
-    </ul>
+    </ol>
```
doctorpod, line 52: Nice spot

doctorpod (approved):
Might also be worth considering renaming those `app/view/help/help_*` templates to remove the _help__

app/views/help/missing_cases.html.erb
ayqazi: The convention in the app is to do it this way, so I'm sticking with that. I know, it's weird, but I'd rather it all looked the same.

app/views/help/help_step2.html.erb
ayqazi: It was my editor, not me 😄

app/views/errors/unauthorized.html.erb
ayqazi: OOH nice spot, that's an error.

ayqazi:
> Might also be worth considering renaming those app/view/help/help_* templates to remove the help_

I definitely agree that these are badly named; however the names are annoyingly very tightly coupled to the code (which is RIDICULOUS as they are static pages). So to minimize the amount of work I have to do for this ticket, I'm leaving it as it is as a future refactoring task. At least the shittiness is localized to one controller.

PR #1973, opened 2022-07-22
Changing next_parole_date to prioritise Target Hearing Date

SebNegahbanMoJ:
There was no 'lower bound' to the date shown for the next parole date, it took the earliest date out of PED, TED and THD. This is somewhat correct, but it lead to situations where a second or third parole application (after previous unsuccessful applications) would show the original PED/TED for a sentence, rather than the more relevant most-recent Target Hearing Date.

As such, the next_parole_date method now returns the target hearing date if there is one, and only if there isn't will it return the PED or TED, as THD should always be given after PED/TED and therefore be a more accurate date.

### DIFF
--- a/app/models/mpc_offender.rb
+++ b/app/models/mpc_offender.rb
@@ -170,10 +170,11 @@ class MpcOffender
   end
 
   def next_parole_date
+    return target_hearing_date if target_hearing_date.present?
+
     [
       tariff_date,
-      parole_eligibility_date,
-      target_hearing_date
+      parole_eligibility_date
     ].compact.min
   end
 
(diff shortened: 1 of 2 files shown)


ayqazi:
We should really be adding tests to any code we touch that isn't tested.

OK so I like saying things like "use testing mostly as a feedback mechanism for designing. good code and focus on testing critical paths", but  that's for mostly tidy and well designed codebases. For MpcOffender, you have code that's just messy and needs testing to keep it bug-free. MpcOffender is such a codebase. So I'd prefer if it was tested until we have time to tidy it up, at which point we can start

ayqazi (changes requested):
We should really be adding tests to any code we touch that isn't tested.

OK so I like saying things like "use testing mostly as a feedback mechanism for designing. good code and focus on testing critical paths", but  that's for mostly tidy and well designed codebases. For MpcOffender, you have code that's just messy and needs testing to keep it bug-free. MpcOffender is such a codebase. So I'd prefer if tests were added for the code you changed here.

ayqazi (approved):

PR #2000, opened 2022-08-10
Use new inactivate complexity of need micro-service endpoint

doctorpod:
Ticket: https://dsdmoj.atlassian.net/browse/MO-676


mgreen-mojhub (approved):
LGTM

PR #2002, opened 2022-08-18
MO-Parole - Testing round 1 changes

SebNegahbanMoJ:
Changes #approaching_parole? to better reflect when a case would actually be counted as approaching parole
Shows parole section of prisoner profile if PED or TED are present
Adds parole section to allocation page for an offender
Hides hearing outcome field from parole section of prisoner profile if target hearing date has not yet passed
Updates text relating to #approaching_parole? to better describe what the tables will show.

### DIFF
--- a/app/views/allocation_staff/index.html.erb
+++ b/app/views/allocation_staff/index.html.erb
@@ -23,7 +23,7 @@
 <% end %>
 
 <%= render partial: 'shared/offence_info', locals: { editable_thd: true }  %>
-<% if @prisoner.tariff_date.present? || @prisoner.parole_eligibility_date.present? || @prisoner.current_parole_record.present? %>
+<% if @prisoner.display_current_parole_info? %>
   <%= render partial: 'shared/parole_info' %>
 <% end %>
 <% if @prisoner.previous_parole_records.any? %>
--- a/app/views/allocation_staff/index.html.erb
+++ b/app/views/allocation_staff/index.html.erb
@@ -23,7 +23,7 @@
 <% end %>
 
 <%= render partial: 'shared/offence_info', locals: { editable_thd: true }  %>
-<% if @prisoner.current_parole_record.present? %>
+<% if @prisoner.tariff_date.present? || @prisoner.parole_eligibility_date.present? || @prisoner.current_parole_record.present? %>
   <%= render partial: 'shared/parole_info' %>
 <% end %>
 <% if @prisoner.previous_parole_records.any? %>
--- a/app/views/allocations/show.html.erb
+++ b/app/views/allocations/show.html.erb
@@ -91,7 +91,7 @@
   </table>
 
   <%= render partial: 'shared/offence_info', locals: { editable_thd: true}  %>
-  <% if @prisoner.tariff_date.present? || @prisoner.parole_eligibility_date.present? || @prisoner.current_parole_record.present? %>
+  <% if @prisoner.display_current_parole_info? %>
     <%= render partial: 'shared/parole_info' %>
   <% end %>
   <% if @prisoner.previous_parole_records.any? %>
--- a/app/views/allocations/show.html.erb
+++ b/app/views/allocations/show.html.erb
@@ -91,6 +91,12 @@
   </table>
 
   <%= render partial: 'shared/offence_info', locals: { editable_thd: true}  %>
+  <% if @prisoner.tariff_date.present? || @prisoner.parole_eligibility_date.present? || @prisoner.current_parole_record.present? %>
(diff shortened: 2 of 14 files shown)


app/views/allocations/show.html.erb
```diff
@@ -91,6 +91,12 @@
   </table>
 
   <%= render partial: 'shared/offence_info', locals: { editable_thd: true}  %>
+  <% if @prisoner.tariff_date.present? || @prisoner.parole_eligibility_date.present? || @prisoner.current_parole_record.present? %>
```
ayqazi, line 94: If you could shove these 3 conditions into a method on MpcOffender, you could reuse it in several places instead of 3 separate conditions. Tidier views and less error-prone future modification.

app/views/allocation_staff/index.html.erb
```diff
@@ -23,7 +23,7 @@
 <% end %>
 
 <%= render partial: 'shared/offence_info', locals: { editable_thd: true }  %>
-<% if @prisoner.current_parole_record.present? %>
+<% if @prisoner.tariff_date.present? || @prisoner.parole_eligibility_date.present? || @prisoner.current_parole_record.present? %>
```
ayqazi, line 26: If you could shove these 3 conditions into a method on MpcOffender, you could reuse it in several places instead of 3 separate conditions. Tidier views and less error-prone future modification.

app/views/prisoners/show.html.erb
```diff
@@ -80,7 +80,7 @@
 
 <%= render 'prisoner_information' %>
 <%= render partial: 'shared/offence_info', locals: { editable_thd: true}  %>
-<% if @prisoner.current_parole_record.present? %>
+<% if @prisoner.tariff_date.present? || @prisoner.parole_eligibility_date.present? || @prisoner.current_parole_record.present? %>
```
ayqazi, line 83: If you could shove these 3 conditions into a method on MpcOffender, you could reuse it in several places instead of 3 separate conditions. Tidier views and less error-prone future modification.

doctorpod (approved):

app/views/shared/_parole_info.html.erb
```diff
@@ -22,13 +22,13 @@
...
     <% end %>
-    <% if @prisoner.target_hearing_date.present? %>
+    <% if @prisoner.current_parole_record&.target_hearing_date.present? %>
       <tr class="govuk-table__row">
         <td class="govuk-table__cell govuk-!-width-one-half">Target hearing date</td>
         <td id='target-hearing-date' class="govuk-table__cell table_cell__left_align  govuk-!-width-one-half"><%= format_date(@prisoner.target_hearing_date) %></td>
       </tr>
     <% end %>
-    <% if @prisoner.current_parole_record.hearing_outcome.present? %>
+    <% if @prisoner.target_hearing_date.present? && @prisoner.target_hearing_date <= Time.zone.today && @prisoner.current_parole_record&.hearing_outcome.present?%>
```
ayqazi, line 31: Maybe another method on MpcOffender, or at least a view helper? The name of that method would act as documentation to anyone reading this in the future.

ayqazi (commented):
I'm seeing a lot of chained method calls like `@prisoner.current_parole_record&.target_hearing_date.present?`

To understand why I'm a bit wary, I refer you to https://wiki.c2.com/?LawOfDemeter and https://en.wikipedia.org/wiki/Law_of_Demeter

Any possibility of wrapping these up in MpcOffender? Perhaps you could put them under a comment like `# Parole methods`, then when we come to building a `ParoleCase` composite model, we know which ones to pull into it.

I'll leave it to you, I'm just uneasy with all the chained method calls because I foresee a big refactoring headache if we ever make fundamental changes to our architecture.

(commenting, not requesting changes, as I leave the decision up to you to make changes or not)

app/views/shared/_parole_info.html.erb
```diff
@@ -1,3 +1,4 @@
+<%= parole_record = @prisoner.current_parole_record%>
```
ayqazi, line 1: Small niggle - if you don't want to actually OUTPUT parole_record, you should just do this (also note the whitespace 😸):

```
<% parole_record = @prisoner.current_parole_record %>
```

`<%=` will output the last returned/evaluated value

ayqazi (changes requested):

app/views/shared/_parole_info.html.erb
SebNegahbanMoJ: Ah, my mistake, thank you!

ayqazi (approved):

PR #2004, opened 2022-08-22
Some incremental changes to the ci pipeline

ayqazi:
* More robust parallel tests by using larger spec containers and locked parallel testing threads to the number of CPUs
* Try to make docker layer caching work on CircleCI - fail miserably
* Upgrade dependency Docker images in CI pipeline - we were previously using deprecated images
  * This includes upgrading the version of PostgreSQL we test against to 14.4, in preparation to upgrading our live environments
* Some other tweaks


.circleci/config.yml
```diff
@@ -274,7 +274,7 @@ jobs:
       - deploy:
           name: Deploy to live test
           command: |
-            sed -i -e s/:latest/:$CIRCLE_SHA1/ deploy/test/cron-*.yaml
+            # sed -i -e s/:latest/:$CIRCLE_SHA1/ deploy/test/cron-*.yaml
```
doctorpod, line 277: Did you mean to comment this out but leave it in?

ayqazi:
This warning has now disappeared from the CircleCi interface, and build times have not been affected:
image

doctorpod (approved):
LGTM - 1 comment

.circleci/config.yml
ayqazi: Good catch, thanks. Removing it

PR #2006, opened 2022-08-23
MO-Parole - Fixing responsibility calculations and adding comments + specs

SebNegahbanMoJ:
Fixes an issue where THD was being taken from the last parole application, rather than the next upcoming one

Also adds a handful of new methods to Offender and MpcOffender to get different parole records depends on state
Also adds comments to explain various methods, as parole can be quite complicated!

### DIFF
--- a/app/services/handover_date_service.rb
+++ b/app/services/handover_date_service.rb
@@ -29,7 +29,7 @@ class HandoverDateService
                                    start_date: nil, handover_date: nil,
                                    reason: :parole_mappa_2_3
 
-      elsif offender.target_hearing_date < offender.hearing_outcome_received + 12.months
+      elsif offender.next_thd.present? && offender.next_thd < offender.last_hearing_outcome_received + 12.months
         CalculatedHandoverDate.new responsibility: CalculatedHandoverDate::COMMUNITY_RESPONSIBLE,
                                    start_date: nil, handover_date: nil,
                                    reason: :thd_within_12_months_of_hearing_outcome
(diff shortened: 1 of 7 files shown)


ayqazi (approved):

app/services/handover_date_service.rb
```diff
@@ -29,7 +29,7 @@ def self.handover(raw_offender)
                                    start_date: nil, handover_date: nil,
                                    reason: :parole_mappa_2_3
 
-      elsif offender.target_hearing_date < offender.hearing_outcome_received + 12.months
+      elsif offender.next_thd.present? && offender.next_thd < offender.last_hearing_outcome_received + 12.months
```
doctorpod, line 32: Be careful with this line. I would put brackets around the left and right portions (either side of the <) and check it still works. In the console I got: 
```
irb(main):005:0> nil.present? && nil < 4+1
=> false
irb(main):006:0> (nil.present? && nil) < (4+1)
Traceback (most recent call last):
        1: from (irb):6
NoMethodError (undefined method `<' for false:FalseClass)
```

doctorpod (approved):

app/services/handover_date_service.rb
SebNegahbanMoJ: I'm not sure I understand the purpose of putting the brackets
Both sides of the && are treated as separate expressions, which is what we want in this scenario. The behaviour seen in your first example is what we're after, short circuiting and escaping a NoMethodError if `next_thd` is nil.

SebNegahbanMoJ: Yes, precisely!

PR #2007, opened 2022-08-23
Increase page width

doctorpod:
Ticket: https://dsdmoj.atlassian.net/browse/MO-503


SebNegahbanMoJ (approved):

SebNegahbanMoJ:
Approved on the basis that the current limitation of screen size seems to come from `govuk-width-container`, and overriding it seems reasonable.

PR #2027, opened 2022-09-12
Mo 1059 add 2 columns

doctorpod:
Ticket: https://dsdmoj.atlassian.net/browse/MO-1059


app/presenters/offender_with_prison_timeline_presenter.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+class OffenderWithPrisonTimelinePresenter
```
ayqazi, line 3: I've got no opinion about most of the code, seems well done.

The only comment I've got is about the OffenderWithTimelinePresenter - I'm actively trying to move us away from offender wrappers and am a little concerned we've added another one to the mix. The whole offender wrapper anti-pattern takes us away from proper OO coding standards, and I'd suggest we start moving back to proper OO.

Why not have a lazy initialized `timeline` method on MpcOffender or AllocatedOffender, which instantiates a PrisonerTimeline object when first called?

`def timeline; @prisoner_timeline ||= PrisonerTimeline.new(self); end`

Happy to discuss it further if I'm missing something that requires more explanation

ayqazi: (oh, and p.s.: you don't need the frozen string literal comment any more on modern Rubies)

app/helpers/offender_helper.rb
```diff
@@ -18,9 +18,9 @@ def pom_responsibility_label(offender)
 
   def case_owner_label(offender)
     if offender.pom_responsible?
-      'Custody'
+      'Responsible'
     elsif offender.com_responsible?
-      'Community'
```
SebNegahbanMoJ, line 23: I believe this change has an impact on the display of `_case_information.html.erb`, picture below. If this has been signed off by content, then great, otherwise it may need review on whether to either change the name of the field or to create a new method that returns 'Responsible' or 'Supporting' while leaving `case_owner_label` unchanged
image

app/presenters/offender_with_prison_timeline_presenter.rb
SebNegahbanMoJ: I'd also agree with this point. I think we should be moving away from wrappers around MpcOffender, so if we can find a way that we don't need to create a new one, that'd be grand!

doctorpod: Sounds like a plan - lets discuss when I come back to this

ayqazi (approved):
Looks good to me. @SebNegahbanMoJ - what do you think?

SebNegahbanMoJ (approved):

PR #2032, opened 2022-09-22
Mo 1059 2 cols for remerge

doctorpod:
Ticket: https://dsdmoj.atlassian.net/browse/MO-1059
This PR replaces https://github.com/ministryofjustice/offender-management-allocation-manager/pull/2027 which was reverted out of main


app/models/mpc_offender.rb
```diff
@@ -189,6 +189,11 @@ def prison_timeline
     @prison_timeline ||= HmppsApi::PrisonTimelineApi.get_prison_timeline(offender_no)
   end
 
+  # Calling a helper from a model? Required to make sorting work in unallocated view
+  def additional_information
+    ApplicationController.helpers.additional_information(self)
+  end
+
```
ayqazi, line 196: I'm not happy with this. I know you are not either.

Why not move `additional_information` into MpcOffender then call it from the helper instead of the other way around?

ayqazi (changes requested):

ayqazi (approved):

ayqazi (approved):

PR #2042, opened 2022-10-05
October gem updates

ayqazi:


spec/spec_helper.rb
```diff
@@ -16,20 +16,22 @@ def format(result)
...
-  add_group "Services", "app/services"
-
-  # Set to very low minimum - now it is simply for mostly informational purposes
-  minimum_coverage 20
-  # sometimes coverage drops between branches - don't fail in these cases
-  maximum_coverage_drop 0.5
-
-  # set merge_timeout to 30 minutes on circle:ci
-  merge_timeout ENV['MERGE_TIMEOUT'].to_i if ENV['MERGE_TIMEOUT']
+if ENV['DISABLE_COVERAGE'].blank?
```
SebNegahbanMoJ, line 19: I may be misunderstanding the purpose of this, but is it worth adding `|| !ENV['DISABLE_COVERAGE']`? Otherwise if someone were to set DISABLE_COVERAGE to false, it would still disable the coverage

doctorpod, line 19: Is it worth putting this in .env_example?

ayqazi: Let's do it.

PR #2047, opened 2022-10-12
Mo 1064 new success msg 2

doctorpod:
Ticket: https://dsdmoj.atlassian.net/browse/MO-1064


ayqazi:
@doctorpod do we need this? Or can we archive this PR?

PR #2058, opened 2022-10-28
New handover UI feature flag is now easier to use

ayqazi:
Set it once by visiting `/feature_flags/activate_new_handovers_ui?token=b00a391b-c672-4c8c-bac8-801bfddaf44d` then the new UI is activated and does not need the feature flag added to the URLs any more.


app/views/handovers/legacy_index.html.erb
```diff
@@ -2,6 +2,12 @@
   <%= render '/layouts/prison_switcher' %>
 <% end %>
 
+<% if session[:new_handovers_ui].present? %>
```
SebNegahbanMoJ, line 5: Non-blocking comment, rather than testing presence, is it not better to ensure that the value of `session[:new_handovers_ui]` is `true`? That way we can implement a true/false switch for the feature flags in future (should we wish to) rather than relying on deletion of the param

SebNegahbanMoJ (approved):

app/views/handovers/legacy_index.html.erb
ayqazi: Done.

PR #2069, opened 2022-11-03
Enhance display of blank/nil allocation data

doctorpod:
Ticket: https://dsdmoj.atlassian.net/browse/MO-1168

### DIFF
--- a/app/controllers/build_allocations_controller.rb
+++ b/app/controllers/build_allocations_controller.rb
@@ -6,6 +6,7 @@ class BuildAllocationsController < PrisonsApplicationController
   before_action :set_referrer
 
   include Wicked::Wizard
+  include OffenderHelper
 
   steps :override, :allocate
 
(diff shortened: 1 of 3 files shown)


app/controllers/build_allocations_controller.rb
```diff
@@ -6,6 +6,7 @@ class BuildAllocationsController < PrisonsApplicationController
   before_action :set_referrer
 
   include Wicked::Wizard
+  include OffenderHelper
```
ayqazi, line 9: I'm not sure about this. I've always thought of helpers for use by views only, what do you think? Happy to proceed if you feel strongly that helpers are meant to be shared between views and controllers.

ayqazi, line 9: I'm not sure about this. I've always thought of helpers for use by views only, what do you think? Happy to proceed if you feel strongly that helpers are meant to be shared between views and controllers. Or perhaps we can't follow ideal practices here because the legacy code is so bad that we can only make it less worse...

ayqazi (approved):

PR #2072, opened 2022-11-23
Update handover progress features from feedback and make the display on the lists page work

ayqazi:
Updated handover progress checklists with feedback from Tom, including a DB model change.

A big refactor of the view code with lots of common markup factored out

image
Note the different case types have different items.


bin/copy-offender-from-dev.rb
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env ruby
+
+raise 'Run it with "bin/rails runner", not on its own' unless Object.constants.include?(:Rails)
+
+if ENV['KUBERNETES_SERVICE_HOST'].present? || ENV['DATABASE_URL'].present? || ! Rails.env.development?
```
doctorpod, line 5: Nice

doctorpod (approved):

PR #2073, opened 2022-11-24
Upcoming and in-progress handover policy changes

ayqazi:
MO-1172

Upcoming handovers
Remove COM allocated date

Handovers in progress
Remove COM allocated date

How cases move through the tables
We should now only take COM responsibility date into account when determining when cases should appear on the different tables.

Cases appear on the Upcoming table 8 weeks before the COM responsible date

They move from Upcoming to In progress on the date the COM becomes responsible, even if a COM has not been allocated. If a COM is allocated early, the case should move to the In progress tab

If a case is missing COM details 48 hours after the COM responsible date, it should remain on the In progress table. However, it should also be displayed on the COM allocation overdue table. Once a case is allocated a COM, it should be removed from this table. The 48 hours condition reflects a stipulation in a service level agreement that states LDUs have this amount of time to allocate a COM.

A case should stay on the In progress table until the release date is reached.

### DIFF
--- a/app/models/calculated_handover_date.rb
+++ b/app/models/calculated_handover_date.rb
@@ -63,37 +63,32 @@ class CalculatedHandoverDate < ApplicationRecord
 
     # Visualization of the calculation:
     #
-    # cad = com allocated date (or start_date in legacy naming)
+    # h = handover date
     # d = days before handover starts that case is considered in upcoming handovers window (e.g. 56 days, or 8 weeks)
     # rd = relative-to date (defaults to "today")
     #
     # "in upcoming handover window":
     #
-    # cad - d                              cad
+    #  h - d                                h
     #   |                                   |
     #   |<----------------------------------|
     #
-    # So, if rd is (cad - d) or later, and rd is less than but not equal to cad, then the
+    # So, if rd is (h - d) or later, and rd is less than but not equal to h, then the
     # case is considered to be in the upcoming handover window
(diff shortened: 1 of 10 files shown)


app/models/calculated_handover_date.rb
```diff
@@ -63,38 +63,32 @@ def by_offender_ids(offender_id_or_ids)
...
     #
     # "in upcoming handover window":
     #
-    # cad - d                              cad
+    #  h - d                                h
     #   |                                   |
     #   |<----------------------------------|
     #
-    # So, if rd is (cad - d) or later, and rd is less than but not equal to cad, then the
+    # So, if rd is (h - d) or later, and rd is less than but not equal to cad, then the
```
SebNegahbanMoJ, line 76: Very minor but just for if you remember in a future PR, there's one instance of `cad` not being renamed to `h` here
Not worth a new commit now, but if you remember to change it in future then 👍

SebNegahbanMoJ (approved):

app/models/calculated_handover_date.rb
ayqazi: It definitely is! That was meant to be an explanation of crazy complicated logic so it has to make sense

ayqazi: fixed

SebNegahbanMoJ (approved):
resubmitting approval because it didn't get dismissed by the force push for some reason 🤷

ayqazi:
> resubmitting approval because it didn't get dismissed by the force push for some reason 🤷

yeah, we have the setting turned off where more commits invalidate an approval. I think we're grown up enough to make ask for a re-approval if we do major work on the PR

PR #2089, opened 2022-12-15
Useful auditing v2

ayqazi:
Improvements/fixes for the last PR


doctorpod (approved):

SebNegahbanMoJ (approved):
I'm not familiar with `meta` but I trust that you know what you're doing! 😁

ayqazi:
> It's a way of adding actual DB columns onto the "versions" table instead of just storing it in the YAML blob, so you can run SQL queries on it.

PR #2093, opened 2022-12-19
Continuous delivery FTW - merge to main now deploys without approval

ayqazi:
We need to stop being scared of deployment. Stop hiding behind approvals, bring any deployment issues forward and fix them instead of being afraid of them occurring.


andrewrlee (approved):
LGTM

PR #2111, opened 2023-01-18
Move hmpps OAUTH secrets to same K8s secret as the API secrets

ayqazi:
This moves them into one the same K8s Secret as the API secrets, which makes them easier to rotate by the HAAR team

Also make some options more configurable


andrewrlee (approved):
LGTM

PR #2115, opened 2023-01-23
Remove com_responsible_date from internal naming

ayqazi:
It is only a name used on the handover screens - internally we call it the handover date


ayqazi:
@doctorpod Testing Github comments being translated to Slack comments, with name references also being translated

ayqazi:
What about me? @ayqazi

doctorpod (approved):

PR #2131, opened 2023-02-07
Keyworker API unavailability is now handled by gracefully degrading

ayqazi:
The error is logged and a null keyworker is returned so the calling code can keep calm and carry on


andrewrlee (approved):
LGTM

PR #2132, opened 2023-02-07
Mo parole 2

SebNegahbanMoJ:


app/jobs/parole_data_import_job.rb
```diff
@@ -0,0 +1,62 @@
...
+  queue_as :default
+
+  def perform(date)
+    @date = date
+    imap = Net::IMAP.new('imap.gmail.com', 993, true)
+    mail = Mail.new(fetch_email(imap))
+    mail.attachments.empty? ? Rails.logger.info('No attachments found') : process_attachments(mail)
+    imap.logout
+    imap.disconnect
+  end
```
ayqazi, line 15: This should probably be in an ensure block so errors ensure logout/disconnect:
```suggestion
  ensure
    if imap
      imap.logout
      imap.disconnect
    end
  end
```

SebNegahbanMoJ:
This has been rewritten and merged in new branches, as this had fallen drastically out of date

PR #2146, opened 2023-02-27
Remove handover old code

ayqazi:
4b2d2582-d4bc-48b8-9922-c4736e1def45_text

MO-1210
MO-1211


app/controllers/feature_flags_controller.rb
```diff
@@ -1,13 +0,0 @@
-class FeatureFlagsController < ApplicationController
```
ayqazi, line 1: We may need to bring this back later but I thought I'd remove it until I spoke to you about libraries that have this kind of functionality built-in.

app/controllers/handovers_controller.rb
```diff
@@ -5,24 +5,7 @@ class HandoversController < PrisonsApplicationController
 
   layout 'handovers'
 
-  before_action :check_prerequisites_and_prepare_variables, except: :index
-  before_action :ensure_spo_user, only: :index
```
SebNegahbanMoJ, line 9: Do we definitely want to remove the `ensure_spo_user` check?

ayqazi: ensure_spo_user check was only being used for the old index page which is now deleted :+1:

SebNegahbanMoJ (approved):

PR #2148, opened 2023-02-28
Refactor MpcOffender#earliest_release_2

ayqazi:
It has been moved from `MpcOffender#earliest_release_2` to `Handover::HandoverCase#earliest_release_for_handover` so now nobody can be confused as to it's purpose.

It is still untested. Testing it at this stage is ultra-low-value since there are no changes being made to it, and tests either provide a benefit from stopping regressions when changes occur, or helping design better code when it is written in the first place.

### DIFF
--- a/app/lib/handover/handover_case.rb
+++ b/app/lib/handover/handover_case.rb
@@ -1,7 +1,7 @@
 class Handover::HandoverCase
   # we pass calculated handover date explicitly instead of finding it ourselves because: (1) its not our job to find it,
   # and (2) the CategorisedHandoverCases factories will have found them already when it initialises this class
-  def initialize(calculated_handover_date, allocated_offender)
+  def initialize(allocated_offender, calculated_handover_date)
     raise ArgumentError unless calculated_handover_date.is_a?(CalculatedHandoverDate)
     raise ArgumentError unless allocated_offender.is_a?(AllocatedOffender)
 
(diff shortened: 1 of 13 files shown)


app/lib/handover/handover_case.rb
```diff
@@ -1,10 +1,15 @@
...
   # and (2) the CategorisedHandoverCases factories will have found them already when it initialises this class
-  def initialize(calculated_handover_date, allocated_offender)
-    raise ArgumentError unless calculated_handover_date.is_a?(CalculatedHandoverDate)
+  def initialize(allocated_offender, calculated_handover_date = nil)
     raise ArgumentError unless allocated_offender.is_a?(AllocatedOffender)
 
+    if calculated_handover_date
+      raise ArgumentError unless calculated_handover_date.is_a?(CalculatedHandoverDate)
+    else
+      calculated_handover_date = CalculatedHandoverDate.find_by!(nomis_offender_id: allocated_offender.offender_no)
```
SebNegahbanMoJ, line 10: I read the descriptions in lines 34-36 as it being possible for a CalculatedHandoverDate to not exist for some offenders (due to be unable to calculate it).
With this being the case, what are the impacts of `CalculatedHandoverDate.find_by!` returning `nil` here?

ayqazi: I think I'll undo that commit because this variation of the constructor is never used.... one sec.

SebNegahbanMoJ (approved):

PR #2161, opened 2023-03-09
Add a step to rerun failed specs after CircleCI runs specs

ayqazi:
This should allow us to deploy quicker when APIs are being flaky


ayqazi:
FAILED

SebNegahbanMoJ (approved):

PR #2182, opened 2023-04-14
Add a link to 'show allocations' page to revert responsibility override

ayqazi:
NOTE: I'm deleting the feature test because it depends on a NOMIS offender existing that has not reached the handover date yet; and since the march of time will always eventually make that offender reach handover, this test will keep breaking and forcing us to find ANOTHER offender that is before handover.

I personally believe that such a test is a cause of more hardship than benefits and am deleting it.

MO-1243


ayqazi:
@doctorpod Please review and let me know if you disagree with the decision to delete the test.

ayqazi:
This might cause a rift between us but I've been actively moving away from code coverage as I don't need it to tell me what to test and what not to. I even disabled it in the CI pipeline.

Shall we discuss on Thursday?

PR #2192, opened 2023-04-27
Add handovers API

ayqazi:
Also, refactor API to remove jbuilder gem


app/controllers/api/offenders_controller.rb
```diff
@@ -1,18 +1,28 @@
...
+private
 
-    def offender_number
-      params.require(:nomis_offender_id)
-    end
+  def offender_number
+    params.require(:nomis_offender_id)
+  end
+
+  def offender_as_json(offender)
```
doctorpod, line 21: Shouldn't these keys be camel case if it's JSON?

doctorpod (approved):
Approved but with a query

app/controllers/api/offenders_controller.rb
ayqazi: That's the 'old' API that I don't want to touch until I've understood who's using it. But yes, ideally.

PR #2211, opened 2023-06-01
Add Fossey Way prison

doctorpod:
Ticket: https://dsdmoj.atlassian.net/browse/MO-1297


ayqazi (approved):
Feels like it should be an ActiveAdmin interface for users to muck with themselves... but probably doesn't happen often enough to bother with.

PR #2212, opened 2023-06-02
`enhanced_handover` boolean added to CaseInformation

ayqazi:
Is automatically filled based on value of service provider (case allocation) field

A one-line is needed to correctly fill existing entries:

```
CaseInformation.find_each { |ci| ci.update! enhanced_handover: (ci.case_allocation == NPS) }
```

MO-1296

### DIFF
--- a/db/structure.sql
+++ b/db/structure.sql
@@ -149,7 +149,8 @@ CREATE TABLE public.case_information (
     local_delivery_unit_id bigint,
     ldu_code character varying,
     com_email character varying,
-    active_vlo boolean DEFAULT false
+    active_vlo boolean DEFAULT false,
+    enhanced_handover boolean
 );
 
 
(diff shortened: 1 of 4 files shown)


db/structure.sql
```diff
@@ -592,8 +593,8 @@ ALTER SEQUENCE public.pom_details_id_seq OWNED BY public.pom_details.id;
 --
 
 CREATE TABLE public.prisons (
-    code character varying NOT NULL,
     prison_type character varying NOT NULL,
+    code character varying NOT NULL,
```
SebNegahbanMoJ, line 597: Is this a change that was missed from a previous migration?

SebNegahbanMoJ (approved):

db/structure.sql
ayqazi: 😮

ayqazi: Ah, I see what happened - it's rearranged itself.

PR #2225, opened 2023-06-20
Audit events page can now filter by tag as well as NOMS number

ayqazi:
MO-1300

### DIFF
--- a/app/controllers/manage/audit_events_controller.rb
+++ b/app/controllers/manage/audit_events_controller.rb
@@ -3,10 +3,15 @@ class Manage::AuditEventsController < PrisonsApplicationController
 
   def index
     @nomis_offender_id = params[:nomis_offender_id]
+    @tags = params.fetch(:tags, '').strip.split(/[^\w]+/)
     query = AuditEvent.order(published_at: :desc)
 
     if @nomis_offender_id.present?
       query = query.where(nomis_offender_id: @nomis_offender_id)
+    end
+
+    if @tags.present?
+      query = query.tags(*@tags)
     end
 
     @audit_events = query.page params[:page]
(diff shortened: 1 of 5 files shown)


app/controllers/manage/audit_events_controller.rb
```diff
@@ -3,12 +3,17 @@ class Manage::AuditEventsController < PrisonsApplicationController
...
   def index
     @nomis_offender_id = params[:nomis_offender_id]
+    @tags = params.fetch(:tags, '').strip.split(/[^\w]+/)
     query = AuditEvent.order(published_at: :desc)
 
     if @nomis_offender_id.present?
       query = query.where(nomis_offender_id: @nomis_offender_id)
     end
 
+    if @tags.present?
```
SebNegahbanMoJ, line 13: Before I go on my rant, I want to recognise that this is absolutely valid and correct code, and is also the right way to do this, but:

This is something that bugs me about some of the design patterns in Ruby as a language itself. My instinct when reading things like `present?` is that it would be a null/undefined safety check, which in this scenario would mean an empty string of tags would pass this check (as `''.split(xyx)` would return `[]`). I've just looked up what `present?` _actually_ does though, and it does `!blank?`. 
That's (small scale) madness to me, an array can be present (ie defined) and not be blank, but here the `present` method dictates that an empty array is essentially undefined.

#Rant over - With the way `present` works in Ruby, this is the correct way to use it. I just have an issue with how Ruby does it 😅

SebNegahbanMoJ (approved):

app/controllers/manage/audit_events_controller.rb
ayqazi: Don't use Clojure then 😆

PR #2255, opened 2023-08-09
Set up all dev-based envs (staging, test, test2) for IRSA

ayqazi:
Add serviceAccountName
Remove hard-coded AWS access keys

MO-1352


ayqazi:
NB: staging already has `serviceAccountName` that's why it doesn't appear in the diff

PR #2260, opened 2023-08-17
Mo 1280 eventify delius job

doctorpod:
Ticket: https://dsdmoj.atlassian.net/browse/MO-1280


app/lib/domain_events/event.rb
```diff
@@ -89,4 +91,13 @@ def self.json_validate!(data)
 
     JSON::Validator.validate!(@schema, data)
   end
+
+private
+
+  def presonal_reference
```
ayqazi, line 97: ARGH spelings

app/lib/domain_events/event.rb
```diff
@@ -89,4 +91,13 @@ def self.json_validate!(data)
...
+    return { 'identifiers' => [{ 'type' => 'NOMS', 'value' => noms_number }] } if noms_number
+    return { 'identifiers' => [{ 'type' => 'CRN', 'value' => crn_number }] } if crn_number
```
ayqazi, line 99: I notice you didn't add a test for this

app/jobs/process_delius_data_job.rb
```diff
@@ -11,28 +11,40 @@ class ProcessDeliusDataJob < ApplicationJob
...
     end
 
+    if identifier_type == :crn
+      nomis_offender_id = probation_record.fetch(:noms_id)
+
+      if nomis_offender_id.blank?
+        return logger.error(
+          "crn=#{identifier},job=process_delius_data_job,event=missing_offender_id|" \
+          'Probation record does not have a NOMIS ID'
+        )
```
ayqazi, line 42: That's quite confusing and not self-documenting - does this mean we're using the value we return?

Let's keep it clearer:

```
        logger.error(
          "crn=#{identifier},job=process_delius_data_job,event=missing_offender_id|" \
          'Probation record does not have a NOMIS ID'
        )
        return
```

app/lib/domain_events/handlers/tier_change_handler.rb
```diff
@@ -0,0 +1,40 @@
...
+  module Handlers
+    class TierChangeHandler
+      def handle(event)
+        return unless ENABLE_EVENT_BASED_PROBATION_CHANGE
+
+        Shoryuken::Logging.logger.info "event=domain_event_handle_start,domain_event_type=#{event.event_type},crn=#{event.crn_number}"
+        case_info = CaseInformation.find_by_crn(event.crn_number)
+
+        if case_info.nil?
+          Shoryuken::Logging.logger.error "event=domain_event_handle_failure,domain_event_type=#{event.event_type}," \
```
ayqazi, line 11: I'm wondering if I should add a `logger` to the arguments of `def handle(...)` so we don't need to keep saying `Shoryuken::Logging.logger`...

ayqazi (changes requested):
Great work as usual. Spotted a few changes needed, sorry

app/lib/domain_events/handlers/tier_change_handler.rb
ayqazi: (Not asking for a change, just thinking out loud)

ayqazi (approved):

ayqazi (approved):

ayqazi (approved):

PR #2263, opened 2023-08-23
Block API access without required role

doctorpod:
Ticket: https://dsdmoj.atlassian.net/browse/MO-1349

### DIFF
--- a/app/services/hmpps_api/oauth/token.rb
+++ b/app/services/hmpps_api/oauth/token.rb
@@ -33,8 +33,6 @@ module HmppsApi
         return false unless payload['scope'].include? scope
 
         if role && !payload.fetch('authorities', []).include?(role)
-          Rails.logger = Logger.new($stdout) if Rails.env.production?
-
           Rails.logger.error(
             "event=api_access_blocked,token_user_name=#{payload['user_name']},token_client_id=" \
             "#{payload['client_id']},missing_role=#{role}|API access blocked due to missing role"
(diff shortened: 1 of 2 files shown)


app/services/hmpps_api/oauth/token.rb
```diff
@@ -32,13 +32,15 @@ def valid_token_with_scope?(scope, role: nil)
         return false if payload['scope'].nil?
         return false unless payload['scope'].include? scope
 
-        # For the time being let this through with just a warning log. When we've gathered enough
-        # data and informed all the callers, we'll return false here to enforce the correct role
         if role && !payload.fetch('authorities', []).include?(role)
-          Rails.logger.warn(
-            "event=api_token_missing_role,token_user_name=#{payload['user_name']}," \
-            "token_client_id=#{payload['client_id']}}|API access with missing role #{role}"
+          Rails.logger = Logger.new($stdout) if Rails.env.production?
```
ayqazi, line 36: I'm very unsure about this... this will affect the logging for the entire application which does it's own "direct to Kibana" thing when running as a web service

ayqazi (changes requested):

app/services/hmpps_api/oauth/token.rb
doctorpod: I shall remove it -- it's an attempt to force the log into Sentry (it hasn't been appearing)

ayqazi: Logs don't go into Sentry, they go into Kibana

ayqazi (approved):
Assuming the logging worked?

PR #2273, opened 2023-09-04
Publish handover events in near real time

ayqazi:
By monitoring changes in NOMIS data and early allocation status changes, we can deliver handover events in near real time should handover dates change.

MO-1344


lib/tasks/trigger_ea_events.rake
```diff
@@ -5,6 +5,8 @@ require 'rake'
 namespace :trigger do
   desc 'Send an event if early allocation changed OOB (due to e.g. time passing)'
   task early_allocation_events: :environment do
+    next if ENABLE_EVENT_BASED_HANDOVER_CALCULATION
```
doctorpod, line 8: Should this be return?

doctorpod (approved):

lib/tasks/trigger_ea_events.rake
ayqazi: No, 'cos we're in a block, not a method 🖕🏾

PR #2277, opened 2023-09-06
Localstack support and fix deployment due to ECR credentials removal (move to Quay.io)

ayqazi:


app/lib/domain_events/event.rb
```diff
@@ -75,7 +75,9 @@ def self.sns_topic
     unless @sns_topic
       topic_arn = ENV.fetch('DOMAIN_EVENTS_TOPIC_ARN')
       aws_region = extract_region(topic_arn)
-      @sns_topic = Aws::SNS::Resource.new(region: aws_region).topic(topic_arn)
+      localstack_url = ENV['LOCALSTACK_URL']
+      client = Aws::SNS::Client.new(endpoint: localstack_url, region: aws_region) if localstack_url
+      @sns_topic = Aws::SNS::Resource.new(region: aws_region, client: client).topic(topic_arn)
```
andrewrlee, line 80: if the localstack_url is not set, then client will be nil and it works the same ways as before?

andrewrlee (approved):
LGTM

app/lib/domain_events/event.rb
ayqazi: Yes, it does.

PR #2288, opened 2023-09-22
Deploy minimal domain events consumer to production

ayqazi:
MO-1360

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -97,7 +97,7 @@ module OffenderManagementAllocationClient
 
     config.domain_event_handlers = {
       'offender-management.noop' => 'DomainEvents::Handlers::NoopHandler',
-      'prisoner-offender-search.prisoner.updated' => 'DomainEvents::Handlers::PrisonerUpdatedHandler',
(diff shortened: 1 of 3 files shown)


config/application.rb
```diff
@@ -97,7 +97,7 @@ class Application < Rails::Application
 
     config.domain_event_handlers = {
       'offender-management.noop' => 'DomainEvents::Handlers::NoopHandler',
-      'prisoner-offender-search.prisoner.updated' => 'DomainEvents::Handlers::PrisonerUpdatedHandler',
+      # 'prisoner-offender-search.prisoner.updated' => 'DomainEvents::Handlers::PrisonerUpdatedHandler',
```
SebNegahbanMoJ, line 100: I assume this is an intentional commenting-out?

ayqazi: Yep, I'll refer you to the commit where the change was made for the explanation

SebNegahbanMoJ (approved):
I can't claim to fully understand the config but it all looks reasonable!

ayqazi:
> I can't claim to fully understand the config but it all looks reasonable!

I copied something that already works 🤷🏾

PR #2295, opened 2023-09-27
Add exponential backoff to event consumption

doctorpod:
Ticket: https://dsdmoj.atlassian.net/browse/MO-1345

### DIFF
--- a/app/workers/domain_events_consumer.rb
+++ b/app/workers/domain_events_consumer.rb
@@ -58,14 +58,14 @@ private
     nil
   end
 
-  def log(sqs_msg, log_event_name, append: '', brief: false)
-    message_parts = [].tap do |m|
-      m << "event=domain_event_consume_#{log_event_name}"
-      m << "sqs_message_id=#{sqs_msg.message_id}" unless brief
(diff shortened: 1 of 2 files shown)


ayqazi:
Could we discuss the logging please? I had some plans to make logging more convenient for handlers, and build some system logging mixins where you pass key/value pairs and it converts them to a log-line. This seems like the right time to do it - maybe we can discuss tomorrow?

Also please get rid of `tap`, I have a passionate hatred of it and anything else involving metaprogramming.

app/workers/domain_events_consumer.rb
```diff
@@ -1,22 +1,24 @@
 class DomainEventsConsumer
   include Shoryuken::Worker
 
-  shoryuken_options queue: ENV.fetch('DOMAIN_EVENTS_SQS_QUEUE_NAME', 'undefined'), auto_delete: true
+  shoryuken_options queue: ENV.fetch('DOMAIN_EVENTS_SQS_QUEUE_NAME', 'undefined'),
+                    auto_delete: true,
+                    retry_intervals: [15.minutes, 1.hour]
```
ayqazi, line 6: So this can't go into a config file? We could always load if from a config file ourselves

ayqazi (approved):

PR #2304, opened 2023-10-04
Proper JWT decoding

ayqazi:
Decode JWTs properly - use the standard JWT library to consume the JWKS keys from the well-known endpoint and then try all of them to decode JWTs. This allows upstream to easily rotate keys - they can add the new key to the list first, start encoding with the new key, then remove the old key from the well-known endpoint.

MO-1350


.env.example
```diff
@@ -9,7 +10,7 @@ COMMUNITY_API_HOST="https://community-api-secure.test.delius.probation.hmpps.dsd
 COMPLEXITY_API_HOST="https://complexity-of-need-staging.hmpps.service.justice.gov.uk"
 ASSESSMENT_API_HOST="https://offender-dev.aks-dev-1.studio-hosting.service.justice.gov.uk"
 ASSESS_RISKS_AND_NEEDS_API_HOST="https://assess-risks-and-needs-dev.hmpps.service.justice.gov.uk"
-TIERING_API_HOST: https://hmpps-tier-dev.hmpps.service.justice.gov.uk
+TIERING_API_HOST="https://hmpps-tier-dev.hmpps.service.justice.gov.uk"
```
doctorpod, line 13: Oooops. Good spot!

doctorpod (approved):

.env.example
ayqazi: To YAML or not to YAML, that is the question

PR #2306, opened 2023-10-10
Eventification of PushPomToDeliusJob

doctorpod:
Ticket: https://dsdmoj.atlassian.net/browse/MO-1382


app/models/allocation_history.rb
```diff
@@ -175,9 +175,19 @@ def deallocate_offender(event:, event_trigger:)
...
 
-  def push_pom_to_community_api
-    if saved_change_to_primary_pom_nomis_id?
-      PushPomToDeliusJob.perform_later(self)
-    end
+  def publish_allocation_changed_event
+    return unless saved_change_to_primary_pom_nomis_id?
+
+    DomainEvents::Event.new(
+      event_type: 'offender-management.pom.changed',
```
ayqazi, line 182: You don't add `offender-management.` - the class adds that automatically.

Also, is `pom` the change here? Or did the `pom-allocation` change? If you've already agreed the event type with the Delius team, then ignore me.

spec/api/allocation_api_spec.rb
```diff
@@ -5,6 +5,10 @@
 describe 'Allocation API', vcr: { cassette_name: 'prison_api/allocation_api' } do
   let(:Authorization) { "Bearer TEST_TOKEN" }
 
+  before do
+    allow_any_instance_of(DomainEvents::Event).to receive(:publish).and_return(nil)
+  end
```
ayqazi, line 10: Should a controller spec be mocking out a dependency of the model it uses?

Or should it be mocking out the whole model itself?

I thought mocks were meant to be used to mock out the dependencies of the class being tested, not the dependencies of the dependencies of the class being tested?

I ask because we already have so much trouble with tests that mock things out 3-4 levels deep, and we have to understand 2-3 layers of the code just to understand and fix our tests, was hoping we didn't make the problem worse.

I noticed you're having to do this for many controllers simply because that's the way the code's been architected already, so I guess it's a consequence of the existing tests being a pile of shit. If that's the case, then please proceed and ignore my comment.

ayqazi (changes requested):
I'm afraid I'm not happy with the blanket `allow_any_instance_of(DomainEvents::Event).to receive(:publish).and_return(nil)` in all the tests. I assume this helps you deal with the after commit hook in AllocationHistory that publishes an event, but it also indiscriminately stubs any events that are ever raised in those controllers, and that's just too much code being ignored - we want stubs to be really specific so unexpected errors in the future aren't covered over accidentally.

Why don't you just stub out the after commit hook in AllocationHistory?

```
  allow_any_instance_of(AllocationHistory).to receive(:publish_allocation_changed_event)
```

app/models/allocation_history.rb
doctorpod: Agree. `allocation.changed` would be better

.env.example
```diff
@@ -1,17 +1,20 @@
...
+DIGITAL_PRISON_SERVICE_HOST='https://digital-dev.prison.service.justice.gov.uk'
+DPS_FRONTEND_COMPONENTS_API_HOST='https://frontend-components-dev.hmpps.service.justice.gov.uk'
+HMPPS_API_CLIENT_ID='string'
+HMPPS_API_CLIENT_SECRET='string'
 HMPPS_OAUTH_CLIENT_ID="string"
 HMPPS_OAUTH_CLIENT_SECRET="string"
-PROMETHEUS_METRICS=off
-DIGITAL_PRISON_SERVICE_HOST='https://digital-dev.prison.service.justice.gov.uk'
+KEYWORKER_API_HOST="https://keyworker-api-dev.prison.service.justice.gov.uk"
+MANAGE_POM_CASES_AND_DELIUS_HOST="https://manage-pom-cases-and-delius-dev.hmpps.service.justice.gov.uk" #
```
ayqazi, line 12: Is there meant to be a `#` at the end of this line?

public/openapi.yml
```diff
@@ -80,6 +80,56 @@ paths:
             application/json:
               schema:
                 "$ref": "#/components/schemas/Status"
+  "/api/allocation/{nomsNumber}/primary_pom":
```
ayqazi, line 83: Have you tested out the OpenAPI interfaace? http://localhost:3000/api-docs/index.html - remember to change the `defaultHost` to localhost:3000 and protocol to `http`

ayqazi (commented):
All looking good. Here's the prior art of using a rspec tag to NOT mock a call if it's mocked by default: https://github.com/ministryofjustice/offender-management-allocation-manager/blob/main/spec/rails_helper.rb#L124

PR #2313, opened 2023-10-23
Add 12m parole HO offset

doctorpod:
Ticket: https://dsdmoj.atlassian.net/browse/MO-1337


spec/lib/handover/handover_calculation_spec.rb
```diff
@@ -75,24 +75,51 @@
...
-      example 'handover date for non-open prison is 8 months before earliest release date' do
-        result = described_class.calculate_handover_date(sentence_start_date: sentence_start_date,
-                                                         earliest_release_date: Date.new(2026, 1, 1),
-                                                         is_determinate_parole: false,
-                                                         is_indeterminate: true,
-                                                         in_open_conditions: false,
-                                                         is_early_allocation: false)
-        expect(result).to eq [Date.new(2025, 5, 1), :indeterminate]
+      context 'with 12m offset enabled' do
+        before { stub_const('ENABLE_12_MONTH_PAROLE_HO_OFFSET', true) }
```
ayqazi, line 88: (Non-blocking suggestion:) What happens when this goes live and the constant is turned on across the board? Will other code be affected? Thinking it might be best to stub the constant globally in rails_helper.rb to make sure the whole codebase is compatible with the change.

ayqazi (approved):

spec/lib/handover/handover_calculation_spec.rb
doctorpod: Yes. Good spot

ayqazi (approved):

PR #2367, opened 2024-03-04
Use default docker image

doctorpod:
Ticket: https://dsdmoj.atlassian.net/browse/MO-1046 
The version specified was being depreciated and prod deploy was failing so this 2nd PR was needed to fix that


andrewrlee (approved):
LGTM

PR #2369, opened 2024-03-13
Add parole to prisoner profile

doctorpod:
Ticket: https://dsdmoj.atlassian.net/browse/MO-952


app/models/mpc_offender.rb
```diff
@@ -156,12 +157,81 @@ def responsibility_handover_date
...
+    if USE_PPUD_PAROLE_DATA
+      most_recent_parole_review&.target_hearing_date
+    elsif @offender.parole_record.present?
+      @offender.parole_record.target_hearing_date
+    end
+  end
+
+  # Returns the target hearing date for the offender's next active parole application, or nil if there isn't one.
+  # Used to exclude THDs of previous (completed/inactive) parole applications.
+  def next_thd
```
ryantk, line 172: As we are introducing a lot of new methods to MpcOffender which all revolve around parole behaviour may I suggest pulling them into a separate concern.

The pattern I like to follow is documented here: https://dev.37signals.com/vanilla-rails-is-plenty/ (scroll to Rich domain models) where you extract groups of methods which all revolve around a specific topic into a concern. I have had good success keeping the codebase tidy with this approach, even with god objects such as offender.

Just a suggestion, let me know what you think.

ryantk: It might be best to disregard my comment and let's just get this merged in so we can both have this code.

ryantk (approved):

PR #2371, opened 2024-03-14
Handle nil prev 2ndary POM

doctorpod:
Ticket: https://dsdmoj.atlassian.net/browse/MO-1464

### DIFF
--- a/app/views/case_history/allocation/_deallocate_secondary_pom.html.erb
+++ b/app/views/case_history/allocation/_deallocate_secondary_pom.html.erb
@@ -7,7 +7,9 @@
 <% else %>
   <% description = capture do %>
     Prisoner unallocated co-working POM
-    <%= deallocate_secondary_pom.previous_secondary_pom_name.titleize %> - <%= format_email(deallocate_secondary_pom.previous_secondary_pom_email) %>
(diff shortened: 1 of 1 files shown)


andrewrlee (dismissed):
LGTM

ryantk (dismissed):
LGTM

spec/features/allocate_feature_spec.rb
```diff
@@ -187,7 +187,7 @@
     expect(page).to have_content('This reason cannot be more than 175 characters')
   end
 
-  scenario 're-allocating', vcr: { cassette_name: 'prison_api/re_allocate_feature' } do
+  scenario 're-allocating', vcr: { cassette_name: 'prison_api/re_allocate_feature' }, local_only: true do
```
andrewrlee, line 190: what does the local_only mean? 
Is this removing them from the test suite on CI?

doctorpod: Yes, it means they won't run on CI, only locally. It's not clear why more and more tests are failing but it's likely because they're using dev API data which slowly changes over time, whereas locally, they use the VCR file. One solution is to include the VCR file in the repo but they can be huge

andrewrlee (approved):
LGTM

PR #2373, opened 2024-03-19
MO-1467: Add test coverage to HandoverFollowUpJob

ryantk:
Ticket: https://dsdmoj.atlassian.net/jira/software/c/projects/MO/boards/518?selectedIssue=MO-1467

Added test coverage to the HandoverFollowUpJob and then refactored the logic out of the job into models.

### DIFF
--- a/spec/jobs/handover_follow_up_job_spec.rb
+++ b/spec/jobs/handover_follow_up_job_spec.rb
@@ -1,0 +1,89 @@
+require "rails_helper"
+
+describe HandoverFollowUpJob, vcr: { cassette_name: "prison_api/handover_follow_up_email" } do
+  before { allow(CommunityMailer).to receive(:with).and_return(double("CommunityMailer", urgent_pipeline_to_community: nil)) }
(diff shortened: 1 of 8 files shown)


doctorpod (approved):
An impressive refactor: LGTM assuming you're confident the specs reflect the exact behaviour we had before the refactor

spec/jobs/handover_follow_up_job_spec.rb
```diff
@@ -0,0 +1,89 @@
+require "rails_helper"
+
+describe HandoverFollowUpJob, vcr: { cassette_name: "prison_api/handover_follow_up_email" } do
+  before { allow(CommunityMailer).to receive(:with).and_return(double("CommunityMailer", urgent_pipeline_to_community: nil)) }
+
+  around do |example|
```
doctorpod, line 6: Might be worth double checking we need these here. We may have something in the spec helper

doctorpod (approved):

PR #2386, opened 2024-04-15
Fix CPU hungry query

doctorpod:
Ticket: https://dsdmoj.atlassian.net/browse/MO-1489
To fix high CPU usage when loading certain pages
Original ticket that introduced the bug: https://dsdmoj.atlassian.net/browse/MO-947

### DIFF
--- a/app/models/offender.rb
+++ b/app/models/offender.rb
@@ -1,8 +1,6 @@
 # frozen_string_literal: true
 
 class Offender < ApplicationRecord
-  attr_reader :current_parole_review, :previous_parole_reviews
-
   has_paper_trail meta: { nomis_offender_id: :nomis_offender_id }
 
   # NOMIS offender IDs must be of the form <letter><4 numbers><2 letters> (all uppercase)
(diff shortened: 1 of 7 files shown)


ryantk (approved):

app/models/offender.rb
```diff
@@ -50,7 +48,7 @@ def handover_progress_task_completion_data
 
   # Returns the most recent parole record (can be a future parole application), regardless of activity status and outcome.
   def most_recent_parole_review
-    filtered_parole_reviews.max_by(&:sortable_date)
+    @most_recent_parole_review ||= parole_reviews.ordered_by_sortable_date.to_a.last
```
ryantk, line 51: How come you to_a the collection?

doctorpod: `.last` directly on the association was causing an error

ryantk: was it ActiveRecord::IrreversibleOrderError?

ryantk (approved):

PR #2387, opened 2024-04-16
Add review_type field to ParoleReview model

ryantk:
In order to query for ISP Recall ParoleReviews I need the review_type column from the imported data.


doctorpod (approved):
Is there any need for a script or rake task to retrospectively populate the new column?

PR #2393, opened 2024-04-29
Mo 1412 add pom tasks

doctorpod:
Ticket: https://dsdmoj.atlassian.net/browse/MO-1412


app/controllers/parole_reviews_controller.rb
```diff
@@ -0,0 +1,50 @@
...
+    end
+  end
+
+private
+
+  def load_offender
+    @offender = OffenderService.get_offender(params[:prisoner_id])
+  end
+
+  def validate_date(params)
```
ryantk, line 27: If you moved this to the model you could remove a lot of manual logic

```
class ParoleReview
  validates :hearing_outcome_received_on, presence: { message: 'Enter the full date the hearing outcome was confirmed' }
  validate :hearing_outcome_received_on_must_be_in_past

  def hearing_outcome_received_on_must_be_in_past
    if hearing_outcome_received_on.present? && hearing_outcome_received_on.past?
      errors.add(:hearing_outcome_received_on, "The date the hearing outcome was confirmed must be in the past")
    end
  end
```

As the model knows how to take in each of the date part params.

app/controllers/prison_staff_application_controller.rb
```diff
@@ -83,7 +83,7 @@ def pom_allocations_summary
       last_allocated_date: @allocations.max_by(&:primary_pom_allocated_at)&.primary_pom_allocated_at&.to_date,
       pending_handover_count: @handover_cases.upcoming.count,
       in_progress_handover_count: @handover_cases.in_progress.count,
-      pending_task_count: PomTasks.new.for_offenders(@pom.allocations).count,
+      pending_task_count: @pom.allocations.map(&:pom_tasks).flatten.count,
```
ryantk, line 86: would be nice if you could just call `@pom.pom_tasks`

app/presenters/pom_task_presenter.rb
```diff
@@ -1,6 +1,62 @@
...
+    @pom_task = pom_task
+  end
+
+  def first_name_with_ownership
+    "#{offender_first_name.capitalize}#{offender_first_name.downcase[-1] == 's' ? '\'' : '\'s'}"
+  end
+
+  def action_label
+    case type
+    when :parole_outcome_date
```
ryantk, line 20: I am seeing the repeated checking of `type`. Might be cleaner to have a PomTaskPresenter per `type` and have a factory method that returns the correct type of PomTaskPresenter for that task.

ie. ParoleOutcomeDatePomTask, EarlyAllocationDecisionPomTask, GenericPomTask

app/models/pom_task.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+class PomTask
+  attr_reader :offender_name, :offender_first_name, :prison_id, :offender_number, :type, :parole_review_id
+
+  # Full name and first name passed individually in case of situations where first name isn't easily determinable from full name.
+  def initialize(offender_full_name, offender_first_name, prison_id, offender_no, type, parole_review_id = nil)
```
ryantk, line 7: Is it worth passing in these values individually, could we pass in the offender?

app/controllers/parole_reviews_controller.rb
```diff
@@ -0,0 +1,50 @@
...
+  before_action :load_offender
+
+  def edit
+    @parole_review = ParoleReview.find_by(review_id: params[:id])
+  end
+
+  def update
+    @parole_review = ParoleReview.find_by(review_id: params[:id])
+    hearing_outcome_received_date = validate_date(params)
+    if @parole_review.errors.empty? && @parole_review.update(hearing_outcome_received_on: hearing_outcome_received_date)
```
ryantk, line 13: Also, `@parole_review.valid?` would become `@parole_review.update(parole_review_params)`

app/controllers/parole_reviews_controller.rb
ryantk: You might want conditional validation https://guides.rubyonrails.org/active_record_validations.html#on

ryantk (approved):
Spot on those changes

ryantk (approved):

PR #2399, opened 2024-05-13
Only link to allocation if it exists

doctorpod:
Ticket: https://dsdmoj.atlassian.net/browse/MO-1499


ryantk (approved):
Not your code, but I find it odd the class is "OffenderWithAllocation..." when half of the methods think there might not be an allocation! "OffenderWithAllocationSometimesPerhaps??" 😎

doctorpod:
Closed because this was deemed the wrong approach. We need to not include unallocated offenders in the approaching parole list

PR #2414, opened 2024-06-06
Refactor away offender wrapper

ryantk:
Removed `OffenderWrapper` and moved it's methods back into `MpcOffender`, and subsequently back out into more correctly named classes.

Long term this will make `MpcOffender` more easy to carve up into smaller pieces, as it's functionality is not being hidden.


app/services/handover_date_service.rb
```diff
@@ -8,143 +8,12 @@
...
-        WOMENS_CUTOFF_DATE
-      elsif welsh_offender?
-        WELSH_CUTOFF_DATE
-      elsif hub_or_private?
-        ENGLISH_PRIVATE_CUTOFF
-      else
-        ENGLISH_PUBLIC_CUTOFF
-      end
-    end
-  end
```
SebNegahbanMoJ, line 149: _Love_ to see this, I hated that wrapper

ryantk: Yes, to me it was a bit "shove this mess under the carpet" when I want all the mess front and centre in MpcOffender so I can decide how to carve it up best

app/models/calculated_handover_date.rb
```diff
@@ -5,6 +5,10 @@ class CalculatedHandoverDate < ApplicationRecord
   CUSTODY_WITH_COM = 'CustodyWithCom'
   COMMUNITY_RESPONSIBLE = 'Community'
 
+  COM_NO_HANDOVER_DATE = new responsibility: COMMUNITY_RESPONSIBLE,
+                             start_date: nil, handover_date: nil,
+                             reason: :com_responsibility
```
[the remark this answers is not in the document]
ryantk: You are correct and moving this to CalculatedHandoverDate was a bit of an after thought which didnt feel quite right. OffenderHandover is actually a great place to put it.

Thank you for throwing up this feedback rather than a LGTM!

SebNegahbanMoJ (approved):

PR #2417, opened 2024-06-12
MO-1479 New 403 error response in SAR API

zheileman:
https://dsdmoj.atlassian.net/browse/MO-1479

MPC returns 401 response when a client hits the SAR endpoint with a valid token that is missing the required role - this should be a 403 instead.

The existing 401 error has been split into two:
- Leave `401` response for missing (no bearer) or invalid tokens (invalid being anything that cannot be decoded, or expired tokens).
- New `403` response when token, despite being valid, lacks the appropriate role.

Note this only affects SAR endpoint, which has its own implementation of the `#verify_token` method. It is not an app-wise change. We might decide to do the same in all other endpoints.

### DIFF
--- a/app/services/hmpps_api/oauth/token.rb
+++ b/app/services/hmpps_api/oauth/token.rb
@@ -28,6 +28,14 @@ module HmppsApi
         @expiry_time - Time.zone.now < 20
       end
 
+      # High level check to see if we can decode the token
+      def valid?
+        payload.present?
+      rescue JWT::DecodeError => e
+        Rails.logger.error("event=api_access_blocked|#{e.message}")
+        false
+      end
+
       def valid_token_with_scope?(scope, role: nil)
         return false if payload['scope'].nil?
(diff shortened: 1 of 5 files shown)


app/services/hmpps_api/oauth/token.rb
```diff
@@ -42,7 +50,8 @@ def valid_token_with_scope?(scope, role: nil)
         end
 
         true
-      rescue JWT::DecodeError, JWT::ExpiredSignature => e
```
doctorpod, line 45: Should we be removing `JWT::ExpiredSignature`?

doctorpod (approved):
Just one query

app/services/hmpps_api/oauth/token.rb
doctorpod: 👍

ryantk (approved):

PR #2418, opened 2024-06-12
Upgrade ruby and rails

doctorpod:
Ticket: https://dsdmoj.atlassian.net/browse/MO-1496


doctorpod:
A more step by step approach will be taken

PR #2421, opened 2024-06-17
Bundle update (2nd attempt)

doctorpod:
Ticket: https://dsdmoj.atlassian.net/browse/MO-1496
- Includes Rails patch upgrade to 6.1.7.8


Gemfile.lock
```diff
@@ -587,10 +596,11 @@ GEM
       railties (>= 3.0.7)
     xpath (3.2.0)
       nokogiri (~> 1.8)
-    zeitwerk (2.6.13)
+    zeitwerk (2.6.16)
 
 PLATFORMS
-  ruby
+  arm64-darwin-21
```
zheileman, line 602: Nice, we are getting there 😄 
I think we might want to either add back `ruby` platform, or add `x86_64-darwin-21` otherwise Intel macs (like mine) will have trouble with the bundle 😛

doctorpod: Done

zheileman (approved):

PR #2425, opened 2024-06-19
Allow handover api to show when no offender

ryantk:
The handover API fails when for some reason the Offender record does not exist.

This should not happen as the primary function of this endpoint us COM/POM responsibility.


doctorpod (approved):
Does this change the API interface?

ryantk:
@doctorpod No because they were nullable previous to this

PR #2427, opened 2024-06-25
MO-1473: Ensure console changes are auditable and visible

ryantk:
Added a method for allowing devs to give a note whilst making changes in the backend console - and have those changes be made visible in the case history (allocation changes only as per ticket)


zheileman (approved):
Nice!

PR #2439, opened 2024-07-01
MO-1520 Use HMPPS CircleCI orb

zheileman:
Ticket: https://dsdmoj.atlassian.net/browse/MO-1520

MPC counterpart to the [work done in CNL](https://github.com/ministryofjustice/hmpps-complexity-of-need/pull/97), very similar approach.

The [HMPPS orb](https://circleci.com/developer/orbs/orb/ministryofjustice/hmpps) contains many useful jobs and commands to DRY the circle config and make consistent docker images, etc.

We can't use it yet to perform the actual deploys, as the orb is coupled to Helm and expects helm charts to be in a specific directory etc.

For now the orb is used to build/push the docker image. There is some config duplication that will be removed once we can also use the orb for deploys. It is not worth it trying to abstract this now if we intend to use Helm.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -212,23 +212,39 @@ jobs:
             kubectl apply -f ./deploy/"${env_name}"
 
 workflows:
-  feature_branch:
+  test_build_deploy:
     jobs:
       - install_dependencies:
-          filters: { branches: { ignore: [main, staging, preprod, test, test2] } }
+          filters: { branches: { ignore: [staging, preprod, test, test2] } }
       - fetch_env_vars:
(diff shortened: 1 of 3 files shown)


zheileman:
A successful build and deploy to staging was done in [this circle build](https://app.circleci.com/pipelines/github/ministryofjustice/offender-management-allocation-manager/13672/workflows/c81546b0-2db7-4637-b02a-2f023ac80047). I temporarily enabled the push and deploy to staging without having to merge to main, as to ascertain if the config was working.
The branch filters are now back as before.

ryantk (approved):

PR #2441, opened 2024-07-02
Added postgres ssl verification

ryantk:


root.crt
zheileman, line 1: Is this file committed by mistake? 😄 
We should not need it, as the cert bundle is downloaded in the docker build and incorporated to the final image.

ryantk: yes, good catch

zheileman (approved):

PR #2447, opened 2024-07-11
Upgrade to rails 7.1 and ruby 3.3.4

ryantk:


Gemfile
```diff
@@ -3,7 +3,7 @@ git_source(:github) { |repo| "https://github.com/#{repo}.git" }
 
 ruby File.read('.ruby-version').chomp
 
-gem 'rails', '~> 6.1.7'
+gem 'rails', '~> 7.1'
```
zheileman, line 6: We might want to be a bit more cautious here and only let bumps up to the patch version `'~> 7.1.3'`
Otherwise I think we could inadvertently update to 7.2 soon once is released 😄

zheileman (approved):
LGTM 🎉 
Just one minor suggestion, no blocker.

zheileman (approved):

PR #2457, opened 2024-07-24
Tweak ProcessDeliusDataJob audit logs

zheileman:
The `ProcessDeliusDataJob` writes to the `AuditEvent` table the before/after attributes but not just the changed ones, all of them.

This is verbose and difficult to actually see what was changed, which is necessary for a related piece of work MO-1227.

Limit the audit to only the attribute(s) that really changed.

Example:
> {"after"=>{"team_name"=>"Warrington 1"}, "before"=>{"team_name"=>"Warrington 1 test"}}

Added some additional logging to facilitate searches in Kibana.

### DIFF
--- a/app/jobs/process_delius_data_job.rb
+++ b/app/jobs/process_delius_data_job.rb
@@ -15,9 +15,9 @@ class ProcessDeliusDataJob < ApplicationJob
   # trigger_method can be :batch (default), or :event
   def perform(identifier, identifier_type: :nomis_offender_id, trigger_method: :batch)
     ApplicationRecord.transaction do
-      logger.info("#{identifier_type}=#{identifier},job=process_delius_data_job,event=started")
+      logger.info("#{identifier_type}=#{identifier},trigger_method=#{trigger_method},job=process_delius_data_job,event=started")
       import_data(identifier, identifier_type, trigger_method)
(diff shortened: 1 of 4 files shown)


app/jobs/process_delius_data_job.rb
```diff
@@ -84,20 +87,24 @@ def process_record(probation_record, nomis_offender_id, trigger_method)
...
       end
+    else
+      logger.info(
+        "nomis_offender_id=#{nomis_offender_id},trigger_method=#{trigger_method},job=process_delius_data_job,event=case_information_unchanged"
+      )
     end
   end
 
-  def map_delius_to_case_info(probation_record, case_info)
+  def map_delius_to_case_info!(probation_record, case_info)
```
SebNegahbanMoJ, line 107: Just to confirm, `tap` called in here is not idempotent right? Meaning that the supplied case_info object will actually be altered by this function, not just the return value?

zheileman: Yep correct, it is intended to be modifying the passed `case_info`. It was always like that but the previous code was a bit misleading, creating a second variable that is not needed. I've added the bang at the end of the method name to be explicit about the in-place modifications.

PR #2458, opened 2024-07-25
MO-1522 Finalise helm deploy production and cleanup

zheileman:
Ticket: https://dsdmoj.atlassian.net/browse/MO-1522

Last remaining bits in this piece of work.

Updated the circle config to deploy to production also using helm (we were already doing this for staging and preprod). Production has now switched over to use helm, including ingress.

Cleanup old production k8s resources/files.

Remove `probation` group from the allowlist.


andrewrlee (approved):
LGTM

PR #2459, opened 2024-07-25
Tidy up circle contexts

zheileman:
We are removing the following contexts:

- offender-management-shared
- offender-management-staging

And integrating their variables into the project settings ENV variables.

Additionally, `hmpps-common-vars` context is only needed in the deploy to production, so removed references where unneeded.


andrewrlee (approved):
LGTM

PR #2460, opened 2024-07-26
Add back required hmpps-common-vars context

zheileman:
It's required to fetch the IP allowlist groups as part of the helm apply step.

It's very easy to miss this one, as it will not fail. Added a few comments just in case.


andrewrlee (approved):
LGTM

PR #2462, opened 2024-08-01
MO-1477 Basic application insights setup

zheileman:
Ticket: https://dsdmoj.atlassian.net/browse/MO-1477

Add the basic scaffolding for application insights.

Note the official app insights gem we are using is archived in github and no longer maintained. There is no other suitable gem at the moment.

We can observe how it goes and if it is fit for purpose.


andrewrlee (approved):
LGTM

PR #2463, opened 2024-08-02
MO-1540 Skip unnecessary Notify calls

zheileman:
Ticket: https://dsdmoj.atlassian.net/browse/MO-1540

Any attempt to send Notify emails in envs other than production will produce this error:

`BadRequestError: Can't send to this recipient using a team-only API key`

This is because Notify has a safety measure where only certain API keys are allowed to send emails to any recipient.

We can intercept these Notify emails in envs that we know can’t send them, to avoid polluting the logs and Sentry with unnecessary errors, even if these are harmless.

We leave deliveries enabled in `staging` so team / guest list members can still receive them if needed to test things.

As part of this PR I've rationalised a bit the multiple notify env variables/names and unified observers and interceptors in one place.


zheileman:
Deployed to preprod and checked it is working as expected.

Screenshot 2024-08-02 at 09 19 36

The observers continue working just as before too. Example audit events, with the new tag:

> [ActiveJob] [ActionMailer::MailDeliveryJob] [538cd514-7ae0-4509-af0d-3bff915fec8b] event=audit_event_published,nomis_offender_id=XYZ,audit_event_id=604eb910-2e31-4354-9fee-a75341069e7e,tag=email,tag=community,tag=assign_com_less_than_10_months,tag=perform_deliveries_false

ryantk (approved):

PR #2465, opened 2024-08-02
MO-1543: Standard recall rules set aside for later

ryantk:
- Implemented standard recall rules in another class so that I can test them now but not have them run in production.
- Added nicer tests for offenderhandover


spec/domains/offender_handover_spec.rb
```diff
@@ -0,0 +1,163 @@
...
+    let(:in_open_conditions) { false }
+    let(:determinate_parole) { false }
+    let(:open_prison_rules_apply) { false }
+    let(:in_womens_prison) { false }
+    let(:sentence_start_date) { double }
+    let(:category_active_since) { double }
+    let(:prison_arrival_date) { double }
+
+    let(:offender) do
+      double(:offender,
```
zheileman, line 26: Wondering if we could use a factory for this? we might already have one.

spec/intergration/responsibility_and_handover_spec.rb
zheileman, line 1: Not new to this PR but I've noticed the folder name has a typo 😅 
`intergration` -> `integration`

zheileman (approved):
I would be lying if I said I know what is going on here but I trust you know a lot more of this, so LGTM 😄

PR #2467, opened 2024-08-02
Some more active job tweaks

zheileman:
Continuation to PR #2466.

Mails use its own internal job class (`ActionMailer::MailDeliveryJob`) that also requires the `log_arguments = false`

Fallback email details when the allocated POM is not included in the list of prison POMs.


andrewrlee (approved):
LGTM

PR #2469, opened 2024-08-07
MO-1554 Handle two additional offender events

zheileman:
Ticket: https://dsdmoj.atlassian.net/browse/MO-1554

We are trying to reduce and possibly eliminate the need for our daily job, by listening to events and triggering the `ProcessDeliusDataJob` in real time.

So far we've reduced the batch job changed records by half what it used to be, but ideally we want to make the batch job obsolete and this is kind of exploratory work to see if we can use additional events to update these records in real time and not having to wait for the daily batch job to run and update them.


andrewrlee (approved):
LGTM

PR #2471, opened 2024-08-09
MO-1550 Update govuk_notify_rails gem

zheileman:
Ticket: https://dsdmoj.atlassian.net/browse/MO-1550

Make code compatible with latest version of the `notifications-ruby-client` (which is a transitive dependency of the `govuk_notify_rails` gem), and was stopping us from updating these dependencies.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -12,10 +12,7 @@ gem 'coffee-rails', '~> 5.0'
 gem 'date_validator'
 gem 'faraday', '~> 1.10.3'
 gem 'net-http' # needed to undo a conflict with system libs
-gem 'govuk_notify_rails'
-# we need the extra is_csv parameter available in 5.2 and above
-# TODO: update code to work with latest version
-gem 'notifications-ruby-client', '>= 5.2', '< 6.0.0'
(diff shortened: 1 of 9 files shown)


Gemfile
```diff
@@ -12,10 +12,7 @@ gem 'coffee-rails', '~> 5.0'
 gem 'date_validator'
 gem 'faraday', '~> 1.10.3'
 gem 'net-http' # needed to undo a conflict with system libs
-gem 'govuk_notify_rails'
-# we need the extra is_csv parameter available in 5.2 and above
-# TODO: update code to work with latest version
-gem 'notifications-ruby-client', '>= 5.2', '< 6.0.0'
```
SebNegahbanMoJ, line 18: I assume the removal of this gem is intentional and no-longer needed with the newer notify gem?

SebNegahbanMoJ (approved):

Gemfile
zheileman: Yes, this gem is a is a transitive dependency of the other gem, thus no need to declare it here. We only had it here temporarily because we needed a specific version until we made the code changes in this PR 👍

PR #2474, opened 2024-08-14
MO-1543: Clean up sentences logic and fix bug with durations

ryantk:
- Refactored sentencing logic into domain objects
- Fixed sentence duration logic so that it adds up each terms duration in the sequence


app/domains/sentences/sentence_term.rb
```diff
@@ -0,0 +1,40 @@
...
+    @sentence_type_description = payload['sentenceTypeDescription']
+    @sentence_term_code = payload['sentenceTermCode']
+    @sentence_start_date = payload['sentenceStartDate']&.to_date
+    @start_date = payload['startDate']&.to_date
+    @years = payload['years'].to_i
+    @months = payload['months'].to_i
+    @days = payload['days'].to_i
+  end
+
+  def indeterminate? = life_sentence || sentence_type.indeterminate?
```
zheileman, line 35: I see you like this compact syntax 😝 but isn't it a bit more difficult to read?
Perhaps a bit of space between the methods would be nice?
No blocker!

zheileman (approved):
LGTM. Big refactor! 🎉

PR #2478, opened 2024-08-15
Take mappa level 0 into account for SD recalls

ryantk:
mappa_level can technically be 0 and the old code was explicitly looking for nil/1 when 0 would also work.


app/models/calculated_handover_date.rb
```diff
@@ -8,6 +8,10 @@ class CalculatedHandoverDate < ApplicationRecord
   REASONS = {
     com_responsibility: 'COM Responsibility',
     recall_case: 'Recall case',
+    recall_release_soon: 'Recall case',
```
zheileman, line 11: Do these values need to be more descriptive of their keys? 🤔
I guess depends where this is used and if it is user facing text or just internal.

zheileman (approved):

PR #2482, opened 2024-08-21
Added console debugging helpers

ryantk:
Added `debug_offender` helper to show the main important fields on an offender which is useful for debugging support queries.

Also fixed `with_audit_notes` which wasn't actually accessible from the console.


zheileman (dismissed):
🙌

zheileman (approved):

PR #2490, opened 2024-09-11
MO-1565 Remove VCR from DPS header/footer spec

zheileman:
Ticket: https://dsdmoj.atlassian.net/browse/MO-1565

We are commencing a cleanup of VCR from some of our specs to reduce its use where it really doesn't add any value.

This PR cleanups VCR from the DPS header/footer spec. That's 4 absurdly big cassettes gone.


ryantk (approved):
I like how you used a before with tag instead of context for mock_api_error/success. I will steal that for some things!

PR #2502, opened 2024-09-26
MO-1556: Performance improvements to parole cases screen

ryantk:
- Removed some duplicate work being done in the controller.
- Removed `to_a` conversion which was resource intensive and replaced with proper use of SQL as intended.


zheileman (approved):
🙌

PR #2505, opened 2024-10-09
MO-1562: ISP recalls refresh

ryantk:
Implemented the missing rule that recalls with a THD > 12 months from now and no mappa are POM responsible.

Refactored tests to read better


zheileman (approved):
🙌

zheileman (approved):

PR #2518, opened 2024-10-25
MO-1213 Filter out fine sentences

zheileman:
Ticket: https://dsdmoj.atlassian.net/browse/MO-1213

We are filtering by `legalStatus` already, however a specific type of sentence can be fines, these have their `imprisonmentStatus` set to `A_FINE`, and we want to also filter out these cases from showing up on MPC.

Note: several VCR cassettes were made invalid by this change as we remove several offenders due to the new filtering, and subsequent API calls are done without these prisoner numbers: A5709DZ, G8115UD, G8455UH and G8498UV.


SebNegahbanMoJ (approved):
It's obviously a lot of VCR/spec changes, but I also understand that even a small change to the logic and make the VCRs completely unworkable.
The changes seem reasonable to me!

PR #2522, opened 2024-10-30
Allow debug rows to show value change to nil

ryantk:
Code was falling back to old value when the new value is nil when its supposed to be nil.


zheileman (approved):
🙌

PR #2524, opened 2024-10-31
Ensure CI / local tests parity

zheileman:
Some tests were running only in local whereas they were skipped in CI.

Sadly this involves committing a pair of VCR cassettes, but soon we can start refactoring and getting rid of them.


zheileman:
Closing as we will be refactoring and removing VCR anyways so this has no real value right now, and it's only 2 tests that are not running on CI but are running on local, so no big deal.

PR #2525, opened 2024-11-01
Assorted refactorings

ryantk:


app/services/hmpps_api/prison_api/offender_api.rb
```diff
@@ -155,6 +124,32 @@ def self.filtered_offenders(offenders)
...
+            .map    { |o| o.fetch('prisonerNumber') }
+        )
+      end
+
+      # Get movement details only for those offenders who are temporarily out of prison (TAP/ROTL)
+      def self.temp_out_of_prison?(offender)
+        offender['inOutStatus'] == 'OUT' && offender['lastMovementTypeCode'] == HmppsApi::MovementType::TEMPORARY
+      end
+
+      def self.complexities_for(offender_nos, prison)
```
zheileman, line 141: Super minor thing but it might be worth naming the argument `prison` to `prison_id` because at first I was thinking it was Prison instance and in that case we could do `prison.womens?` but this is the ID instead.

Same in the method below.

zheileman (approved):
Nice!

app/services/hmpps_api/prison_api/offender_api.rb
ryantk: Good point, I'll fix that

PR #2528, opened 2024-11-05
Prison model tidying

ryantk:
Mostly around putting all allocation logic into the allocation summary class and then sharing that.
There is still some allocation logic around but I wanted to get this chunk of code merged in case of conflicts


zheileman (approved):
🥇

PR #2530, opened 2024-11-07
Tidying up sorting, filtering and pagination

zheileman:
Reorganise and avoid some code repetition.
Centralise anything related to sorting and pagination in the module, and delete duplicated inclusion of the module where it is part of the parent controller.

Also make sure the sorting by POM name in handovers list is done by full name instead of last name, as we are now showing POM name as "Forename Surname".


ryantk (approved):
Amazing

PR #2532, opened 2024-11-08
Attempt to fix code making pods OOM

ryantk:
1. Reapplied the PR that was revered (#2528)
2. I found some code which in effect was calling prisoner search API multiple times
3. Rewrote a method which was "caching" potentially huge objects to be a simple find

### DIFF
--- a/app/controllers/prisoners_controller.rb
+++ b/app/controllers/prisoners_controller.rb
@@ -5,6 +5,8 @@ class PrisonersController < PrisonsApplicationController
 
   before_action :load_all_offenders, only: [:allocated, :missing_information, :unallocated, :search]
 
+  after_action -> { MetricsService.instance.increment_search_count }, only: :search
+
   def allocated
     retrieve_latest_allocation_details
     load_summary :allocated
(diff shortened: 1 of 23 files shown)


app/controllers/prisoners_controller.rb
```diff
@@ -114,22 +121,6 @@ def load_summary(summary_type)
     @offenders = sort_and_paginate(items, default_sort: :last_name)
   end
 
-  def get_slice_for_page(offender_list, user_id)
-    offenders = []
-    user_allocations = []
-
-    offender_list.map do |offender|
-      allocation = @prison.allocation_for(offender)
```
ryantk: `@prison.allocation_for(offender)` Loads a `Prison::AllocationSummary` which is populated with `Prison#policy_offenders` - 1 call to prisoner search here

app/controllers/prisoners_controller.rb
```diff
@@ -21,11 +23,16 @@ def unallocated
 
   def search
     @q = search_term
-    offenders = SearchService.search_for_offenders(@q, @prison.policy_offenders)
```
ryantk: 1 call to prisoner search here

app/controllers/prisoners_controller.rb
```diff
@@ -21,11 +23,16 @@ def unallocated
...
-    @offenders, @user_allocations = get_slice_for_page(offenders, @current_user.staff_id)
-    MetricsService.instance.increment_search_count
 
-    render @current_user.allocations.empty? ? 'search' : 'search_global'
```
ryantk, line 28: `@current_user.allocations.empty?` Will call prisoner search again with its own `Prison::AllocationSummary`

zheileman (approved):
LGTM! 😅

PR #2535, opened 2024-11-12
chore: Bump dependencies

zheileman:
Update bundle, which solves a dependabot alert.
However refrained from updating Rails as there is a regression with `rackup` and `puma` when running the tests. Locked rspec versions temporarily for this same reason. Awaiting new version of puma shortly, that should solve the issues.

Update circle orbs, and helm charts, to latest versions.

Added elasticache prometheus monitoring.


ryantk (approved):

app/presenters/offender_with_allocation_presenter.rb
```diff
@@ -29,4 +29,8 @@ def allocation_date
       (@allocation.primary_pom_allocated_at || @allocation.updated_at)&.to_date
     end
   end
+
+  def primary_pom_allocated_at
```
zheileman, line 33: Reintroducing this method back, as it is needed to sort by this column in certain pages. Was deleted by mistake.

PR #2536, opened 2024-11-12
Reintroduce prison refactoring changes

ryantk:
Un revert my refactoring changes, this time reinstating a scope to the allocations query which I think has been causing memory spikes.


zheileman (approved):
🤞

PR #2544, opened 2024-11-15
Delete unused code and files and code tweaks

zheileman:
Nothing major, just deleting a bunch of unused classes and files, including a DB table, and minor code tweaks for improved readability.


config/environments/test.rb
```diff
@@ -14,7 +14,7 @@
   # test suite. You never need to work with it otherwise. Remember that
   # your test database is "scratch space" for the test suite and is wiped
   # and recreated between test runs. Don't rely on the data there!
-  config.cache_classes = true
+  config.cache_classes = false
```
zheileman, line 17: Without turning this false, I couldn't run the test DB migrations `RAILS_ENV=test rails db:migrate`
Doesn't seem to affect performance or anything else so leaving it.

ryantk (approved):
nice 🗑️🗑️🗑️

PR #2545, opened 2024-11-15
Add in attr_reader to make delegate work

ryantk:
I have previously moved fields into AllocatedOffender to be delegated to MpcOffender. 

The following line would not work as there is no `offender` method only `@offender`:
```
delegate :last_name, to: :offender, prefix: true
```

So I put in a `attr_reader :offender` for now


app/models/allocated_offender.rb
```diff
@@ -25,6 +25,8 @@ class AllocatedOffender
            to: :@allocation
   delegate :full_name_ordered, to: :staff_member, prefix: true
 
+  attr_reader :offender
```
zheileman, line 28: I think this is already declared in the SortableAllocation module. Shouldn't cause any issues tho even if it is dupe.

zheileman (approved):

PR #2546, opened 2024-11-19
New tests to cover handover summary

ryantk:
Starting to make the tests as end to end as possible to allow proper refactoring.
As new tests get written in this style we will look at simplifying setup where possible and improving existing
test setup methods.


zheileman (approved):
🙌

PR #2547, opened 2024-11-19
Check for allowed sorting fields

zheileman:
There are many tables in the application with sortable columns.

Compile a list of allowed ones we use, and reject anything else, as a precaution, and also to help avoid mistakes in code when referencing a field that is not part of the list.
Increased test coverage for some sorting fields where it was missing.


ryantk (approved):
Looks good!

PR #2582, opened 2025-04-01
MO-1669 Replace DB refresh with helm chart

zheileman:
Ticket: https://dsdmoj.atlassian.net/browse/MO-1669

The helm chart now has an Active Record compatible db refresh script, so we can start using it. MPC didn't have db refresh setup before.


andrewrlee (approved):
LGTM

PR #2583, opened 2025-04-02
MO-1662 Enable LDUs import in preprod and prod

zheileman:
Ticket: https://dsdmoj.atlassian.net/browse/MO-1662

DB refresh has been sorted out so we can proceed with enabling this in the remaining envs.

Currently mailbox register preprod/prod have the exact same LDU data as MPC preprod/prod.


andrewrlee (approved):
LGTM

PR #2626, opened 2025-06-18
Cleanup unused files and bump deps

zheileman:


app/models/staff_member.rb
```diff
@@ -6,7 +6,7 @@ class StaffMember
   # maybe this method shouldn't be here?
   attr_reader :staff_id, :prison
 
-  delegate :position, :position_description, :probation_officer?, :prison_officer?, to: :pom, allow_nil: true
```
zheileman, line 9: Removed delegation as `#position` method is already implemented down below (was overwriting the delegation anyways). Rubocop found this.

ryantk (approved):

PR #2655, opened 2025-09-01
MO-1759 Ensure cache is expired after add/rm POM

zheileman:
Ticket: https://dsdmoj.atlassian.net/browse/MO-1759

This came up after some testing on staging/preprod. After adding a POM, if going straight to the POM list, some times the new POM will not be there, as the list request could have been cached from before the onboarding. The POM would appear after the cache expired.

With this PR we force a cache expire upon adding or removing POMs so the list always return most up to date POMs.

Also, improved the back link on the onboarding search page.

### DIFF
--- a/app/services/hmpps_api/client.rb
+++ b/app/services/hmpps_api/client.rb
@@ -77,6 +77,11 @@ module HmppsApi
       )
     end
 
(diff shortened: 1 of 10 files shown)


andrewrlee (approved):
LGTM

PR #2658, opened 2025-09-02
Remove navigation spec vcrs

ryantk:
The navigation specs could be replaced with simpler faster view specs and negate the need for the vcr file


zheileman (approved):
🚀

PR #2663, opened 2025-09-05
MO-1764 Enable onboarding feature

zheileman:
Ticket: https://dsdmoj.atlassian.net/browse/MO-1764

Removes the feature flag code around the two entry points and also enabled the banner/badge callout.

Intended to be deployed to prod to make the onboarding journey visible (previously only in staging/preprod).


andrewrlee (approved):
lgtm

PR #2673, opened 2025-10-07
Added banner to inform user of potential postgres upgrade downtime

ryantk:
- Also fixed minor css issue where handover tasks ticks/crosses where not center aligned


justin-gardiner (approved):
LGTM

PR #2795, opened 2026-04-10
MO-1936 Fix caseload switch in non-root routes

zheileman:
Ticket: https://dsdmoj.atlassian.net/browse/MO-1936
Slack: https://mojdt.slack.com/archives/C05SDFXAVSP/p1775747103734939

The DPS header component caseload switch recently started redirecting back to the URL where the user was before changing prison. Previously it always redirected back to the root (`/`) of the service where we would reset the session and fetch again the new caseload.

This session reset was only on the root of the service so other routes would not know of the caseload change and keep the stale session with the previous caseload/prison in the URL.

This PR updates the DPS frontend components API to use the new endpoint as previous ones are deprecated. The new endpoint returns a `meta` object that includes the active caseload that we can now use to do the check and restore previous behaviour in any landing page, not just in the root.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -60,11 +60,7 @@ class ApplicationController < ActionController::Base
     return @dps_header_footer if @dps_header_footer
 
     begin
-      @dps_header_footer ||= {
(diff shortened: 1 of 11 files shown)


andrewrlee (approved):
LGTM

PR #2796, opened 2026-04-10
MO-1879 Save rosh details to case

zheileman:
Ticket: https://dsdmoj.atlassian.net/browse/MO-1879

Not yet using it for anything, just getting it from the probation record and saving it in preparation for follow-up work.

As part of this I've also cleaned up the unused 'N/A' tier code / dry tests, as tiering changes are also coming next so this is again prep work.

### DIFF
--- a/app/models/mpc_offender.rb
+++ b/app/models/mpc_offender.rb
@@ -24,7 +24,7 @@ class MpcOffender
   def responsibility_override? = @offender.responsibility.present?
 
   private def responsibility
-    # Take user-overridden responsbility if it exists
+    # Take user-overridden responsibility if it exists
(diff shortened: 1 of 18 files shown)


app/models/mpc_offender.rb
```diff
@@ -431,12 +430,7 @@ def attributes_to_archive
       complexity_level
       category_code
       category_active_since
-      first_name
```
zheileman, line 434: Trimmed some attributes from the attributes_to_archive that are not strictly needed as can be obtain by other means, and reduce stored PII. This is only for versioning.

andrewrlee (approved):
LGTM

PR #2797, opened 2026-04-13
Better handling of retryable handover email errors

zheileman:
There may be situations where the recipient email is `nil` if we've not been able to obtain the staff details (which is an API request) due to timeout / upstream issues.

However the previous retry mechanism would retry the actual email delivery, which already had the stale recipient and will not refresh it.

Observed in a sentry error:
https://ministryofjustice.sentry.io/issues/7406737924/?project=5584139

Bit of refactor to actually retry the job instead (we have dedup already so successful emails will not be re-sent) that should refresh the stale recipient.


andrewrlee (approved):
LGTM

PR #2799, opened 2026-04-17
MO-1923 Add report to help with new tiering model

zheileman:
Ticket: https://dsdmoj.atlassian.net/browse/MO-1923

New allocations report requested, that includes details required to evaluate the new tiering model changes.

Bit of DRY in other reports as we were repeating some ancillary code.


andrewrlee (approved):
LGTM

PR #2801, opened 2026-04-21
Remove noisy tier handler log trace

zheileman:
As part of the normal day to day we receive many tiering calculation events for cases that are not in our system (i.e. not OMiC eligible, etc.).

An error level trace was polluting the logs for something we don't have control or care about. Removed this logging.


AledWEvans (approved):
LGTM

PR #2802, opened 2026-04-22
MO-1942 Changes to rosh widget

zheileman:
Ticket: https://dsdmoj.atlassian.net/browse/MO-1942

Add a warning when there is rosh mismatch between nDelius and OASys.

Minor copy tweaks and layout as per design.


louisenorris (approved):
LGTM

PR #2803, opened 2026-04-23
Cleanup legacy code / small refactor

zheileman:
Cleanup or rename code related to legacy push to Delius (service used to have an API to directly write to nDelius handover dates). We don't do that anymore and instead we publish an event for consumers to deal with.

Changed default value for "manual entry" in the case information factory and update tests accordingly.


andrewrlee (approved):
LGTM

PR #2805, opened 2026-04-27
MO-1919 Add extended tiers (behind feature flag)

zheileman:
Ticket: https://dsdmoj.atlassian.net/browse/MO-1919

Add new extended tiers, along with validations, everywhere in the service where tiers are used (note recommendations still run on current rules, no changes there).

Everything is behind feature flag only showing on staging for now.

Some cleanup of unused code along the way.


app/assets/stylesheets/case_mix.scss
```diff
@@ -72,6 +72,11 @@
 .case-mix__tier_d {
   background: govuk-colour("magenta", $variant: "tint-50");
 }
+.case-mix__tier_e,
+.case-mix__tier_f,
+.case-mix__tier_g {
+  background: govuk-colour("black", $variant: "tint-80");
```
zheileman, line 78: In the future, each of these new tiers might have different colours, like the others.
Steph is looking into this.

andrewrlee (approved):
LGTM

PR #2806, opened 2026-04-27
Improve missing complexity level action buttons

zheileman:
Some improvements to the buttons:

- when case info is incomplete: single primary button "Save and continue".

- when case info is complete: two buttons. Primary "Save" (returns to missing information list) and secondary "Save and allocate" (goes to review/allocation page).


andrewrlee (approved):
LGTM

PR #2807, opened 2026-04-30
Guard against nil allocated complexity levels

zheileman:
This came out a few times in the logs but was being swallowed by the exception handler and not pushed to Sentry.

There are genuine occasions where an allocated case inactivates their CNL at a later point, so we guard against this, allowing sorting, and not blowing up the view.

I've made the sentry config more strict and should report more exceptions than before but updated the filter to not be overwhelming, we need to observe and iterate if needed.

### DIFF
--- a/config/initializers/sentry.rb
+++ b/config/initializers/sentry.rb
@@ -26,7 +26,13 @@ if sentry_dsn
 
     config.before_send = lambda do |event, hint|
       return nil if hint[:exception]&.full_message&.match?(/ApplicationInsights::TelemetryClient/)
-      return nil unless SentryCircuitBreakerService.check_within_quota
+
+      begin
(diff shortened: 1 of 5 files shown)


config/initializers/sentry.rb
```diff
@@ -8,10 +8,20 @@
     config.dsn = sentry_dsn
     config.release = ENV['BUILD_NUMBER']
     config.enable_metrics = false
+    config.rails.report_rescued_exceptions = true
```
zheileman, line 11: A lot more exceptions will be reported now but updated the exclusion list to filter the ones we don't really have control over or are just noise. I'll observe if this is a good balance or iterate.

andrewrlee (approved):
LGTM

PR #2808, opened 2026-04-30
Further sentry error handling tweaks

zheileman:
There is a bit of a mystery as some exceptions show up on logs but are not sent to Sentry, so this is for some extra guard in case is dropping events.


andrewrlee (approved):
LGTM

PR #2809, opened 2026-05-01
Further error reporting refactor/modernisation

zheileman:
Follow-up to previous PR #2808.

Move to use Rails error reporting API, with sentry as the backend.

Additional error reporting in the /500 route.


louisenorris (approved):
LGTM

PR #2810, opened 2026-05-01
Guard against nil offenders

zheileman:
Additional guard against invalid offender URLs or offenders no longer returned by the API, etc. so it gives an early 404 instead of blowing up somewhere down the line and ending on a 500.


andrewrlee (approved):
LGTM

PR #2814, opened 2026-05-06
MO-1950 Improve Alerts API error handling

zheileman:
Ticket: https://dsdmoj.atlassian.net/browse/MO-1950

Ahead of a scheduled Alerts API downtime next week, we need to ensure our service is not misleading when alerts can't be fetched. Currently we were defaulting to "None".

With this PR we will have more granular messaging.


andrewrlee (approved):
LGTM

PR #2816, opened 2026-05-08
Disable app insights track_exception

zheileman:
After much digging I think I finally found the reason why we were missing some exceptions in Sentry. They were showing up in logs but not in Sentry.

It is now clear the app insights gem we use is way outdated and at some point (potentially when upgrading to Ruby 4) it stopped being able to parse backtraces, blowing up, and stopping other error reporting (Sentry) from functioning correctly.

This PR disables exception reporting for AppInsights, leaving any other telemetry or metrics intact, and Sentry will take care of the exceptions.


andrewrlee (approved):
LGTM

PR #2818, opened 2026-05-11
MO-1956 Refactor timeline and add responsibility events

zheileman:
Ticket: https://dsdmoj.atlassian.net/browse/MO-1956

Continuation to previous PR #2815.

Adds the responsibility override created/removed events to the case history/timeline, along with 2 related missing emails.

Ensures we have creator first/last names in the paper trail versions so it can be used not only for Responsibility and VLO but others (in following PRs).

There is a big amount of tests because of the changes to the presenters subclassing from BaseHistoryPresenter (and adding previously missing specs).


andrewrlee (approved):
LGTM

PR #2819, opened 2026-05-13
MO-1956 Add key case info details to timeline

zheileman:
Ticket: https://dsdmoj.atlassian.net/browse/MO-1956

Another PR related to case history / timeline, following previous ones #2815 and #2818.

This one adds key information from the case information (currently tier, rosh and resourcing). Rosh is behind feature flag for consistency with the other work we have in flight.

Will produce create/update/destroy entries in the timeline.

Refactor and improvement of timeline related specs so they are not so brittle when adding new events.

### DIFF
--- a/app/presenters/case_information_history.rb
+++ b/app/presenters/case_information_history.rb
@@ -1,0 +1,70 @@
+# frozen_string_literal: true
+
+class CaseInformationHistory < BaseHistoryPresenter
+  DETAIL_ATTRIBUTES = %w[tier rosh_level enhanced_resourcing].freeze
+  DETAIL_LABELS = { 'tier' => 'Tier', 'rosh_level' => 'ROSH', 'enhanced_resourcing' => 'Resourcing' }.freeze
+
+  Detail = Struct.new(:label, :from_value, :to_value)
+
+  delegate :created_at, :event, to: :@version
+
+  def initialize(version)
+    super()
+    @version = version
+  end
+
+  def self.timeline_entries_for(nomis_offender_id)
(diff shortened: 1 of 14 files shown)


app/presenters/case_information_history.rb
```diff
@@ -0,0 +1,69 @@
...
+    super()
+    @version = version
+  end
+
+  def self.timeline_entries_for(nomis_offender_id)
+    PaperTrail::Version
+      .where(item_type: 'CaseInformation', nomis_offender_id:)
+      .filter_map do |version|
+        history = new(version)
+        history if history.event == 'destroy' || history.change_details.any?
```
SebNegahbanMoJ, line 24: I'm not sure if this conditional is needed given the early return of `[]` in `change_details`?
I may be misunderstanding the context, please ignore if so :)

spec/features/early_allocation_18month_feature_spec.rb
```diff
@@ -30,14 +30,18 @@
       stub_signin_spo user, [prison.code]
     end
 
-    it 'does some funky history record stuff' do
```
SebNegahbanMoJ, line 33: 😂

SebNegahbanMoJ (approved):
Approved with minor question

app/presenters/case_information_history.rb
zheileman: No worries! yep, we need this check here too, because if we remove it, `history.change_details.any?` would be `false` for destroys, and we wouldn't have the `history` item returned back to the map.

PR #2822, opened 2026-05-14
MO-1965 Cover new recall codes and other hardening

zheileman:
Ticket: https://dsdmoj.atlassian.net/browse/MO-1965

Some recall sentences were not being recognised upstream due to new codes being added recently but the `recall` flag still coming as false for these.

In the interim we are going to do the check on our side, but soonish this will be reverted back to using trusting upstream `recall` flag.

As part of this, other lists of codes have been hardened/aligned that could produce wrong handover calculations as well as improving the sentenced civil cases check.


andrewrlee (approved):
LGTM

PR #2823, opened 2026-05-15
Improve calculated handover history in debugging

zheileman:
Better calculated responsibility & handover history in the debugging page, showing the before and after values and more clearly indicating which ones changed or were unset.

Also fixed a bug with the date of the last_event in the debugging page. Adding rosh level too.


andrewrlee (approved):
LGTM

PR #2825, opened 2026-05-18
Tag word wrapping and styling fixes

zheileman:
Some minor fixes for tags to avoid unnecessary word wrapping unless explicitly done (which we need to do for 2 tags that are otherwise too lengthy).


andrewrlee (approved):
LGTM

PR #2826, opened 2026-05-18
Remove dead tasks page code

zheileman:
While working on the handovers tasklist this was confusing. It seems this is dead code, leftovers from the past, as we use a separate page now at /prisons/LEI/staff/12345/caseload/updates_required.

Removing to avoid further confusion.


andrewrlee (approved):
LGTM

PR #2828, opened 2026-05-19
Cleanup no longer needed recall check

zheileman:
Ticket: https://dsdmoj.atlassian.net/browse/MO-1967

In PR #2822, we introduced an interim solution to make sure new recall sentence types were being correctly detected as recalls in our service, because the upstream flag from prisoner-search was still using old codes.

Now this has been fixed upstream and we can rely again on the flag from the API response, instead of having to maintain the codes ourselves.


andrewrlee (approved):
LGTM

PR #2829, opened 2026-05-19
Use sentenceExpiryDate value from the API

zheileman:
This might have been true in the past but I've checked and we now receive the `sentenceExpiryDate` attribute from the upstream API payload, so we can use it.

Also added a few more attributes to the archive (showing up in the debugging page).

### DIFF
--- a/app/models/mpc_offender.rb
+++ b/app/models/mpc_offender.rb
@@ -272,6 +272,8 @@ class MpcOffender
   # Sentencing / Movements
   #
 
+  delegate :sentenced_to_additional_future_isp?, to: :sentences
+
   def sentences = @sentences ||= Offenders::Sentences.new(booking_id:)
 
   def prison_timeline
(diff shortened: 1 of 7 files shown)


app/models/mpc_offender.rb
```diff
@@ -412,14 +416,15 @@ def attributes_to_archive
...
+      sentence_expiry_date
+      actual_parole_date
       parole_eligibility_date
       tariff_date
       post_recall_release_date
       licence_expiry_date
       home_detention_curfew_actual_date
       home_detention_curfew_eligibility_date
       prison_arrival_date
-      earliest_release_date
```
zheileman, line 422: Removed this attribute from the archive because the below one (`earliest_release`) already contains the date and the type, so this one is redundant.

andrewrlee (approved):
LGTM

PR #2831, opened 2026-05-21
Upgrade docker image alpine packages

zheileman:
Updates the docker build to pull patched alpine packages in both the build and runtime stages, which should address a few Snyk reported vulns.


louisenorris (approved):
LGTM

PR #2832, opened 2026-05-21
MO-1969 Migrate SAR template

zheileman:
Ticket: https://dsdmoj.atlassian.net/browse/MO-1969
Parent: https://dsdmoj.atlassian.net/browse/MO-1960

Move existing SAR template from the `hmpps-subject-access-request-html-renderer` repo to be served by our own service.

Minor template defects fixed: using convertBoolean helper in 3 boolean attributes (was using optionalValue before) and fixing responsibility reason/reasonText column width.

Removed `ROLE_MPC_ADMIN` role from being able to access SAR endpoints.

Removed `roshStartDate` from the JSON payload as it is not used in the template.


andrewrlee (approved):
LGTM

PR #2834, opened 2026-05-26
Retry sporadic 401 API requests refreshing token

zheileman:
Although we usually have code to cope with these (mostly on queued jobs) we can retry sporadic 401 at the http client level which should be more optimal.

It also makes the tiering handler which is synchronous more resilient (I've seen a handful 401 in the logs over the past month, which is the main reason for this PR).

### DIFF
--- a/app/lib/domain_events/handlers/tier_change_handler.rb
+++ b/app/lib/domain_events/handlers/tier_change_handler.rb
@@ -15,12 +15,7 @@ module DomainEvents
         new_tier = api_tier_info[:tier][0]
         old_tier = case_info.tier
 
-        if new_tier == old_tier
-          logger.info "event=domain_event_handle_noop,domain_event_type=#{event.event_type},crn=#{event.crn_number}" \
-                      '|Tier value has not changed'
-
-          return
-        end
+        return if new_tier == old_tier
 
         case_info.tier = new_tier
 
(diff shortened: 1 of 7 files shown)


app/lib/domain_events/handlers/tier_change_handler.rb
```diff
@@ -15,12 +15,7 @@ def handle(event, logger: Shoryuken::Logging.logger)
         new_tier = api_tier_info[:tier][0]
         old_tier = case_info.tier
 
-        if new_tier == old_tier
-          logger.info "event=domain_event_handle_noop,domain_event_type=#{event.event_type},crn=#{event.crn_number}" \
```
zheileman, line 19: Removing this kind of pointless trace as we happen to receive a few thousand of these events per hour and most of them don't vary the existing tier. We do log already changes, and also start/end of the event.

config/initializers/health.rb
```diff
@@ -33,10 +33,8 @@
     name: 'tieringApi',
     get_response: -> { HmppsApi::Client.new(config.tiering_api_host).get('/health/ping', cache: false) })
   .add_check(
-    # TODO: we nil the `Authorization` header because of a bug with this service, returning a different
-    # response body when passing the Bearer token. A nil authorization will make it behave like the other services.
     name: 'assessRisksAndNeedsApi',
-    get_response: -> { HmppsApi::Client.new(config.assess_risks_and_needs_api_host).get('/health/ping', extra_headers: { 'Authorization' => nil }, cache: false) })
```
zheileman, line 39: I've seen we don't need this workaround anymore. The ping for ARNS works fine with/without auth like the other endpoints we probe.

andrewrlee (approved):
LGTM

PR #2835, opened 2026-05-28
MO-1975 Add bulk reallocation emails

zheileman:
Ticket: https://dsdmoj.atlassian.net/browse/MO-1975

First pass to the new emails sent in the bulk reallocation journey and, as agreed with content designer, skipping individual deallocation emails (we leave the allocation ones as these contain useful details per-offender).


andrewrlee (approved):
LGTM

PR #2836, opened 2026-06-01
MO-1977 Handle batch reallocation errors

zheileman:
Ticket: https://dsdmoj.atlassian.net/browse/MO-1977

Handles the edge case were, in a batch of N-cases, one case fails to reallocate for whatever reason, so that it does not stop the other cases from being reallocated successfully.

Some copy tweaks made in the confirmation page, as per content designer.


andrewrlee (approved):
LGTM

PR #2837, opened 2026-06-04
MO-1978 Handle tier v3 events/endpoints

zheileman:
Ticket: https://dsdmoj.atlassian.net/browse/MO-1978

In preparation for the new tier calculations and v3 events that we will start receiving soon.

Enabling our feat-flag (in `preprod`) allows these new tier levels to pass our validations and also to show up in the UI and be added in the missing info page.

### DIFF
--- a/app/services/hmpps_api/tiering_api.rb
+++ b/app/services/hmpps_api/tiering_api.rb
@@ -5,10 +5,8 @@ module HmppsApi
       HmppsApi::Client.new(host)
     end
 
-    def self.get_calculation(crn, calculation_id)
-      safe_crn = URI.encode_www_form_component(crn)
-      safe_calculation_id = URI.encode_www_form_component(calculation_id)
-      route = "/crn/#{safe_crn}/tier/#{safe_calculation_id}"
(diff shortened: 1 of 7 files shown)


app/services/hmpps_api/tiering_api.rb
```diff
@@ -5,10 +5,8 @@ def self.client
       HmppsApi::Client.new(host)
     end
 
-    def self.get_calculation(crn, calculation_id)
-      safe_crn = URI.encode_www_form_component(crn)
-      safe_calculation_id = URI.encode_www_form_component(calculation_id)
-      route = "/crn/#{safe_crn}/tier/#{safe_calculation_id}"
+    def self.get_calculation(crn, calculation_id, version:)
+      route = "/v#{version}/crn/#{crn}/tier/#{calculation_id}"
```
zheileman, line 9: Note up until now we were using an unversioned endpoint, they've deprecated that, however the behaviour was identical to v2, so moving from unversioned to v2 is fine. Also all endpoints, even v3, return the same content so are compatible between them.

andrewrlee (approved):
LGTM

PR #2841, opened 2026-06-17
MO-1984 Fix outside OMiC debugging page

zheileman:
Ticket: https://dsdmoj.atlassian.net/browse/MO-1984

The debugging page would not load due to an exception raised for offenders previously "known" to the service but then deemed outside OMiC policy.

Also, added early allocations debugging data and some other minor improvements.


andrewrlee (approved):
LGTM

PR #2842, opened 2026-06-19
MO-1855 Guard against stale POM notifications

zheileman:
Ticket: https://dsdmoj.atlassian.net/browse/MO-1855

Under certain circumstances (transfers) there might be a race condition between events/jobs.

In the transfer window, we don't know who the correct POM is, as the old POM no longer manages this prisoner and there is no new POM yet as the prisoner hasn't been re-allocated at the new prison.

This should be a very rare situation and not happening frequently but logging has been added to better understand the impact.


andrewrlee (approved):
LGTM

PR #2843, opened 2026-06-22
Fix error in import_with_catchup for blank history

zheileman:
There was a bug in the parole import, when we didn't have any previous history records (as these are purged based on date every time we run the import job), trying to convert a date to an array (date.to_a)

Fixed the logic and also improved the accumulation counters when multiple days are imported at once. Counters are now properly reset per-date and accumulated for the full range.


andrewrlee (approved):
LGTM

PR #2845, opened 2026-06-23
MO-1987 Removal/cleanup of POM upon bulk realloc

zheileman:
Ticket: https://dsdmoj.atlassian.net/browse/MO-1987

Perform the final cleanup and soft-delete the POM upon finalising a bulk reallocation. Deleted POMs will not show up in the inactive POMs tab, but if they still have cases, they will show up in the "attention needed" tab (as in limbo).

Additional copy tweaks.

### DIFF
--- a/app/services/nomis_user_roles_service.rb
+++ b/app/services/nomis_user_roles_service.rb
@@ -51,7 +51,7 @@ class NomisUserRolesService
       nomis_staff_id, prison.code, event_trigger: AllocationHistory::INACTIVE_POM
     )
 
-    prison.pom_details.destroy_by(nomis_staff_id:)
+    prison.pom_details.find_by(nomis_staff_id:)&.deleted!
 
     pom = HmppsApi::PrisonApi::PrisonOffenderManagerApi.list(
       prison.code, staff_id: nomis_staff_id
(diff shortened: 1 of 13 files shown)


app/services/nomis_user_roles_service.rb
```diff
@@ -51,7 +51,7 @@ def self.remove_pom(prison, nomis_staff_id)
       nomis_staff_id, prison.code, event_trigger: AllocationHistory::INACTIVE_POM
     )
 
-    prison.pom_details.destroy_by(nomis_staff_id:)
+    prison.pom_details.find_by(nomis_staff_id:)&.deleted!
```
SebNegahbanMoJ, line 54: Forgive my potential misunderstanding of the code here (and also my shaking memory of Ruby!), is this supposed to be `deleted` as opposed to `delete`? Looking at the previous code I assumed the intention is to delete/destroy the record.

zheileman: It is a status, we have an enum of different statuses, and deleted is one of them, we use it now for soft-delete the PomDetail record, so it will effectively do an `update(status: 'deleted')` 😄

SebNegahbanMoJ (approved):

PR #2847, opened 2026-06-24
MO-1979 Away from work bulk reallocation

zheileman:
Ticket: https://dsdmoj.atlassian.net/browse/MO-1979

New entry-point to the bulk reallocation. Upon marking a POM as "away from work" (internal status `inactive`), we present the bulk reallocation journey if they had any primary cases still allocated.

If SPO drop from the journey and POM still has primary cases allocated, they can re-enter the bulk reallocation fro the "Away from work" staff tab.

Refactored the profile edit page to use the form builder and fixed error handling and layout/markup.

### DIFF
--- a/app/assets/stylesheets/custom.scss
+++ b/app/assets/stylesheets/custom.scss
@@ -107,3 +107,7 @@
     margin-left: govuk-spacing(5);
   }
 }
+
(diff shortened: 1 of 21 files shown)


andrewrlee (approved):
LGTM

PR #2849, opened 2026-06-26
MO-1980 POM offboarding bulk reallocation

zheileman:
Ticket: https://dsdmoj.atlassian.net/browse/MO-1980

New entry-point to the bulk reallocation. Upon marking a POM as "No longer at $prison" (internal status `deleted`), we present confirmation and, then enter the bulk reallocation journey if they had any primary cases still allocated.

If SPO drops from the journey and POM still has primary cases allocated, they can re-enter the bulk reallocation from the "Attention needed" staff tab.

Lots of specs refactor/cleanup now that we have an almost finalised feature and we can link together some previously isolated journeys.

### DIFF
--- a/app/controllers/poms_controller.rb
+++ b/app/controllers/poms_controller.rb
@@ -3,9 +3,8 @@
 class PomsController < PrisonStaffApplicationController
   before_action :ensure_spo_user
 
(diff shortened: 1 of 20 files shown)


andrewrlee (approved):
LGTM

PR #2850, opened 2026-06-29
MO-1986 Add allocations with responsibility report

zheileman:
Ticket: https://dsdmoj.atlassian.net/browse/MO-1986

This is a similar report to another we already had, but this one will not filter out non-responsible, and also adds a few new columns.


andrewrlee (approved):
LGTM

PR #2851, opened 2026-06-29
MO-1989 Handle `tier.calculation.changed` event

zheileman:
Ticket: https://dsdmoj.atlassian.net/browse/MO-1989

We are switching to this event that is more appropriate for us as we only care about changes in tier, whereas previous event is too noisy as it is triggered even if the end result is the same (no tier level change).

Improved some logging.


andrewrlee (approved):
LGTM

PR #2852, opened 2026-06-29
Merge two bulk reallocation feature flags into one

zheileman:
Just a bit of tidying. Functionally is the same and it aligns with the intended future release of all the journeys at once.


andrewrlee (approved):
LGTM

PR #2853, opened 2026-06-30
Minor improvements to the early alloc email job

zheileman:
Scope allocations by active, as we must have a `primary_pom_nomis_id` present down the line in order to obtain the POM from the API.

Do not call unnecessary APIs to fetch data we don't need in this job (complexities, categories, etc.)

Move the prison transfer check to the point where it actually skips the email if we had sent one.


andrewrlee (approved):
LGTM

PR #2855, opened 2026-07-01
MO-1992 Handle reallocations to same co-worker POM

zheileman:
Ticket: https://dsdmoj.atlassian.net/browse/MO-1992

A POM can't be both the primary and secondary (co-worker) on the same case. There was an edge case where it could be possible on the bulk reallocation to select a target POM that had already some of the cases being reallocated as co-worker, thus giving an error.

Agreed with UCD to present these cases but disable the checkbox and also place a copy in the row to say the target POM is the co-worker on that case.


AledWEvans (approved):
LGTM had a look, not rb dev so always might miss something but your tests seem to cover a lot so please double check on dev when deployed!

PR #2856, opened 2026-07-02
Add additional audit trail to DB records

zheileman:
This is an ongoing tasks and will probably add some more in the future, as it is helpful for the debugging page / timeline and understanding certain actions in the service.

Also, added a missing OMIC gate check in the `review_case_details` page, and DRY code.


andrewrlee (approved):
LGTM

PR #2860, opened 2026-07-10
MO-1945 Trigger handover recalculations via events

zheileman:
Ticket: https://dsdmoj.atlassian.net/browse/MO-1945

Work related to MO-1996 and MO-1993.

We need more frequency with regards to handover recalculations where, in worst case scenario it could take up to 24h for the sweep job to run and catch these.

We can use sentence events to decide if we should enqueue a recalculation. These are debounced similarly to how we did for Delius imports but with a way longer debounce window (1h) to give plenty of time to avoid flip-flopping sentences etc.


andrewrlee (approved):
LGTM

PR #2869, opened 2026-07-20
MO-1945 Skip recalculation if no local case exists

zheileman:
Ticket: https://dsdmoj.atlassian.net/browse/MO-1945

Small follow-up refinement to PR #2860.

If we don't have yet a local case info (probation case) there is no point in attempting to run the handover calculations, and will fail as some data will not be present.

The nightly job already filters these cases but the events can be about cases not yet imported into our system, so this is needed.


andrewrlee (approved):
LGTM

PR #2870, opened 2026-07-20
Cleanup stint data after offender release

zheileman:
We had some leftovers not needed to keep upon offender release, and in fact could cause trouble under certain circumstances if the same offender re-enters custody (like the responsibility override, that would be preserved).


andrewrlee (approved):
LGTM

PR #2871, opened 2026-07-22
MO-2005 Allow blank rosh case reallocations

zheileman:
Ticket: https://dsdmoj.atlassian.net/browse/MO-2005

When a case is re-allocated to a different POM, sometimes the RoSH has been updated to a blank value. Therefore we cannot make a recommendation if the case is not Tier A.

The content on several screens have been updated as per UCD to reflect this lack of recommendation, as well as forcing going through the override screen for these cases.


andrewrlee (approved):
LGTM

PR #2873, opened 2026-07-22
Add additional contextual tags to the audit

zheileman:
Allows for better filtering in the debugging page.

Particularly useful for the handover recalculation as it can be triggered via several methods and it follows the pattern we have in other places like delius import.


andrewrlee (approved):
LGTM

PR #2874, opened 2026-07-23
Handle delete error when removing stale LDUs

zheileman:
`destroy_all` on stale LDUs fails if any have associated `case_information` records, which can happen after prod -> preprod DB resets.

Destroy individually and rescue the exception so the import continues and logs the skipped codes.

This only happens rarely on desync'd preprod databases (as a consequence of the db-refresh job) and self-heal on next import.


andrewrlee (approved):
LGTM

PR #2877, opened 2026-07-23
MO-2007 Fetch authoritative tier from Tier API

zheileman:
Ticket: https://dsdmoj.atlassian.net/browse/MO-2007
Slack convo: https://mojdt.slack.com/archives/C04J513PDLM/p1782739944391899

When a new case is imported from nDelius, the tier from the probation record may be stale or correspond to the wrong tiering version (v2 vs v3, controlled by the `new_tiers` feature flag). Subsequent tier updates arrive via domain events, but the initial value could be incorrect (particularly post 1st Sep).

Now, after saving a new CaseInformation record, we enqueue a tiny job to retrieve the authoritative tier. The probation-record tier is kept as a fallback. Further updates are handled via events.

Also added a simple rake task to do a global refresh of all case tiers if needed. This can run on a cronjob closer to the actual transition date, or manually.


andrewrlee (approved):
LGTM

PR #2879, opened 2026-07-24
MO-2003 Skip bulk reallocation interstitial page

zheileman:
Ticket: https://dsdmoj.atlassian.net/browse/MO-2003

As per UCD, when using the reallocate cases link from either Attention needed or Away from work tabs, take the user to the "Choose a POM" screen directly (and skip the Reallocate X cases screen).

Keep the interstitial when updating the POM status through the edit profile page.

Also added a missing hint text as per https://dsdmoj.atlassian.net/browse/MO-2002
And a copy change as per https://dsdmoj.atlassian.net/browse/MO-2015


andrewrlee (approved):
LGTM

PR #2880, opened 2026-07-24
MO-2014 Only send bulk realloc summary emails

zheileman:
Ticket: https://dsdmoj.atlassian.net/browse/MO-2014

For each case that was reallocated in the bulk reallocation journey, the service sent an individual reallocation email. It also sent an email summarising all the cases that had been reallocated.

As per UCD, only the summary email will be sent, skipping the individual emails.


andrewrlee (approved):
LGTM

PR #2884, opened 2026-07-28
MO-2011 Select all cases and table tweaks

zheileman:
Ticket: https://dsdmoj.atlassian.net/browse/MO-2011

As per UCD review, added another "Select all" checkbox toggle to the top of the table (keeping the bottom one), and also do not disable continue button.

Changed default sorting column. Some styling fixes.

Copy updated as per https://dsdmoj.atlassian.net/browse/MO-2009


andrewrlee (approved):
LGTM

PR #2885, opened 2026-07-29
MO-2017 Away from work table empty state

zheileman:
Ticket: https://dsdmoj.atlassian.net/browse/MO-2017

If no POMs are away from work, we remove the table and display the empty state message 'No POMs away from work'.

Also, small markup and unused css class cleanup.


andrewrlee (approved):
LGTM

PR #2886, opened 2026-07-29
MO-2018 Show ghost POMs in attention needed tab

zheileman:
Ticket: https://dsdmoj.atlassian.net/browse/MO-2018

POMs who have active case allocations but no corresponding `PomDetail` record in our database are invisible to the "Attention needed" tab on the staff page. This creates a limbo state where cases are stuck and the SPO has no clear path to resolve them.

This PR extends the detection of limbo POMs to these ghosts so they also appear in the list.

DRY the handling of PomDetail, however there is a further ticket in the backlog to actually stop pulling/creating POM records automatically (forward-populate), that will clean up much of that code:
https://dsdmoj.atlassian.net/browse/MO-1892

### DIFF
--- a/app/models/pom_detail.rb
+++ b/app/models/pom_detail.rb
@@ -23,6 +23,18 @@ class PomDetail < ApplicationRecord
 
   belongs_to :prison, foreign_key: :prison_code, inverse_of: :pom_details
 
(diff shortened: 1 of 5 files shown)


andrewrlee (approved):
LGTM

PR #2887, opened 2026-07-30
MO-2021 Filter soft-deleted from onboarding search

zheileman:
Ticket: https://dsdmoj.atlassian.net/browse/MO-2021

Ensures soft-deleted POMs whose NOMIS role hasn't yet been expired (e.g. still awaiting case reallocation) are excluded from the onboarding search results, preventing them from being re-added mid-removal.

Once fully removed (role expired), they reappear in search results and can be re-onboarded.


AledWEvans (approved):
LGTM

PR #2889, opened 2026-07-31
MO-2025 Fix conditional message in Notify emails

zheileman:
Ticket: https://dsdmoj.atlassian.net/browse/MO-2025

In the bulk reallocation summary emails there is an optional SPO message. The template needs to conditionally show a sentence depending whether there is message or not.


dave-virden (approved):
LGTM

PR #2891, opened 2026-07-31
MO-2019 Update limbo cases report

zheileman:
Ticket: https://dsdmoj.atlassian.net/browse/MO-2019

This will now also report on ghost POMs, if any, and will structure the output in a nicer way.

Basically allow us to know which prisons will have the "attention needed" tab visible and the number of primary cases in limbo.


AledWEvans (approved):
LGTM I've approved this because it's an internal report, please test throughly

PR #2892, opened 2026-07-31
Extend audit to record destroys

zheileman:
On destroy, logs the full record attributes as "before", with an empty "after", instead of silently skipping (although we still had papertrail for these).

Concern now owns the action tag and appends 'changed' or 'destroyed' to the tags automatically, so models only declare identity tags.

DRY the tests.


nishanthmahasamudram (approved):
LGTM

PR #2896, opened 2026-08-04
Scale down additional targets in non-prod envs

zheileman:
During the night (and weekends) we scale down already the main application pods, but with a recently released improvement to the [Helm chart](https://github.com/ministryofjustice/hmpps-helm-charts/pull/261), we can also scale down additional deployments like jobs processor.

Move the process movements cronjob schedule a few minutes in staging/preprod so it triggers after the scaling up.


AledWEvans (approved):
LGTM - deploy and see if it runs

PR #2898, opened 2026-08-05
Scheduled downtime tweaks

zheileman:
Follow-up to PR #2896.

Add prometheus exporter to the scheduled downtime. DRY config.


AledWEvans (approved):
LGTM

PR #2899, opened 2026-08-05
MO-2034 Remove tier calculation API call

zheileman:
Ticket: https://dsdmoj.atlassian.net/browse/MO-2034

Confirmed with Tiering team we can stop using the /calculation endpoint and just use the /tier endpoint, which also has the advantage of consolidate and simplify our code.

Actual behaviour is the same, but centralised in TierUpdateService that both, the event handler, and the job, call for consistency.

Slack convo:
https://mojdt.slack.com/archives/C04J513PDLM/p1782739944391899


dave-virden (approved):
LGTM

PR #2900, opened 2026-08-05
Clarify how to override the downtime schedule

zheileman:
Making it clear by default is the deploy action who determines the schedule values and set them when executing helm upgrade.

The previous `shutdown` override had not effect because of this.

Leaving the autogenerated deterministic values from the deploy action for now.


andrewrlee (approved):
LGTM

PR #2901, opened 2026-08-07
MO-1892 Stop forward populating POMs - part 1

zheileman:
Ticket: https://dsdmoj.atlassian.net/browse/MO-1892

First part of a wider piece of work to stop forward populating local `PomDetail` records whenever we see a new POM coming back from NOMIS (as we have now an in-service onboarding feature).

Big majority of changed files are just specs that need to use a new `stub_onboarded_poms` helper (as we are not autocreating DB records anymore).

A follow-up PR will continue the work.


andrewrlee (approved):
LGTM

PR #2904, opened 2026-08-10
MO-1892 Cleanup task for auto-created pom details

zheileman:
Ticket: https://dsdmoj.atlassian.net/browse/MO-1892

Follow-up to previous PRs #2901 and #2902.

Creates a task that will cleanup forward-populated POMs that have been auto-created, their working pattern remains at 0.0, and their role in NOMIS is no longer returned.

Also a few minor tweaks to clarify audit tags.

### DIFF
--- a/app/models/concerns/auditable.rb
+++ b/app/models/concerns/auditable.rb
@@ -24,11 +24,9 @@ private
       end
     end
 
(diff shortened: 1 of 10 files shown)


AledWEvans (approved):
LGTM

PR #2907, opened 2026-08-11
Improve admin audit search page

zheileman:
The audit page was veeery slow, particularly on first load, because it had to do a full count. Removed the count in the queries as we don't really care for that (we keep next/previous links).

Added index to username column, and ability to filter by this in the search.

Other minor styling tweaks.


AledWEvans (approved):
LGTM

PR #2908, opened 2026-08-11
MO-2042 Include soft-deleted in POM onboard search

zheileman:
Ticket: https://dsdmoj.atlassian.net/browse/MO-2042

There are scenarios where POMs marked as deleted in our service still return the NOMIS role, and the onboarding search was filtering this out, thus making it impossible to re-onboard the POM.

The search filter has been updated to only filter existing "non-deleted" POMs (active, unavailable or inactive) as these can be updated via the profile edit page (they show on the staff lists). Deleted POMs will be returned so they can be re-onboarded.


AledWEvans (approved):
LGTM

PR #2909, opened 2026-08-12
MO-2043 Hardening of LDU import, handle edge cases

zheileman:
Ticket: https://dsdmoj.atlassian.net/browse/MO-2043

A few changes to the import so an API failure or misconfiguration that potentially results in many LDUs missing, will abort, and also to be able to update existing LDUs that have been deleted and then recreated with the same code in mailbox register.


AledWEvans (approved):
LGTM

PR #2911, opened 2026-08-14
MO-2044 Improve/optimise OmicEligibility job

zheileman:

First pass on a series of PRs intended to refactor/improve some nightly jobs that use prisoner-search, to decide if a case/offender has been released or is no longer inside OMIC policy, etc.

- Switched from per-record create/update flow to upsert_all, removing old race/window defects and making persistence atomic and much faster.
- Kept cleanup simple (single-pass delete of missing offenders).
- Added not null to prison DB field, and a bunch of logging.

The idea being, we can rely more on this table if we trust its data, to be used in other existing or new jobs that also run nightly and requires an almost identical dataset (and is costly to build).


app/services/persist_omic_eligibility_service.rb
```diff
@@ -0,0 +1,97 @@
+# frozen_string_literal: true
+
+class PersistOmicEligibilityService
+  API_ARGS = { fetch_complexities: false, fetch_categories: false, fetch_movements: false }.freeze
```
zheileman, line 4: `ignore_legal_status: true` removed, so the default (false) applies, as that's what we really want here, we want the pre-filtered by legal status offender collection (as we only apply OMIC rules upon those who pass this first filter), NOT to ignore the legal status.

andrewrlee (approved):
LGTM

PR #2914, opened 2026-08-17
MO-2044 Debounce prisoner status changes

zheileman:

PR part of a series of improvements to better handle released or unsentenced offenders.

Same as we did with other events, we tend to sometimes get a few of these status changes in quick succession, we can batch them for a few seconds before processing them, to ensure we fetch the latest data. Also disabled cache in this API call.

Added the CNL inactivation as part of the status changes as well, which saves us a nightly job that was doing the same in a more inefficient way. Removed the cronjob.


andrewrlee (approved):
LGTM

PR #2916, opened 2026-08-18
MO-2044 Hardening of nightly movements job

zheileman:

PR part of a series of improvements to better handle released or unsentenced offenders.

Hardens the scheduled movement processing flow. Allows to configure the number of days and date to process in the rake task and batch it per-offender, with retries.

Disable cache in the API call to avoid using stale data.

### DIFF
--- a/app/services/hmpps_api/prison_api/movement_api.rb
+++ b/app/services/hmpps_api/prison_api/movement_api.rb
@@ -8,7 +8,7 @@ module HmppsApi
       ADMISSION_TYPES = [HmppsApi::MovementType::ADMISSION,
                          HmppsApi::MovementType::TRANSFER].freeze
 
-      def self.movements_on_date(date)
+      def self.movements_on_date(date, cache: true)
         route = '/movements'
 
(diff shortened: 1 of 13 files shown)


app/services/hmpps_api/prison_api/movement_api.rb
```diff
@@ -17,7 +17,7 @@ def self.movements_on_date(date)
         data = client.get(route, queryparams: {
           movementDate: date.strftime('%F'),
           fromDateTime: (date - 1.year).strftime('%FT%R')
-        })
+        }, cache:)
```
andrewrlee, line 20: is this valid syntax? 🤯

andrewrlee (approved):
LGTM

app/services/hmpps_api/prison_api/movement_api.rb
zheileman: Yeah, if there is a local variable in scope with the same name. In this case, ruby auto-resolves `cache` because it is a named argument in the method signature.

PR #2917, opened 2026-08-18
Add missing indexes to optimise queries

zheileman:
Some of the reports and jobs we have rely on certain fields from table that didn't have indexes so it was very slow.

These will not block DB connections while the index is added.


andrewrlee (approved):
LGTM

PR #2919, opened 2026-08-19
MO-2048 Improve async handling of tier events

zheileman:
Ticket: https://dsdmoj.atlassian.net/browse/MO-2048

We’ve had a few instances of an influx of thousands (60k+) `tier.calculation.changed` events, usually in non-prod envs but yesterday also in production.

The events processor was not enqueuing these as jobs, and instead run these inline, which takes a trip to the DB and API and blocks the thread. It took a long time to all clear the events.

This PR makes it so we enqueue these events processing as we do with the rest of events, so it is consistent, and handles transient API errors retries.

Also improved validation of tier ahead of attempting to save it.


andrewrlee (approved):
LGTM

PR #2920, opened 2026-08-19
MO-2047 Check POM exists in caseload pages

zheileman:
Ticket: https://dsdmoj.atlassian.net/browse/MO-2047

Additional validation check to ensure we are not trying to load non-existent POMs or not properly onboarded (no role).

Copy on the error page might change but wanted to get the protection up asap.


andrewrlee (approved):
LGTM

PR #2921, opened 2026-08-20
MO-2046 Reconcile task for released/non-omic cases

zheileman:
Ticket: https://dsdmoj.atlassian.net/browse/MO-2046

A manual, ad-hoc cleanup task to remove stale local data for offenders who have left the prison service but whose records weren't cleaned up automatically (for example, because a domain event was missed, or we didn't have event handling back then).

It uses OmicEligibility. Anyone absent from this table is considered a potential cleanup candidate but then we narrow down by checking with prisoner-search the orphaned IDs.

Based on the API response, we do the following:

* Not found: treat as released and call release mechanism and prune
* prisonId is OUT and not a restricted patient: same as above
* prisonId is OUT and is a restricted patient: do nothing
* Found on ROTL: do nothing
* Found with an allowed legal status: do nothing
* Found with a disallowed legal status: queue status change job (can deallocate but will not prune)


andrewrlee (dismissed):
LGTM

andrewrlee (approved):
LGTM

PR #2924, opened 2026-08-21
Configure sensible http timeout defaults

zheileman:
We didn't have any way to set the timeout defaults for http API calls.

Introduce some configuration and sensible defaults. Added additional logging. Timeouts are transient errors and will be retry (as part of our existing retry mechanism).


SebNegahbanMoJ (approved):
Looks reasonable even if I don't fully understand the context!

